PR #13, opened 2022-06-27
axis_geo()

### LINKED ISSUE #1: axis_geo() development
Functionality of the function

- [x] Plotting a period bar
- [x] Plotting a stage bar
- [x] Plotting a joint period + stage bar
- [x] manual age limits input
- [x] option for stage / period colours (default to white bar with black labels)
- [ ] intelligent label lengths
- [ ] intelligent switching between periods and stages if not specified
- [ ] intelligent bar widths (adjusting to the plotting window size?)
- [x] documentation
- [x] tests

willgearty:
This adds a function to plot geological timescales as an axis on a base R plot:
image

See #1 for discussion of development and some goals. Note that **none** of the `R cmd check` problems are related to this pull request.

Closes #1.


LewisAJones:
# Review 1 (Lewis)
Guys, this is a great effort - Well done! It works really well.

The general direction of the function has changed from generating a quick time-series plot to just adding the axis of the geological timescale. Personally, I think this works well and has greater flexibility. However, we should consider that this is now something that requires extra steps from the user (setting plot axes to FALSE, adding box() or additional axis()/labels(). If you see an opportunity to reduce these additional steps for our users, please try to implement them. Regardless, you have made this way easier for our community.

See below for some comments. I've tried to be as thorough as possible at this stage to prevent excessive rounds of review. However, they are all relatively minor, so please don't be intimidated by the length of this review.

## Documentation:

1. Repeated 'of' in the description documentation.
2. What geological timescale is used? I think you should be more explicit in the description area of documentation.

### Arguments

For all arguments (where possible), please start the argument description with what class is needed as an input value, i.e. `integer`. You should also state the 'default' option for all arguments. 

1. side 
- I would update text to: "`integer`. Which side should the axis be added to? (1 = bottom, 2 = left, 3 = top, 4 = right)". 
2. intervals
- This description should be made a little more explicit. I would start the description with a sentence (similar to `fill`) of "The interval information to use to plot the axis: (A) a `character` string..." or something...
- The built-in dataframes are calling from the deeptime package right? We already have GTS2012 and GTS2020 in the package. I don't think we should be calling the same data from different places. Let's chat about this. It is also unclear from a user-perspective which GTS is being used (I think it is very important that we provide all of this information)? 
- Maybe the link to macrostrat should be place in the details section of the function?
3. height
- All good!
4. fill
- Preference for UK spelling of 'colour' to be consistent with the rest of the package (including name)
5. lab
- What is required by this argument? A logical, a character string? Please specify. If logical, perhaps make it into a formulated question: "Should interval labels be included?"
6. lab_color
- See above WRT the use of 'color' 
7.  size
- maybe change to lab_size? This might prevent confusion with the height argument?
8. rot
- All good!
9. abbrv
- Again, I would favour an explicit question here. Currently, I do not think it reads well even if any reasonable person can interpret exactly what you mean. 
- Standardise abbrv/abbr?
10. center_end_labels
- All good!
11. skip
- We need to state here what is being skipped by default. This could be confusing/frustrating from the user perspective if working on more recent stuff?
12. bord_color
- All good!
13. lty
- All good!
14. lwd
- All good!
15. neg
- Do not use 'true', use `TRUE`.
- Do we want to allow negative values as an input? This is part of a wider discussion relating to the time_bins() function too. 
16. exact 
- Do not use 'true', use `TRUE`.
17. round
- Looks good!
18. at
- Looks good!
19. labels
- Can labels argument not be grouped with lab and lab_color?
- Be more explicit here, it is not clear from a user perspective how this interacts with the other label arguments. 
- This was not immediately clear to me that this is numeric labels, and not the GTS labels.
- What about axis title? 

### Details

"The color column is also optional and lists a for the background" 'colour' missing here?

Add 'auditor(s):' 

## EXAMPLES:

They work as expected, and show off the functionality of `axis_geo()`. A couple of suggestions:
- Consider also showing how the user can add an appropriate label such as "Time (Ma)" to the appropriate axis. You could also consider including this as an additional argument in the function. 
- Consider examples using custom or macrostrat scales?

## FUNCTIONALITY:

1. Some stage abbreviations are quite long (4 characters, perhaps try to use only 3 characters or less), 'Trns' for example.
2. Add a black line to the maximum/minimum age of the scale when plotting? I appreciate that the plot can't be extended to cover the full range of the plotted intervals as axis_geo() is an add-on feature. However, it is currently a bit unappealing to not show the limits of plotted interval. 
3. Additional white space when adding axis_geo() due to base R slightly expanding the plotting area. Did you consider adding a grey box here or similar to fill this empty space? This would make figures more 'publication-ready' in my opinion.
4. Options exist for changing the lwd and lty, do we imagine a scenario where somebody might want to change the line colour? I suspect not, what do you think?

## CODE:

The function seems well coded to me. It is easy to follow your code, and it works as expected. You comment the code at each important step and the variable names make sense. As we discussed at the start of our project, we want to minimise the number of packages that our functions depend on. However, as deeptime is your own package that you maintain Will, I do not think that is a problem here.

My main suggestion would be to add additional error handling for incorrect inputs. For example, if somebody inputs "stage" instead of "stages" for intervals, it is not very clear what the error is. Consider what common mistakes people might make and try to provide explicit error messages.

## TESTS:

Tests seem appropriate for the function and a high code coverage is generated.

## NOTES:
`R CMD check` issues are not related to the developed function.

willgearty:
Thanks for all of the great comments and suggestions @LewisAJones! So far, I've done the following:

1. Clean up and improve the overall/usage documentation and some argument names.
2. Add more examples that exemplify how a title would be added and how a user could use different timescales.
3. Added a neutral background color to the scale when intervals don't extend all the way to the axis limits and added lines at the end of the scales.

Still to be done:

- [ ] More error handling

I also have the following response comments:

1. Some of your suggestions are useful, but should be directed towards changes in `deeptime` rather than changes here. For example, there should probably be better error handling in `getScaleData` and there could even be partial matching (e.g., "stage" matches to "stages"). Also, the long abbreviations are a result of the automated abbreviations created [here](https://github.com/willgearty/deeptime/blob/master/data-raw/ICS_data.R). If the abbreviations are long, it probably means a shorter abbreviation wouldn't be unique. I can put together a github issue for these suggestions, but I don't think this PR should be responsible for making those changes.
2. Note that line color can already be modified using the `bord_color` argument.
3. Technically, `deeptime` has the 2021 timescale (which I should probably update to 2022 soon), so there isn't actually any overlap in data.
4. In terms of British spellings, this wasn't specified in the standards, so I didn't think about it. I really don't think it's a big deal to have American spellings, but I also know I'll get outvoted over this if we bring it to a vote among the team. One way to deal with this is to accept both spellings like lots of tidyverse packages do (I would need to investigate how this works, though).
5. I really don't think defaults need to be included in the argument documentation as they are already included above at the top of the Usage section. Looking at various other packages (including those in the tidyverse), this appears to be the standard.

willgearty:
Update: I added more name/error handling to `getScaleData()` in `deeptime` and I updated the built-in timescales to the 2022/02 versions.

LewisAJones:
Thanks Will for all your changes so far, they look great to me! I think the additional examples are particularly useful for users. 

A few quick comments to your replies to my comments:

1. Happy with that - partial matching would definitely be a better way to go than throwing an error. I see you've already updated to 2022/02 version of the GTS too. I still think you should specify in the documentation of the axis_geo() function which version of GTS it is that axis_geo() is using, and link to the source. This could be added into the details section. I don't really think there is a need to be able to plot different versions of GTS (e.g. the popular GTS2012 and GTS2020). If users want to do this, they can use your custom intervals functionality. However, I do feel it is important to provide specific details to the user regarding which data we use. I am sure they could work this out fairly easily, but why not just state it upfront?
2. Thanks Will, sorry I missed this before.
3. Happy with that, but see point 1. 
4. If there is a way to accept both forms of spelling, I think this would be amazing. I know that tidyverse are doing this, but not sure how easy it is to implement. Let me know if you find a solution. From a code perspective, I don't think it actually matters whether it is US or British spelling within the function body. However, it would be nice from a user perspective to have arguments more or less standardised across functions.
5. I don't see why our differences in opinion here should stop this pull request being accepted. I personally think it can be useful for the user for some specific arguments when the default behaviour might result in unexpected behaviour. For example, in axis_geo() skipping Quaternary, Holocene, and Late Pleistocene. Obviously the default arguments are already provided above, but I think a sentence saying something like "Quaternary, Holocene, and Late Pleistocene are skipped by default. Set to NULL if this is not desired." in the argument section could be welcomed. However, I leave this with you.

willgearty:
^ Force-pushing to update to the recent changes on main

willgearty:
It looks like it's fairly easy to make aliases of function names, e.g.:
https://github.com/tidyverse/ggplot2/blob/759c63c2fd9e00ba3322c1b74b227f63c98d2e06/R/zxx.r#L128-L223
And then if you have arguments that you want to have aliases, it takes a little bit more work (and will end up listing both arguments in the documentation):
https://github.com/tidyverse/ggplot2/blob/7571122f6a9455ccce9f8ceb9a2196c0a70be646/R/theme-elements.r#L64-L67

willgearty:
OK, I added a bunch of error handling and added some minor documentation about defaults where I thought they were useful (re: # 5 above).

1. For the source details, I'm hesitant to add any more details than I already have. The documentation links to the documentation for the data that is used (`deeptime::periods` etc), and that documentation will always be up-to-date, so I'd rather not make a second place that I'd need to update every time `deeptime` is updated.
4. See my comment above. Let me know what you think.

LewisAJones:
> OK, I added a bunch of error handling and added some minor documentation about defaults where I thought they were useful (re: # 5 above).
> 
> 1. For the source details, I'm hesitant to add any more details than I already have. The documentation links to the documentation for the data that is used (`deeptime::periods` etc), and that documentation will always be up-to-date, so I'd rather not make a second place that I'd need to update every time `deeptime` is updated. 4. See my comment above. Let me know what you think. I'm happy to discuss with the team and make a decision before we merge all three of the active PRs.

1. Regarding source details, I think the links you've included make this easy enough to access this information. So, I am happy with this this. It would be nicer to have this clearly stated in the function documentation, but I understand your reasoning and I don't think it is too much of an issue as long as the information is accessible.
4. WRT British/American spelling and whether to accept both, we can talk with the team. I think ultimately it would be cool to be able to input both, but I am not sure the reward for the effort will be worth it? This could also be an extra step for future contributors to overcome? I also don't like the idea of having multiple arguments of the same thing in the documentation. So I am leaning towards accepting both spelling as a standard (rather than coded) at the potential annoyance of the user on the odd occasion, or just accepting one form of spelling.

Once we've sorted point 4, I think this pull request is ready for acceptance. Thanks for all your work!

LewisAJones (approved):
Well, I think this is good to go! Thanks for all your work Will and Kilian. I think this will be very useful for the community. Great job!

PR #33, opened 2022-08-05
Bin spatial

### LINKED ISSUE #24: bin_spatial()
A function to bin occurrence data into spatial bins. I know @LewisAJones and @bethany-j-allen have their own ways of doing this. Perhaps it would also be useful to incorporate other methods, despite potential reservations (e.g., rectangular bins, the Close et al. MST method, etc).

LewisAJones:
# Function name
bin_spatial()

## Description
A function to assign fossil occurrences to spatial bins/samples using a hexagonal equal-area grid, or a distance-based approach.

## Motivation and Context
Palaeobiologists have a growing interest in spatial analyses using fossil occurrence data (e.g., considering the impact of spatial bias or correcting for it). This function allows users to bin their data into spatial bins using two different approaches: 
-  Equal-area grid: The "grid" method bins fossil occurrence data into equal-area grid cells using a hexagonal grid generated via the h3jsr package.
- Distance: The "dist" method identifies unique localities in the input occdf (i.e., unique pairs of coordinates) and generates a spatial sample for each locality based on a user-defined distance. All occurrences within the specified distance from the reference locality are drawn as a single spatial sample. As high density of occurrences might result in numerous samples for the same area, the buffer argument has been implemented to allow users to generalise the definition of a unique locality to incorporate a broader geographic area. The functionality of this approach is heavily dependent on the sf::st_is_within_distance() function.

As Gwen Antell is developing a purpose made R package for spatial subsampling, I do not think we need greater functionality than what is generically implemented here. However, let me know your thoughts! :-)

Closes #24


R/bin_spatial.R
```diff
@@ -0,0 +1,220 @@
...
+#' @return If the `method` argument is specified as "grid" and `return` as
+#' `FALSE`, a dataframe is returned of the original input `occdf` with
+#' cell information: ID number (`cell_ID`) and cell centroid coordinates
+#' (`cell_centroid_lng` and `cell_centroid_lat`). If `return` is
+#' set to `TRUE`, a list is returned with both the input `occdf` and grid
+#' information.
+#' If the `method` argument is specified as "dist", a list is returned of
+#' spatial samples drawn around unique localities based on the
+#' user-defined `dist`. Each element of the list contains occurrences within
+#' the geographic distance of the reference locality defined by the user.
```
bethany-j-allen, line 44: Was it a conscious decision to output the data in this way? Almost any downstream analysis will require these data frames to be unlisted into a single frame (as the 'grid' output is), maybe it would be better, as well as more consistent, to do this.

R/bin_spatial.R
```diff
@@ -0,0 +1,220 @@
...
+#' slightly in their coordinates (e.g., < 1 kilometre).
+#' Note: `buffer` should be provided in kilometres. The default value is `NULL`,
+#' which skips the buffer implementation.
+#' @param return \code{logical}. Should the equal-area grid information be
+#'  returned? Only useful if the method argument is set to "grid".
+#'  The default is `FALSE.`
+#'
+#' @return If the `method` argument is specified as "grid" and `return` as
+#' `FALSE`, a dataframe is returned of the original input `occdf` with
+#' cell information: ID number (`cell_ID`) and cell centroid coordinates
```
bethany-j-allen, line 37: Does the ID number have any useful external meaning? If not, they seem overly complicated, would it be cleaner to just renumber them?

R/bin_spatial.R
```diff
@@ -0,0 +1,220 @@
...
+#' # Get internal data
+#' data("tetrapods")
+#' # Smaller dataframe for examples
+#' occdf <- tetrapods[1:500, ]
+#' # Bin data using a hexagonal equal-area grid
+#' bin_spatial(occdf = occdf, method = "grid", dist = 250)
+#' # Bin data based on distance for each unique location
+#' bin_spatial(occdf = occdf, method = "dist", dist = 250)
+#' # Bin data based on distance with a buffer to define each unique location
+#' bin_spatial(occdf = occdf, method = "dist", dist = 250, buffer = 100)
```
bethany-j-allen, line 91: Examples using the output of palaeorotate should be included when that's ready, and we could also add an example using tax_unique looking for unique taxa within different grid cells :)

R/bin_spatial.R
```diff
@@ -0,0 +1,220 @@
...
+  if (any(occdf[, lat] > 90) || any(occdf[, lat] < -90)) {
+    stop("Latitudinal coordinates should be more than -90 and less than 90")
+  }
+
+  if (any(occdf[, lng] > 180) || any(occdf[, lng] < -180)) {
+    stop("Longitudinal coordinates should be more than -180 and less than 180")
+  }
+
+  if (!is.character(method)) {
+    stop("`method` should be of class character")
```
bethany-j-allen, line 124: I think this error is redundant given the following one? Not that it really matters if it is!

R/bin_spatial.R
```diff
@@ -0,0 +1,220 @@
...
+#' allow users to generalise the definition of
+#' a unique locality to incorporate a broader geographic area. The
+#' functionality of this approach is heavily
+#' dependent on the
+#' \code{\link[sf:st_is_within_distance]{sf::st_is_within_distance()}} function.
+#'
+#' Note: prior to implementation of either method, the coordinate reference
+#' system (CRS) for input data is defined as EPSG:4326 (World Geodetic System
+#' 1984). The user might wish to update their data accordingly if this is
+#' not appropriate.
```
bethany-j-allen, line 73: I still find CRSs so confusing, I often struggle to find information about the different systems and what the differences are. If you know of any good resources along these lines, maybe you could link them here?

R/bin_spatial.R
```diff
@@ -0,0 +1,220 @@
...
+#' data("tetrapods")
+#' # Smaller dataframe for examples
+#' occdf <- tetrapods[1:500, ]
+#' # Bin data using a hexagonal equal-area grid
+#' bin_spatial(occdf = occdf, method = "grid", dist = 250)
+#' # Bin data based on distance for each unique location
+#' bin_spatial(occdf = occdf, method = "dist", dist = 250)
+#' # Bin data based on distance with a buffer to define each unique location
+#' bin_spatial(occdf = occdf, method = "dist", dist = 250, buffer = 100)
+#' @export
```
bethany-j-allen, line 92: Following our previous discussion about plotting latitude bands, I think having a simple way to plot the grid cells / distance-based points would be more useful!

bethany-j-allen (commented):
Thanks @LewisAJones, I think this is looking good! My main criticism is that I don't understand the aim of the 'distance' approach, so I'm struggling to evaluate this half of the code, but I'm happy to come back to it once the explanation is a bit more detailed. I'll also have a think about whether there's anything else that might be useful to add...

R/bin_spatial.R
```diff
@@ -0,0 +1,220 @@
...
+#' Using this distance, the function identifies which resolution is most
+#' similar to the input `dist`, and uses this resolution.
+#' - Distance: The "dist" method identifies unique localities in the input
+#' `occdf` (i.e., unique pairs of coordinates) and generates a spatial sample
+#' for each locality based on a user-defined distance. All occurrences within
+#' the specified distance from the reference locality are drawn as a single
+#' spatial sample. As high density of occurrences might result in numerous
+#' samples for the same area, the `buffer` argument has been implemented to
+#' allow users to generalise the definition of
+#' a unique locality to incorporate a broader geographic area. The
```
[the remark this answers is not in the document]
LewisAJones: This has all been dropped and is not quite relevant now. As such, I'll resolve this conversation for now.

R/bin_spatial.R
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  if (!is.numeric(dist)) {
+    stop("`dist` should be of class numeric")
+  }
+
+  if (!is.null(buffer) && !is.numeric(buffer)) {
+    stop("`buffer` should be NULL or of class numeric")
+  }
+
```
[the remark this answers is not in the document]
LewisAJones: Dropped, so I am resolving.

R/bin_spatial.R
LewisAJones: Dropped, resolving.

R/bin_spatial.R
LewisAJones: Yes... I am also not completely happy with this as it is overly-complicated. However, these ID numbers do link back to the discrete equal-area grids. With these IDs you can link back to the specific resolution and exact cell, so I am reluctant to modify. Thoughts?

R/bin_spatial.R
LewisAJones: Dropped, resolving.

R/bin_spatial.R
LewisAJones: I've linked to a book that I consider more or less the bible of geo computation in R. I think this is a great resource as it covers many themes. It is also freely available.

R/bin_spatial.R
LewisAJones: I've now added plotting functionality.

LewisAJones:
> Thanks @LewisAJones, I think this is looking good! My main criticism is that I don't understand the aim of the 'distance' approach, so I'm struggling to evaluate this half of the code, but I'm happy to come back to it once the explanation is a bit more detailed. I'll also have a think about whether there's anything else that might be useful to add...

Hi Bethany, thanks for all the feedback. I've had a major revamp after our chat on Monday. Hopefully it is all a bit more straightforward now and makes a bit more sense. Let me know your thoughts!

@KEichenseer I know you started taking a look yesterday, but the function has been majorly updated today, so please check this version.

KEichenseer (approved):
Thanks Lewis for the reworking of this function! The functionality and documentation is clear and easy to understand now. The sub-grid is a cool addition. 

Here are a few things that could still be improved:

- The plotting of the grid cells appears to bug out in some cases when the longitude is close to 180 or -180.  For example:
```
df <- data.frame(lat = c(0,-15,90), lng=c(-179,0,179))
occdf <- bin_spatial(occdf = df, spacing = 20000, sub_grid = 200, plot = T)
```
- The grid information is not returned when `return = T` if `plot = F`. 

- this is a matter of preference, but it could be nice to have the whole grid plotted, not only the cells which contain data. Maybe this could be the default for coarse grids. For high-res grids it is probably too messy and hard to see. Also, I'd restrict the plot area to -90 and 90 deg lat to avoid the label "-100 deg S".

- The sub-grid argument is introduced a bit abruptly. Maybe start like this: "numeric. For an optional sub-grid, the desired spacing between the centers of adjacent cells in the sub-grid. This value should be provided in kilometres. See details for information on sub-grid usage."

- Generally, what do you think about standardising the name of input data frames across functions? I find "occdf" a bit cumbersome, maybe something like "df" or "data" would suffice.

LewisAJones:
@bethany-j-allen did you have any further comments on this PR, or are we good to go? :-) It would be great to have this merged before tomorrow's meeting!

bethany-j-allen:
No, this looks good to me. I like the rarefy example :)

PR #34, opened 2022-08-06
Tax range

### LINKED ISSUE #28: tax_range()
A function for calculating the temporal, latitudinal, and/or spatial range of taxa (based on occurrence data).

LewisAJones:
# Function name
tax_range_time() and tax_range_geo()

## Description

tax_range_time() 
A function to calculate the temporal range of fossil taxa.

tax_range_geo()
A function to calculate the geographic range of fossil taxa. The function can calculate geographic range in four ways: convex hull, latitudinal range, maximum great circle distance, and the number of occupied equal-area hexagonal grid cells.

## Motivation and Context

These functions allow the user to calculate common metrics used in palaeobiology (temporal and spatial ranges). Although their use is common, no standard implementation exists. Their inclusion in the package will allow researchers to avoid re-inventing the wheel and provides readable/documented code for how these metrics are calculated.

Closes #28


R/tax_range_geo.R
```diff
@@ -0,0 +1,251 @@
...
+  if (!is.character(method)) {
+    stop("`method` is not of character class")
+  }
+
+  if (!is.numeric(occdf[,lat]) || !is.numeric(occdf[,lng])) {
+    stop("`lng` and/or `lat` columns are not of numeric class")
+  }
+
+  if (any(is.na(occdf[,name]))) {
+    stop("The `name` column contains NA values")
```
bethany-j-allen, line 111: It should maybe be mentioned somewhere in the documentation that these should be filtered out beforehand.

R/tax_range_geo.R
```diff
@@ -0,0 +1,251 @@
...
+
+  # Possible methods specified?
+  possible_methods <- c("con", "lat", "gcd", "occ")
+  method_match <- charmatch(method, possible_methods)
+
+  if (is.na(method_match) == TRUE) {
+    # If the user has entered a non-valid term for the "method" argument,
+    # generate an error and warn the user.
+    stop("Invalid `method`. Choose either:
+  'con', 'lat', 'gcd', and 'occ'.")
```
bethany-j-allen, line 126:
```suggestion
  'con', 'lat', 'gcd', or 'occ'.")
```

R/tax_range_time.R
```diff
@@ -0,0 +1,178 @@
...
+      points(x = temp_df$FAD_ma,
+             y = temp_df$taxa_id,
+             pch = 20,
+             col = temp_df$taxa_id)
+      points(x = temp_df$LAD_ma,
+             y = temp_df$taxa_id,
+             pch = 20,
+             col = temp_df$taxa_id)
+      axis_geo(side = 1, intervals = "periods")
+      title(xlab = "Time (Ma)", line = 4)
```
bethany-j-allen, line 173: For the plot (but also maybe for the table?), it would be nice to be able to sort by time, or at least that's how this information is typically displayed in figures.

R/tax_range_time.R
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    # Return dataframe
+    return(temp_df)
```
bethany-j-allen, line 177: Might be useful to add a column with the number of occurrences?

R/tax_range_time.R
```diff
@@ -0,0 +1,178 @@
...
+#' and last appearance. The temporal duration of each taxa is also calculated.
+#'
+#' @section Developer(s):
+#' Lewis A. Jones
+#' @section Reviewer(s):
+#' To be reviewed
+#' @importFrom graphics points
+#' @examples
+#' # Grab internal data
+#' occdf <- tetrapods
```
bethany-j-allen, line 44: As above, I think it would be best to clean this for either species or genera in the example.

R/tax_range_time.R
```diff
@@ -0,0 +1,178 @@
+#' Calculate the temporal range of fossil taxa
+#'
+#' A function to calculate the temporal range of fossil taxa.
```
bethany-j-allen, line 3:
```suggestion
#' A function to calculate the temporal range of fossil taxa from occurrence data.
```

R/tax_range_time.R
```diff
@@ -0,0 +1,178 @@
...
+#'
+#' @param occdf \code{dataframe}. A dataframe of fossil occurrences. The
+#' dataframe should contain at least three columns: names of taxa (`character`)
+#' and the maximum (`numeric`) and minimum (`numeric`) age which constrain the
```
bethany-j-allen, line 7:
```suggestion
#' and the maximum (`numeric` or `character`) and minimum (`numeric` or `character`) age which constrain the
```

R/tax_range_time.R
```diff
@@ -0,0 +1,178 @@
...
+#' age range (in millions of years) of the fossil occurrence. If the age range
```
bethany-j-allen, line 8:
```suggestion
#' age range of the fossil occurrence. If the age range
```

bethany-j-allen, line 8: Do the values have to be in millions of years? Surely the unit you put in is the unit you get out. I'm just thinking that maybe some people who work with near-recent data might want to use the function. In this case, it might be worth renaming some of the variables, e.g. 'max_ma' -> 'max_time'.

R/tax_range_time.R
```diff
@@ -0,0 +1,178 @@
...
+#' columns are provided as (`character`) values (i.e., stage names), the
+#' function will attempt to extract numeric age data from the
+#' Geological Timescale by linking stage names.
```
bethany-j-allen, line 11:
```suggestion
#' Geological Timescale by matching the provided character strings to the stage names.
```

R/tax_range_time.R
```diff
@@ -0,0 +1,178 @@
...
+#' @param name \code{character}. The name of the column you wish to be treated
+#' as the input names (e.g., "species" or "genus").
```
bethany-j-allen, line 13:
```suggestion
#' as the taxonomic names (e.g., "species" or "genus").
```

R/tax_range_time.R
```diff
@@ -0,0 +1,178 @@
...
+#' @param min_ma \code{character}. The name of the column you wish to be treated
+#' as the input minimum age range (e.g., "min_ma").
```
bethany-j-allen, line 15:
```suggestion
#' as the minimum limit of the age range (e.g., "min_ma").
```

R/tax_range_time.R
```diff
@@ -0,0 +1,178 @@
...
+#' @param max_ma \code{character}. The name of the column you wish to be treated
+#' as the input maximum age range (e.g., "max_ma").
```
bethany-j-allen, line 17:
```suggestion
#' as the maximum limit of the age range (e.g., "max_ma").
```

R/tax_range_time.R
```diff
@@ -0,0 +1,178 @@
...
+#' @param scale \code{character}. Specify the desired geological timescale to
+#' be used "GTS2020" or "GTS2012". This argument is only useful if the supplied
```
bethany-j-allen, line 19:
```suggestion
#' be used, either "GTS2020" or "GTS2012". This argument is only useful if the supplied
```

R/tax_range_time.R
```diff
@@ -0,0 +1,178 @@
...
+#' "max_ma" and "min_ma" columns are of class \code{character}.
+#' "GTS2020" is the default.
```
bethany-j-allen, line 21: Could link to look_up() here?

R/tax_range_time.R
```diff
@@ -0,0 +1,178 @@
...
+               col = temp_df$taxa_id)
+      points(x = temp_df$FAD_ma,
+             y = temp_df$taxa_id,
+             pch = 20,
+             col = temp_df$taxa_id)
+      points(x = temp_df$LAD_ma,
+             y = temp_df$taxa_id,
+             pch = 20,
+             col = temp_df$taxa_id)
+      axis_geo(side = 1, intervals = "periods")
```
bethany-j-allen, line 172: axis_geo isn't working for me here, although I guess it shouldn't be a problem after merging.

R/tax_range_time.R
```diff
@@ -0,0 +1,178 @@
...
+#' "max_ma" and "min_ma" columns are of class \code{character}.
+#' "GTS2020" is the default.
+#' @param plot \code{logical}. Should a plot of the ranges be generated?
+#' @param return_error \code{logical}. Should a vector of numbers be returned
+#' to flag the rows of the `occdf` that cannot be matched to `character`
+#' interval names? This is only relevant if `max_ma` and `min_ma`
+#' are `character` values.
+#'
+#' @return A \code{dataframe} containing the following columns:
+#' unique taxa (`taxa`), taxa ID (`taxa_id`), first appearance of taxa
```
bethany-j-allen, line 29: See above re: taxon_id

bethany-j-allen, line 29:
```suggestion
#' unique taxa (`taxa`), taxa ID (`taxa_id`), first appearance of taxon
```

R/tax_range_time.R
```diff
@@ -0,0 +1,178 @@
...
+#' (`FAD_ma`), last appearance of taxa (`LAD_ma`), and temporal range duration
```
bethany-j-allen, line 30:
```suggestion
#' (`FAD_ma`), last appearance of taxon (`LAD_ma`), and duration of temporal range
```

R/tax_range_time.R
```diff
@@ -0,0 +1,178 @@
...
+#' (`range_myr`) is returned.
```
bethany-j-allen, line 31: Need to also mention plot and return_error here

R/tax_range_time.R
```diff
@@ -0,0 +1,178 @@
...
+#'
+#' @details The temporal range(s) of taxa are calculated by extracting all
+#' unique taxa (`name` column) from the input `occdf`, and checking their first
+#' and last appearance. The temporal duration of each taxa is also calculated.
```
bethany-j-allen, line 35:
```suggestion
#' and last appearance. The temporal duration of each taxon is also calculated.
```

bethany-j-allen, line 35: I think you need to mention here that these values may be confounding both information about the true duration of taxa and uncertainty in the dating of the occurrences.

R/tax_range_time.R
```diff
@@ -0,0 +1,178 @@
...
+  if (is.data.frame(occdf) == FALSE) {
+    stop("`occdf` should be a dataframe")
+  }
+
+  if (is.logical(plot) == FALSE) {
+    stop("`plot` should be logical (TRUE/FALSE)")
+  }
+
+  if (class(occdf[, max_ma]) != class(occdf[, min_ma])) {
+    stop("The class of max_ma and min_ma must be the same")
```
bethany-j-allen, line 67:
```suggestion
    stop("The class of max_ma and min_ma must be the same (`numerical` or `character`")
```

R/tax_range_time.R
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  if (any(c(name, min_ma, max_ma) %in% colnames(occdf) == FALSE)) {
+    stop("Either `name`, `min_ma`, or `max_ma`, do not exist in `occdf`")
```
bethany-j-allen, line 71:
```suggestion
    stop("Either `name`, `min_ma`, or `max_ma`, is not a named column in `occdf`")
```

R/tax_range_time.R
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  if (any(is.na(occdf[, name]))) {
+    stop("The `name` column contains NA values")
```
bethany-j-allen, line 75: Again, it should maybe be mentioned in the documentation that these need to be filtered out.

R/tax_range_time.R
```diff
@@ -0,0 +1,178 @@
...
+#' as the input maximum age range (e.g., "max_ma").
+#' @param scale \code{character}. Specify the desired geological timescale to
+#' be used "GTS2020" or "GTS2012". This argument is only useful if the supplied
+#' "max_ma" and "min_ma" columns are of class \code{character}.
+#' "GTS2020" is the default.
+#' @param plot \code{logical}. Should a plot of the ranges be generated?
+#' @param return_error \code{logical}. Should a vector of numbers be returned
+#' to flag the rows of the `occdf` that cannot be matched to `character`
+#' interval names? This is only relevant if `max_ma` and `min_ma`
+#' are `character` values.
```
bethany-j-allen, line 26: Is this needed? I think you could just output NAs for these values in the dataframe and explain that this is the source of the problem.

R/tax_range_time.R
```diff
@@ -0,0 +1,178 @@
...
+        paste(c(
+          "Check spelling of specified max_ma and min_ma intervals.
+  Available interval names are accessible via GTS2020 and GTS2012.
+  Check rows:",
+          capture.output(print(error_vec))
+        ),
+        collapse = "\n"
+      ))
+      }
+    }
```
bethany-j-allen, line 117: For names, how does this work in practice? NAs aren't allowed, but I could imagine that the PBDB data structure could be a typical one, using 'early_interval' and 'late_interval', but with only 'early_interval' occupied in the case that an occurrence is restricted to a single named interval.

bethany-j-allen (commented):
Thanks for putting this together Lewis! I think these will be really fundamental functions in the package. I've put together some thoughts on both the functionality and the documentation. Happy to discuss anything you'd like clarification on or you disagree with.

R/tax_range_time.R
LewisAJones: Great suggestion! I have added an additional argument (by), which allows the user to specify how to sort both the output and plot. Options include FAD, LAD and name.

R/tax_range_time.R
LewisAJones: Yes, why not! Integrated.

R/tax_range_time.R
LewisAJones: Updated to use genus-level assignments. Requires some NA removal before, but that is fine!

R/tax_range_time.R
LewisAJones: Added!

R/tax_range_time.R
LewisAJones: Added!

R/tax_range_time.R
LewisAJones: I think we should retain this. If the data is not provided in millions of years the plotting functionality will behave oddly!

LewisAJones: As above! We can modify this, but would require some fiddling with the plotting functionality (specifically, axis_geo).

R/tax_range_time.R
LewisAJones: Updated!

R/tax_range_time.R
LewisAJones: Might be worth leaving this more generic. In reality, the user could supply any names here (e.g. species_1)? What do you think?

R/tax_range_time.R
LewisAJones: Added, this will likely show up as an error until both branches are merged though!

R/tax_range_time.R
LewisAJones: I think this provides an easy way for users to identify exactly what rows are the issue. With this vector, they can easily remove or update their data. I think this functionality puts emphasis on the data and that there is an issue they should resolve.

R/tax_range_time.R
LewisAJones: In this case, the user should supply the min_ma and max_ma as the same (i.e. copy the data across from the early interval to late interval). I agree this requires extra handling, but I think it is good that we don't automate this as it might not be the intention of the user, and it should be flagged.

LewisAJones:
Thanks for all the great comments @bethany-j-allen! I've updated tax_range_time for now, and I will try and sort out tax_range_geo next week. Have a great weekend :-)

R/tax_range_geo.R
```diff
@@ -0,0 +1,251 @@
...
+#' A function to calculate the geographic range of fossil taxa. The function
+#' can calculate geographic range in four ways: convex hull, latitudinal range,
+#' maximum great circle distance, and the number of occupied equal-area
+#' hexagonal grid cells.
+#'
+#' @param occdf \code{dataframe}. A dataframe of fossil occurrences.
+#' This dataframe should contain at least three columns: names of taxa
+#' (`character`), longitude (`numeric`) and latitude (`numeric`).
+#' @param name \code{character}. The name of the column you wish to be treated
+#' as the input names (e.g., "species" or "genus").
```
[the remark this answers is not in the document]
LewisAJones: As before, maybe better to leave this more generic in case the user is not using taxonomic names?

R/tax_range_geo.R
```diff
@@ -0,0 +1,251 @@
...
+#' taxa ID (`taxa_id`), maximum latitude of occurrence (`max_lat`),
+#' minimum latitude of occurrence (`min_lat`), and latitudinal
+#' range (`range_lat`) is returned.
+#' - For the "gcd" method, a \code{dataframe} with unique taxa (`taxa`), taxa
+#' ID (`taxa_id`), coordinates of the two most distant points
+#' (`lng` & `lat`), and the 'Great Circle Distance' (`GCD`) between these
+#' points in km is returned.
+#' - For the "occ" method, a \code{dataframe} with unique taxa (`taxa`), taxa
+#' ID (`taxa_id`), the number of occupied cells (`cells`), and the spacing
+#' between cells (`spacing`) in km is returned.
```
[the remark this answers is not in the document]
LewisAJones: Integrated.

R/tax_range_geo.R
```diff
@@ -0,0 +1,251 @@
...
+    # Generate dataframe for population
+    oc_df <- data.frame(taxa = unique_taxa,
+                         taxa_id = seq(1, length(unique_taxa), 1),
+                         cells = rep(NA, length(unique_taxa)),
+                         spacing = rep(NA, length(unique_taxa)))
+    # Generate equal area hexagonal grid
+    # Which resolution should be used based on input distance/spacing?
+    # Use the h3jsr::h3_info_table to calculate resolution (however, this
+    # table is not exported in their package, added into palaeoverse)
+    grid <- h3_info_table[
```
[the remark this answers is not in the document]
LewisAJones: I think I've resolved this now, but can you check from your side? It seems to be passing testing and R CMD checks so not sure what the issue might be if it still standing!

R/tax_range_geo.R
```diff
@@ -0,0 +1,251 @@
...
+#' - For the "occ" method, a \code{dataframe} with unique taxa (`taxa`), taxa
+#' ID (`taxa_id`), the number of occupied cells (`cells`), and the spacing
+#' between cells (`spacing`) in km is returned.
+#'
+#' @details Four approaches (methods) exist in the `tax_range_geo` function for
+#' calculating ranges:
+#' - Convex hull: the "con" method calculates the geographic range of a taxa
+#' using a convex hull for each taxa in `occdf`, and calculates the area of
+#' the convex hull (in km\ifelse{html}{\out{<sup>2</sup>}}{\eqn{^2}}) using
+#' \code{\link[geosphere:areaPolygon]{geosphere::areaPolygon()}}.
```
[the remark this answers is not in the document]
LewisAJones: Added a little more details here, but I don't want to go overboard with this! Let me know what you think...

R/tax_range_geo.R
```diff
@@ -0,0 +1,268 @@
...
+    # Run for loop over all unique taxa
+    for (i in seq_along(unique_taxa)) {
+      # Subset df
+      tmp <- occdf[which(occdf[, name] == unique_taxa[i]),]
+      # Extract cell ID
+      cells <- suppressMessages(
+        h3jsr::point_to_h3(tmp[,c(lng, lat)], res = grid$h3_resolution)
+      )
+      # Calculate number of unique cells occupied
+      oc_df$cells[i] <- length(unique(cells))
```
ChristopherDavidDean, line 263: So the only thing I'm unsure about here is the use of what is being returned. Occupancy isn't super helpful unless there's an understanding of proportion of area occupied - by providing just which cells are occupied you obviously miss this info. Would it be possible to add an additional column to supply this info? You could use the total number of cells which all occurrences appear within as the total number, and use that to propvide a percentage of total for each unique taxa. I think this would be a more useful figure to return to the user. 

Ideally in the future, you'd want to be able to provide extent co-ordinates or some kind of mask to estimate this, but lets save that for a future edition!

R/tax_range_geo.R
```diff
@@ -0,0 +1,268 @@
...
+                         taxon_id = seq(1, length(unique_taxa), 1),
+                         cells = rep(NA, length(unique_taxa)),
+                         spacing = rep(NA, length(unique_taxa)))
+    # Generate equal area hexagonal grid
+    # Which resolution should be used based on input distance/spacing?
+    # Use the h3jsr::h3_info_table to calculate resolution
+    grid <- h3jsr::h3_info_table[
+      which.min(abs(h3jsr::h3_info_table$avg_cendist_km - spacing)), ]
+    # Add resolution spacing
+    oc_df$spacing <- grid$avg_cendist_km
```
ChristopherDavidDean, line 253: Would it be better to return 'spacing' as a separate object instead of adding it to this dataframe? It seems a bit redundant to just have the same information listed over and over again (I guess it could be useful for further analysis as a category, but it's a bit of a stretch). 

One option would be to return a list, with the dataframe as one entry and the spacing as another. Unsure if this is something that you'd want to implement as it does make getting the dataframe marginally harder, but it would make it a lot more elegant. Just a thought!

R/tax_range_geo.R
```diff
@@ -0,0 +1,268 @@
...
+#'
+#' @section Reference(s):
+#' Darroch, S. A., Casey, M. M., Antell, G. S., Sweeney, A., & Saupe, E. E.
+#' (2020). High preservation potential of paleogeographic range size
+#' distributions in deep time. The American Naturalist, 196(4), 454-471.
+#'
+#' @section Developer(s):
+#' Lewis A. Jones
+#' @section Reviewer(s):
+#' Bethany Allen
```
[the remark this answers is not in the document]
LewisAJones: Done!

R/tax_range_geo.R
```diff
@@ -0,0 +1,268 @@
...
+#' @importFrom geosphere areaPolygon
+#' @importFrom grDevices chull
+#' @importFrom geosphere distm distHaversine
+#' @importFrom h3jsr point_to_h3
+#' @examples
+#' # Grab internal data
+#' occdf <- tetrapods
```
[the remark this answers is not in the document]
LewisAJones: I don't think so, but I will update it anyway to make it more explicit for users!

R/tax_range_geo.R
```diff
@@ -0,0 +1,268 @@
...
+  if (!is.numeric(occdf[,lat]) || !is.numeric(occdf[,lng])) {
+    stop("`lng` and/or `lat` columns are not of numeric class")
+  }
+
+  if (any(is.na(occdf[,name]))) {
+    stop("The `name` column contains NA values")
+  }
+
+  if (any(is.na(occdf[,lat])) || any(is.na(occdf[,lng]))) {
+    stop("`lng` and/or `lat` columns contain NA values")
```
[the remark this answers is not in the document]
LewisAJones: Done!

R/tax_range_geo.R
```diff
@@ -0,0 +1,268 @@
...
+
+  if (is.na(method_match) == TRUE) {
+    # If the user has entered a non-valid term for the "method" argument,
+    # generate an error and warn the user.
+    stop("Invalid `method`. Choose either:
+  'con', 'lat', 'gcd', or 'occ'.")
+  } else {
+    method <- possible_methods[method_match]
+  }
+
```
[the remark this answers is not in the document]
LewisAJones: I don't think this is necessary? Unless the method is "occ", spacing isn't used. As implemented, we allow a default for spacing too (100).

man/tax_range_geo.Rd
```diff
@@ -0,0 +1,130 @@
+% Generated by roxygen2: do not edit by hand
```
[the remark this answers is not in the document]
LewisAJones: Absolutely!

R/tax_range_geo.R
```diff
@@ -0,0 +1,268 @@
...
+#' for each taxon in `occdf`? Four options exist in this function:
+#' "con", "lat", "gcd", and "occ". See Details for a description of each.
+#' @param spacing \code{numeric}. The desired spacing (in km) between the
+#' center of adjacent grid cells. Only useful if the `method` argument is set
+#' to "occ". The default is 100.
+#'
+#' @return A \code{dataframe} with method-specific columns:
+#' - For the "con" method, a \code{dataframe} with unique taxa (`taxa`),
+#' taxon ID (`taxon_id`), convex hull coordinates (`lng` & `lat`), and area
+#' (`area`) in km\ifelse{html}{\out{<sup>2</sup>}}{\eqn{^2}} is returned.
```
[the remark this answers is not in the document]
LewisAJones: I've updated using your first suggestion. I think we've agreed on multiple fronts now that lists are not a desirable output for a lot of people. The return format is also currently in long format which makes it easy to work with tidyverse/ggplot2, etc which I think is desirable for some users.

R/tax_range_geo.R
```diff
@@ -0,0 +1,268 @@
...
+#' - For the "lat" method, a \code{dataframe} with unique taxa (`taxa`),
+#' taxon ID (`taxon_id`), maximum latitude of occurrence (`max_lat`),
+#' minimum latitude of occurrence (`min_lat`), and latitudinal
+#' range (`range_lat`) is returned.
+#' - For the "gcd" method, a \code{dataframe} with unique taxa (`taxa`), taxon
+#' ID (`taxon_id`), coordinates of the two most distant points
+#' (`lng` & `lat`), and the 'Great Circle Distance' (`GCD`) between these
+#' points in km is returned.
```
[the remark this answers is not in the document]
LewisAJones: Updated accordingly!

R/tax_range_geo.R
```diff
@@ -0,0 +1,268 @@
+#' Calculate the geographic range of fossil taxa
+#'
+#' A function to calculate the geographic range of fossil taxa from occurrence
+#' data. The function can calculate geographic range in four ways: convex hull,
+#' latitudinal range, maximum great circle distance, and the number of
+#' occupied equal-area hexagonal grid cells.
+#'
+#' @param occdf \code{dataframe}. A dataframe of fossil occurrences.
+#' This dataframe should contain at least three columns: names of taxa,
```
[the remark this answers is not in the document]
LewisAJones: I'm not sure this is necessary... but I've added "(see `name`, `lng`, and `lat` arguments)" to appease the reviewer :-D

R/tax_range_geo.R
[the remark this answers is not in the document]
LewisAJones: Take that back, this will cause the example to fail unless I assign to another object first. To prevent additional lines being added I've left as is (it works as data is loaded with the package)

R/tax_range_time.R
```diff
@@ -0,0 +1,158 @@
+#' Calculate the temporal range of fossil taxa
+#'
+#' A function to calculate the temporal range of fossil taxa from occurrence
+#' data.
+#'
+#' @param occdf \code{dataframe}. A dataframe of fossil occurrences containing
+#' at least three columns: names of taxa, maximum age and minimum age.
```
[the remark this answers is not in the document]
LewisAJones: Implemented as above. I don't think "names of taxa at a chosen taxonomic level" is required here... in theory you could use any names or type of ID, or have different taxonomic levels.

R/tax_range_time.R
```diff
@@ -0,0 +1,158 @@
...
+#' as the minimum limit of the age range, e.g. "min_ma" (default).
+#' @param max_ma \code{character}. The name of the column you wish to be treated
+#' as the maximum limit of the age range, e.g. "max_ma" (default).
+#' @param by \code{character}. How should the output be sorted?
+#' Either: "FAD" (first-appearance date; default), "LAD" (last-appearance data),
+#' or "name" (alphabetically by taxon names).
+#' @param plot \code{logical}. Should a plot of the ranges be generated?
+#'
+#' @return A \code{dataframe} containing the following columns:
+#' unique taxa (`taxa`), taxa ID (`taxon_id`), first appearance of taxon
```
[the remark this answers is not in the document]
LewisAJones: This is useful for plotting in this function rather than plotting by taxonomic names. I'd also argue it is easier to see differences between numbers than taxonomic names when they are similar!

R/tax_range_time.R
```diff
@@ -0,0 +1,158 @@
...
+#' occurrences, and incomplete sampling and preservation.
+#'
+#' @section Developer(s):
+#' Lewis A. Jones
+#' @section Reviewer(s):
+#' Bethany Allen
+#' @importFrom graphics points
+#' @examples
+#' # Grab internal data
+#' occdf <- tetrapods
```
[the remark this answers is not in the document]
LewisAJones: As above!

R/tax_range_geo.R
LewisAJones: Proportional values added!

R/tax_range_geo.R
LewisAJones: As the number/proportion of occupied cells is hugely dependent on this value, I think it is good for the user to have a column with this data. An alternative would be to have a message returned but this would just get lost if the user is not actively watching the console (i.e. pipeline analysis). I see this as comparable as having a column in your dataframe for quorum level in SQS... ultimately, much easier for a user to drop a column, than to add it manually or handle a list?

ChristopherDavidDean: Yeah, I think now you've added prop. of occupied cells this should be kept. You could have a case where you want to vary the size of grids, and test proportions - in which case you'd want that kept in long format. Happy with this!

R/tax_range_time.R
```diff
@@ -0,0 +1,158 @@
...
+           ylim = y_range,
+           axes = TRUE,
+           xaxt = "n",
+           xlab = NA,
+           ylab = "Taxon ID",
+           main = "Temporal range of taxa")
+      segments(x0 = temp_df$max_ma,
+               x1 = temp_df$min_ma,
+               y0 = temp_df$taxon_id,
+               col = temp_df$taxon_id)
```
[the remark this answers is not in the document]
LewisAJones: Updated to only use black to appease the reviewer! :-D

R/tax_range_time.R
```diff
@@ -0,0 +1,158 @@
...
+#' If the input data columns contain NAs, these should be removed prior to
+#' function call.
+#'
+#' Note: this function provides output based solely on the user input data. The
+#' true duration of a taxon is likely confounded by uncertainty in dating
+#' occurrences, and incomplete sampling and preservation.
+#'
+#' @section Developer(s):
+#' Lewis A. Jones
+#' @section Reviewer(s):
```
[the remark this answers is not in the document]
LewisAJones: Done!

ChristopherDavidDean:
Hi Lewis, such a speedy turn around! Merge away, I think it's time to get these functions on the road. Well done everyone!!

willgearty:
It looks like there might be an error in one of the tests?

LewisAJones:
Hi @ChristopherDavidDean, thank you for all the great comments!

I've implemented the vast majority of your comments, and provided a response if not. 

> It looks like there might be an error in one of the tests, but only with the old version of R?

This is related to the function tax_time_expand, not the functions here. However, let's try and fix this here and then merge. I've removed the specific test that seems to be causing the issue for now.

R/tax_range_time.R
bethany-j-allen: I got there eventually, should be fine as long as deeptime and all other functions in palaeoverse are switched on when the package is loaded.

bethany-j-allen:
I can still see a couple of stray 'a taxa' or 'each taxa' that should be 'taxon' in the documentation... But otherwise think this is good to go!

PR #37, opened 2022-08-25
phylo_check

bethany-j-allen:
# Function name
phylo_check

## Description
Check a list of taxon names against the tips in a phylogeny, and/or remove tips from the tree if not included in the list.

## Motivation and Context
I think learning to wrangle trees can be quite daunting, so this is a simple function allowing those who don't have this experience to perform simple checking and editing on already-published phylogenies. Similar functions may already exist but I think this would still be a nice addition to the tools offered in this package.

- [X] I have read palaeoverse's [standards and structure](https://github.com/palaeoverse-community/palaeoverse/raw/main/man/tutorials/structure-and-standards.pdf) 
- [X] I have performed a self-review of my code.
- [X] I have added function documentation.
- [X] I have provided sufficient examples of implementation.
- [X] If it is a core feature, I have added thorough tests.


bethany-j-allen:
Hi both, the function is now fully drafted so I'm keen to hear your thoughts. It's quite simple at the moment, I considered adding something for partial matching (e.g. a species name in the list could be matched with a genus name in the tree), but I'm not sure it's necessary, especially as you can see this when out = "table". But if you think of any other ways to improve it, let me know! :)

DESCRIPTION
```diff
@@ -18,7 +18,9 @@ Imports:
     utils,
     graphics,
     methods,
-    deeptime (>= 0.2.1)
+    deeptime (>= 0.2.1),
+    ape,
+    paleotree
```
willgearty, line 23: Since we don't actually import anything from `paleotree` and only use it for the examples/tests, we can probably just put it in "Suggests"

R/phylo_check.R
```diff
@@ -0,0 +1,133 @@
+#' Check phylogeny tip names
+#'
+#' A function to check the list of tip names in a phylogeny against a vector of
+#' taxon names, and if desired, to trim the phylogeny to only include taxon
+#' names within the vector.
+#'
+#' @param tree \code{phylo}. A phylo object. Phylogenies can be read into R from
+#' .txt or .tree files containing the Newick formatted tree using
+#' \code{ape::read.tree}.
```
willgearty, line 9: Part of me thinks the extra details about reading phylogenies should be put in a "Details" section.

willgearty, line 9: Make sure to link to the ape documentation

R/phylo_check.R
```diff
@@ -0,0 +1,133 @@
...
+#' @param list \code{character}. A vector of taxon names. Binomials can be
+#' separated with either a space or an underscore. The list should not contain
```
willgearty, line 11:
```suggestion
#' separated with either a space or an underscore. The names should not contain
```

R/phylo_check.R
```diff
@@ -0,0 +1,133 @@
...
+#' #Table of taxon names in list, tree or both
+#' phylo_check(tree = ceratopsianTreeRaia, list = dinosaurs, out = "table")
+#'
+#' #Trim tree to tips in the list
+#' my_ceratopsians <- phylo_check(tree = ceratopsianTreeRaia, list = dinosaurs,
+#' out = "tree")
+#' plot(my_ceratopsians)
+#' @export
+
+phylo_check <- function(tree = NULL, list = NULL, out = "counts") {
```
willgearty, line 54: I wonder if the default for `out` should be "table"? That seems the most useful to me.

R/phylo_check.R
```diff
@@ -0,0 +1,133 @@
...
+#' of taxa included and not included in the tree ("counts", the default), a
+#' \code{dataframe} describing which taxa are included or not included in the
+#' tree ("table"), or the phylogeny trimmed to only include taxa in the provided
+#' list ("tree").
+#' @return If out = "counts", a summary table containing the number of taxa in
+#' the list but not the tree, in the tree but not the list, and in both. If out
+#'  = "table", a \code{dataframe} describing whether taxon names are present in
+#' the list and/or the tree. If out = "tree", the input phylogeny trimmed to
+#' only include the tips present in the list; you can save your new tree using
+#' \code{ape::write.tree}.
```
willgearty, line 23: Again, I think this information about `write.tree` should go in a details section. And also make sure to link to the ape documentation.

R/phylo_check.R
```diff
@@ -0,0 +1,133 @@
...
+#' any punctuation.
+#' @param out \code{character}. Determine whether to return either the counts
+#' of taxa included and not included in the tree ("counts", the default), a
+#' \code{dataframe} describing which taxa are included or not included in the
+#' tree ("table"), or the phylogeny trimmed to only include taxa in the provided
+#' list ("tree").
+#' @return If out = "counts", a summary table containing the number of taxa in
+#' the list but not the tree, in the tree but not the list, and in both. If out
+#'  = "table", a \code{dataframe} describing whether taxon names are present in
+#' the list and/or the tree. If out = "tree", the input phylogeny trimmed to
```
willgearty, line 21: Maybe mention that it is a `phylo` object?

R/phylo_check.R
```diff
@@ -0,0 +1,133 @@
...
+    stop("List of taxa to check against must be provided")
+  }
+
+  if (is.vector(list) == FALSE) {
+    stop("List of taxa must be a vector")
+  }
+
+  if (any(grepl("[^[:alnum:][:space:]_]", list))) {
+    stop("List of taxa should not contain punctuation except spaces or
+         underscores")
```
willgearty, line 74:
```suggestion
    stop("Taxon names should not contain punctuation except spaces or
         underscores")
```

R/phylo_check.R
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  if (out != "counts" && out != "table" && out != "tree") {
+    stop("out must either be counts, table or tree")
```
willgearty, line 78:
```suggestion
    stop("out must either be 'counts', 'table' or 'tree'")
```

R/phylo_check.R
```diff
@@ -0,0 +1,133 @@
...
+                        c(in_both, only_tree, only_list))
+    colnames(counts) <- c("Category", "Number of taxa")
+
+    return(counts)
+  }
+
+  #Determine which names are in which lists and table them
+  if (out == "table") {
+    table <- data.frame(all_names, names_in_tree, names_in_list)
+    table <- table[order(table$all_names), ]
```
willgearty, line 119: Part of me thinks the taxa in the list should be at the top of the table?

willgearty (changes requested):
Thanks @bethany-j-allen! I definitely have to write this code multiple times for most projects I work on, so I think this will be very useful! The code looks great to me, but I did have some comments about the documentation and other miscellaneous things.

R/phylo_check.R
bethany-j-allen: That's reasonable. I have changed this and rearranged the documentation accordingly.

DESCRIPTION
bethany-j-allen: I tried this and all the tests started failing so I put it back. It might be a syntax error with the way the package is loaded in test_that, but I'm struggling to find any clear guidance, do you know of any?

R/phylo_check.R
bethany-j-allen: I think an alphabetical sort would be helpful for identifying e.g. spelling mistakes, but agree that a presence sort would be most useful for seeing what is in the tree, so I've added this as a choice in the function. What do you think? Happy to change the default to 'presence' if you think that would be better.

DESCRIPTION
willgearty: Seems to work now?

R/phylo_check.R
willgearty: I personally think the 'presence' sort would be most useful, because I'm normally interested in which taxa in my list are not in my tree (or vice versa), but I suppose the next step is figure out why, and that could be because of spelling mistakes...I guess I'm really happy either way but 'presence' seems like the most immediately useful to me.

bethany-j-allen: Done :)

willgearty (approved):

willgearty:
Everything looks good to me! @pedrolgodoy do you have any comments? If not, I think we can merge!

pedrolgodoy:
> Everything looks good to me! @pedrolgodoy do you have any comments? If not, I think we can merge!

I do have a few basic comments! Working on them right now

R/phylo_check.R
```diff
@@ -0,0 +1,151 @@
...
+    colnames(counts) <- c("Category", "Number of taxa")
+
+    return(counts)
+  }
+
+  #Trim names not in the list from the phylogeny
+  if (out == "tree") {
+    list_in_tree <- match(tree$tip.label, list)
+    no_match <- tree$tip.label[which(is.na(list_in_tree))]
+    smaller_tree <- drop.tip(tree, no_match)
```
pedrolgodoy, line 147: First of all, I'm not sure I'm commenting in the right section. Sorry, it's my first time doing this!
About the `drop.tip` function used here, I would recommend using paleotree's `DropPaleoTip`. This is because `drop.tip` messes up with branch lengths of non-ultrametric trees and `DropPaleoTip` kind of keeps the same node ages (of the remaining nodes). Oh, and it looks like `DropPaleoTip` was recently merged into function `modifyTerminalBranches`, but  apparently `DropPaleoTip` will still work...

R/phylo_check.R
pedrolgodoy: About the "table" output format, maybe a table listing all taxa in both tree and list will be too long with very large trees/lists of taxa? Comparing this with the output from geiger's `name.check`, they only mention the taxa in the tree but not in the list ("Tree.not.data") and the opposite ("Data.not.tree"), so it will only inform you what's missing. Maybe we want both output formats? A more exhaustive one, as we currently have, a more concise one.

R/phylo_check.R
willgearty: Note that `DropPaleoTip` requires the phylogeny to have a $root.time attribute, which may not always be the case. In fact, all it does is call `drop.tip` then call `fixRootTime`, so I would say that we could mention in the details that users may want to call `fixRootTime` after running this function if their tree is not ultrametric.

pedrolgodoy: Sounds like a great solution! Thanks @willgearty!

LewisAJones:
Sorry to jump in @bethany-j-allen @willgearty @pedrolgodoy! It looks like this branch is essentially ready to go and can be merge? It would be great to have this done before tomorrow's meeting if so!

willgearty:
I don't think @pedrolgodoy's comments have been addressed yet? Or maybe they have and @bethany-j-allen never pushed the changes?

pedrolgodoy (approved):

pedrolgodoy (approved):

pedrolgodoy (approved):

pedrolgodoy:
All looking good to me! Great job @bethany-j-allen! This will be very useful!

PR #41, opened 2022-09-20
tax_time_expand

willgearty:
# Function name

`tax_time_expand()`

## Description

This function takes FAD-LAD pairs and a specified timescale (and rank) and replicates each taxon once for each interval it lived during. There is also the option to add columns to designate which intervals correspond to origination and extinction of each taxon.

## Motivation and Context

While this is by no means a replacement for actual occurrence data, this is roughly the opposite action of `tax_range_time()`, expanding taxa from ranges to pseudo-occurrences. This can often be useful for a variety of statistical analyses where the user might want to loop through time intervals and needs to know which taxa were alive in each interval. While the code is by nothing fancy, this will standardize and simplify this process. The addition of the origination and extinction columns may also be valuable for various statistical analyses, including tests of extinction selectivity.


willgearty:
Looks like the R-CMD-check is failing because I linked to a function that doesn't exist yet on main (`tax_range_time`). Otherwise, everything looks good?

R/tax_time_expand.R
```diff
@@ -0,0 +1,102 @@
...
+#'                     min_ma = c(110, 20, 0))
+#' tax_time_expand(taxdf)
+tax_time_expand <- function(
+    taxdf,
+    max_ma = "max_ma",
+    min_ma = "min_ma",
+    scale = "GTS2020",
+    rank = "stage",
+    ext_orig = TRUE) {
+  # Handle errors
```
LewisAJones, line 50: Maybe some additional error handling could be useful here? I can imagine:
- max_ma and min_ma being mixed up
- negative values used for ages

R/tax_time_expand.R
```diff
@@ -0,0 +1,102 @@
...
+  if (is.data.frame(taxdf) == FALSE) {
+    stop("`taxdf` should be a dataframe")
+  }
+
+  if (!is.numeric(taxdf[, max_ma])) {
```
LewisAJones, line 55: Does this need to be numeric only? What if the user has interval names that can be matched to GTS2020/2012?

R/tax_time_expand.R
```diff
@@ -0,0 +1,102 @@
...
+#'   as the minimum limit (LADs) of the age range (e.g., "min_ma").
+#' @param scale \code{character}. Specify the desired geological timescale to
+#'   be used, either "GTS2020" or "GTS2012".
+#' @param rank \code{character}. Specify the desired stratigraphic rank. Choose
+#'   from: "stage", "epoch", "period", "era", and "eon".
+#' @param ext_orig \code{logical}. Should two additional columns be added to
+#'   identify the intervals in which taxa originated and went extinct?
+#'
+#' @return A \code{dataframe} where each row represents an interval during which
+#'   a taxon in the original user-supplied data was alive. The columns are
```
LewisAJones, line 28:
```suggestion
#'   a taxon in the original user-supplied data persisted. The columns are
```
Leave it or take it

R/tax_time_expand.R
```diff
@@ -0,0 +1,102 @@
+#' Expand taxa to interval-level pseudo-occurrences
+#'
+#' A function to generate interval-level pseudo-occurrences for taxa based on
+#' temporal ranges (e.g., the output of \code{\link{tax_range_time}}). While the
+#' resulting pseudo-occurrences should not be treated as equivalent to actual
+#' occurrence data (e.g., like that from the Paleobiology Database), such
+#' pseudo-occurrences may be useful for performing statistical analyses where
+#' the row representing a taxon must be replicated for each interval through
+#' which the taxon survived.
```
LewisAJones, line 9:
```suggestion
#' which the taxon persisted.
```
Leave it or take it

LewisAJones (changes requested):
Hey @willgearty! Thanks for this contribution! I think this is a simple, but really useful function to have in palaeoverse. Good job on getting it together so quickly.

I have very few suggestions... let me know if anything is unclear or you want to discuss.

Best,

Lewis

R/tax_time_expand.R
willgearty: Hmm...they could use look_up first, then use this function? (honestly I just didn't want to deal with that possibility)

LewisAJones: I think it would be better to have the function able to handle interval names directly (I believe we have done this in bin_time()). However, you're right, the user should be able to use the look_up function, and then call this function. Maybe some examples of this are sufficient.

LewisAJones (approved):
Hi Will, 

Thanks for implementing all these changes. I think this PR can be merged. It looks like the R CMD check issues are unrelated to this request. One final suggestion from me would be to include some reference to the look_up function in the documentation if users have interval names for the age range instead of numeric values. However, I'll leave that with you.

Thanks again for your work on this!

Lewis

PR #48, opened 2022-10-11
group_apply

LewisAJones:
# Function name
`group_apply`

## Description
A function to apply palaeoverse functionality across subsets (groups) of data, delineated using one or more variables. This function is a wrapper which can be used to run other palaeoverse functions on various groupings of data. For example, this enables the separate analysis of occurrences from different time intervals, spatial regions, or trait values. 

## Motivation and Context
Currently, this is most relevant to `tax_unique`. However, I can imagine future functions in which it would be useful to have implementation across groups (e.g. formation counts). 

## Notes
- Examples to be updated once `tax_unique` has been finalised.
- I am still on the fence whether this function should be included in the initial release.
- Currently, I am only allowing the input of certain functions from `palaeoverse` that will work. I can imagine this function being useful for other data types and functions outside of `palaeoverse`. However, I am not keen to open this box right now as it will delay progress.


KEichenseer (changes requested):
Hi Lewis, Bethany and Will,

The function works very well, nicely done! I have commented a few things in the code lines here. The main improvement I would like to see is related to the additional function arguments passed to `...` - currently, this only works if they are supplied in the correct order. 

It didn't seem to work with `tax_unique` yet (I only tried briefly), I assume we'll need to do a bit of testing once that function is finalised.

Kilian

bethany-j-allen:
I think it's overly generous to describe me as a developer on this one - I literally only created the branch and populated the R file. Feel free to remove me, or perhaps I can review it instead?

R/group_apply.R
```diff
@@ -0,0 +1,101 @@
+#' Apply a function over grouping(s) of data
+#'
+#' A function to apply palaeoverse functionality across subsets (groups) of
+#' data, delineated using one or more variables. See details for compatible
+#' functions.
+#'
+#' @param df \code{dataframe}. A dataframe of the fossil data you
```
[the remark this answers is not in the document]
LewisAJones: Updated accordingly!

R/group_apply.R
```diff
@@ -0,0 +1,101 @@
...
+#' \cr
+#' Available compatible `palaeoverse` functions:
+#' - \code{\link{tax_unique}}: return the number of unique taxa per grouping
+#' variable.
+#' - \code{\link{tax_range_time}}: return the temporal range of taxa per
+#' grouping variable.
+#' - \code{\link{tax_range_geo}}: return the geographic range of taxa per
+#' grouping variable.
+#' - \code{\link{tax_check}}: return potential spelling variations of the
+#' same taxon per grouping variable. Note: `verbose` should be set to FALSE.
```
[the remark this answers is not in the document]
LewisAJones: Done.

R/group_apply.R
```diff
@@ -0,0 +1,101 @@
...
+  # Apply function
+  results <- do.call(rbind, lapply(lst, fun))
+  # Get row names (groups)
+  grouping <- row.names(results)
+  # Clean up group names
+  grouping <- tools::file_path_sans_ext(grouping)
+  # Remove funky row names
+  row.names(results) <- NULL
+  # Add group column
+  results <- cbind.data.frame(results, grouping)
```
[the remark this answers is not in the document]
LewisAJones: Updated to return NULL if the number of rows == 0

R/group_apply.R
```diff
@@ -0,0 +1,101 @@
...
+  fun_name <- deparse(substitute(fun))
+  avail <- c("tax_unique", "tax_range_time", "tax_range_geo", "tax_check")
+  if (!fun_name %in% avail) {
+    stop("Specified function not available or compatible with group_apply.
+    Choose from:", paste(" ", avail))
+  }
+
+  # Update default arguments with supp arguments
+  supp_args <- list(...)
+  indx <- which(names(formals(fun)) %in% names(supp_args))
```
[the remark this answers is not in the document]
LewisAJones: Thanks Kilian, this was an important catch! Thanks. I've implemented both changes. I've avoided using your sapply approach by matching names and updating the order. It's basically the same thing though.

LewisAJones:
> I think it's overly generous to describe me as a developer on this one - I literally only created the branch and populated the R file. Feel free to remove me, or perhaps I can review it instead?

Hi Bethany, 
Sorry, I like to err on the side of acknowledging individuals! 
I've updated this to put you down as a reviewer. Please take a look when you have chance :-)

LewisAJones:
> Hi Lewis, Bethany and Will,
> 
> The function works very well, nicely done! I have commented a few things in the code lines here. The main improvement I would like to see is related to the additional function arguments passed to `...` - currently, this only works if they are supplied in the correct order.
> 
> It didn't seem to work with `tax_unique` yet (I only tried briefly), I assume we'll need to do a bit of testing once that function is finalised.
> 
> Kilian

Hi Kilian,

Thanks for the quick turnaround with this review! Helpful comments as always. I've now implemented all of your changes and also increased functionality a bit. User can now use any function that takes a data.frame as an input, and can be transformed into a data.frame as an output. This allows useful functionality like counting the number of rows per group (e.g. occurrences per collection or country).

R/group_apply.R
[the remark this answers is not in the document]
KEichenseer: Cool, matching names makes even more sense!

R/group_apply.R
```diff
@@ -0,0 +1,119 @@
+#' Apply a function over grouping(s) of data
+#'
+#' A function to apply `palaeoverse` functionality across subsets (groups) of
+#' data, delineated using one or more variables. `data.frame` related functions
+#' (e.g. `nrow`, `ncol`, `lengths`, `unique`) may also be used.
+#'
+#' @param occdf \code{dataframe}. A dataframe of fossil occurrences.
```
KEichenseer, line 7: not sure about calling it "occdf" - for `tax_check` it would rather be a taxdf. This can result in messages like

```
group_apply(occdf = occdf[,1:3],
group = c("collection_no"),
fun = tax_check,verbose = FALSE)
```

Error in FUN(X[[i]], ...) : 
  Please specify `name` as a single column name in `taxdf`

KEichenseer: I still wonder whether we could use "occdf" as the name for the df argument for taxdf

KEichenseer (commented):
Hi Lewis,

very cool that the function can now also take other dataframe-related functions!

It all seems to work well, including the `...` shenanigans :-)

I added a few comments mainly related to the documentation.

Kilian

KEichenseer (approved):

R/group_apply.R
LewisAJones: Hi Kilian, I think you got this error because you didn't update the `name` argument in tax_check. The default is the "genus" column. We could update taxdf to occdf in tax_check. It works with both. I think taxdf just makes it more clear to the user that it should contain taxonomic info?

R/group_apply.R
```diff
@@ -0,0 +1,119 @@
...
+#' appended column indicating the user-defined grouping (`grouping`).
+#'
+#' @details This function is a wrapper which can be used to run other
+#' `palaeoverse` functions on various groupings of data. `data.frame` related
+#' functions (e.g. `nrow`, `ncol`, `lengths`, `unique`) may also be used.
+#' For example, `group_apply` enables the separate analysis of occurrences from
+#' different time intervals, spatial regions, or trait values.
+#' \cr
+#' \cr
+#' Relevant `palaeoverse` functions:
```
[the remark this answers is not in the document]
LewisAJones: Technically other functions work, but it seems pointless to use them like this e.g. bin_lat, bin_time, etc

R/group_apply.R
```diff
@@ -0,0 +1,119 @@
...
+#' `occdf`. See details for compatible functions.
+#' @param ... Additional arguments available in the called
+#' function. These additional arguments may be required for function arguments
+#' without default values, or if you wish to overwrite the default argument
+#' value (see examples).
+#'
+#' @return A \code{dataframe} of the outputs from the selected function, with an
+#' appended column indicating the user-defined grouping (`grouping`).
+#'
+#' @details This function is a wrapper which can be used to run other
```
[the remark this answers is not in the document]
LewisAJones: I've moved the `group_apply` section to the beginning of the text, but I think it's good to keep the other details here too for sake of completeness.

R/group_apply.R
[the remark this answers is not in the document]
KEichenseer: Ok then it's fine

R/group_apply.R
LewisAJones: I've updated taxdf to occdf in group_apply :-)

R/group_apply.R
```diff
@@ -0,0 +1,119 @@
+#' Apply a function over grouping(s) of data
+#'
+#' A function to apply `palaeoverse` functionality across subsets (groups) of
+#' data, delineated using one or more variables. `data.frame` related functions
```
bethany-j-allen, line 4:
```suggestion
#' data, delineated using one or more variables. Functions which receive a `data.frame` as input
```

R/group_apply.R
bethany-j-allen, line 7:
```suggestion
#' @param occdf \code{dataframe}. A dataframe of fossil occurrences or taxa, as relevant to the desired function.
```

R/group_apply.R
```diff
@@ -0,0 +1,119 @@
...
+#' (e.g. `nrow`, `ncol`, `lengths`, `unique`) may also be used.
+#'
+#' @param occdf \code{dataframe}. A dataframe of fossil occurrences.
+#' This dataframe must contain the grouping variables and the necessary
+#' variables for the function you wish to call (see function specific
```
bethany-j-allen, line 9:
```suggestion
#' variables for the function you wish to call (see function-specific
```

R/group_apply.R
```diff
@@ -0,0 +1,119 @@
...
+#' documentation for required columns).
+#' @param group \code{character}. A vector of column names, specifying the
+#' grouping (e.g. "collection_no", "stage_bin").
```
bethany-j-allen, line 12:
```suggestion
#' desired subgroups (e.g. "collection_no", "stage_bin"). Supplying more than one grouping variable will produce an output containing subgroups for each unique combination of values.
```
Think my wording could be improved here but I think a sentence to this effect is a good idea.

R/group_apply.R
```diff
@@ -0,0 +1,119 @@
...
+#' @param fun \code{function}. The function you wish to apply to
+#' `occdf`. See details for compatible functions.
+#' @param ... Additional arguments available in the called
```
bethany-j-allen, line 15:
```suggestion
#' @param ... Additional arguments required by, or available to, the called
```
I think you could say this and delete the second sentence (maybe add 'see examples' to the end).

R/group_apply.R
```diff
@@ -0,0 +1,119 @@
...
+#' function. These additional arguments may be required for function arguments
+#' without default values, or if you wish to overwrite the default argument
+#' value (see examples).
+#'
+#' @return A \code{dataframe} of the outputs from the selected function, with an
+#' appended column indicating the user-defined grouping (`grouping`).
```
bethany-j-allen, line 21:
```suggestion
#' appended column indicating the user-defined groups (`grouping`).
```

R/group_apply.R
```diff
@@ -0,0 +1,119 @@
...
+#'
+#' @details `group_apply` enables the separate analysis of occurrences from
```
bethany-j-allen, line 23:
```suggestion
#' @details `group_apply` applies functions to subgroups of data within a supplied dataset, enabling the separate analysis of occurrences or taxa from
```

R/group_apply.R
```diff
@@ -0,0 +1,119 @@
...
+#' different time intervals, spatial regions, or trait values. The function
+#' serves as a wrapper around `palaeoverse` functions on groupings of
+#' data. `data.frame` related functions (e.g. `nrow`, `ncol`, `lengths`,
```
bethany-j-allen, line 26:
```suggestion
#' data, defined using one or more variables. Other functions which can be applied to a `data.frame` (e.g. `nrow`, `ncol`, `lengths`,
```

R/group_apply.R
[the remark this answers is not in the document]
bethany-j-allen: I agree that we need something to make this clearer. I guess the vignette will be useful for this. But perhaps we could also add a note to the documentation of the functions for which group_apply is relevant, and/or add an example?

R/group_apply.R
```diff
@@ -0,0 +1,119 @@
...
+#' Relevant `palaeoverse` functions:
+#' - \code{\link{tax_unique}}: return the number of unique taxa per grouping
+#' variable.
+#' - \code{\link{tax_range_time}}: return the temporal range of taxa per
+#' grouping variable.
+#' - \code{\link{tax_range_geo}}: return the geographic range of taxa per
+#' grouping variable.
+#' - \code{\link{tax_check}}: return potential spelling variations of the
+#' same taxon per grouping variable. Note: `verbose` needs to be set to FALSE.
+#'
```
bethany-j-allen, line 39: I would say 'group' is better than 'grouping variable' for the above section as the former fits better in the case of using multiple variables.

R/group_apply.R
```diff
@@ -0,0 +1,119 @@
...
+#' \cr
+#' Relevant `palaeoverse` functions:
+#' - \code{\link{tax_unique}}: return the number of unique taxa per grouping
+#' variable.
+#' - \code{\link{tax_range_time}}: return the temporal range of taxa per
+#' grouping variable.
+#' - \code{\link{tax_range_geo}}: return the geographic range of taxa per
+#' grouping variable.
+#' - \code{\link{tax_check}}: return potential spelling variations of the
+#' same taxon per grouping variable. Note: `verbose` needs to be set to FALSE.
```
bethany-j-allen, line 38: If this is necessary, could add an 'if' switch to ensure this is the case?

R/group_apply.R
```diff
@@ -0,0 +1,119 @@
...
+#' Kilian Eichenseer & Bethany Allen
+#' @importFrom stats as.formula
+#' @examples
+#' # Examples
+#' # Get tetrapods data
+#' occdf <- tetrapods
+#' # Remove NA data
+#' occdf <- subset(occdf, !is.na(genus))
+#' # Count number of occurrences from each country
+#' ex1 <- group_apply(occdf = occdf, group = c("cc"), fun = nrow)
```
bethany-j-allen, line 53: 'group = "cc"' also works here, perhaps would be cleaner? Especially as you're still giving an example using a vector below.

bethany-j-allen (commented):
Really nice job on this! As I mentioned, I had no idea how to write half of this, but it works well :) most of my comments are suggestions for the documentation. If I can help with the examples, especially if you want to add tax_unique, let me know!

R/group_apply.R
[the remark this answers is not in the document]
LewisAJones: As you've both highlighted this, I felt it was worth updating. It now reads as: 

All `palaeoverse` functions which require a dataframe input can be used in conjunction with the `group_apply` function. However, this is unnecessary for many functions (e.g. \code{\link{bin_time}}) as groups do not need to be partitioned before binning. This list provides users with `palaeoverse` functions that might be interesting to apply across groups:

R/group_apply.R
LewisAJones: Hmmm I think it's better to spell this out clearly here, particularly for overwriting default arguments.

R/group_apply.R
LewisAJones: Hmmm I would prefer to update the function to set FALSE as default behaviour. I can imagine having issues with many if switches as we accumulate functions! Thoughts?

R/group_apply.R
LewisAJones: OK!!

R/group_apply.R
```diff
@@ -0,0 +1,119 @@
...
+#' # Temporal range of taxa per time bin with default arguments
+#' ex2 <- group_apply(occdf = occdf, group = c("cc"), fun = tax_range_time)
+#' # Temporal range of taxa per time bin with updated arguments
+#' ex3 <- group_apply(occdf = occdf,
+#'                    group = c("cc"),
+#'                    fun = tax_range_time,
+#'                    name = "family") # Run at family level (default: "genus")
```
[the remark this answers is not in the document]
LewisAJones: OK!!

R/group_apply.R
```diff
@@ -0,0 +1,119 @@
...
+#' # Use multiple variables (number of occurrences per collection and country)
+#' ex4 <- group_apply(occdf = occdf,
+#'                    group = c("collection_no", "cc"),
```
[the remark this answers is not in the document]
LewisAJones: Updated to formation!

R/group_apply.R
```diff
@@ -0,0 +1,119 @@
...
+#'                    fun = nrow)
```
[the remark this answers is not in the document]
LewisAJones: Nice idea! Added!

R/group_apply.R
```diff
@@ -0,0 +1,119 @@
...
+    stop("`occdf` should be a dataframe")
+  }
+
+  if (!any(group %in% colnames(occdf))) {
+    stop("Supplied `group` is not a named column in `occdf`")
+  }
+
+  if (!is.function(fun)) {
+    stop("Supplied `fun` is not a function")
+  }
```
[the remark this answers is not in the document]
LewisAJones: Yes, error handling within the respective function will flag it!

R/group_apply.R
LewisAJones: OK!

R/group_apply.R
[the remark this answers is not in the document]
bethany-j-allen: Sounds good, perhaps we can add cross-references in the documentation of the relevant functions during the final clean-up and consistency checks.

R/group_apply.R
bethany-j-allen: Sure, whatever you think is most helpful!

R/group_apply.R
bethany-j-allen: Given that this would be the only switch at present, I'm not sure it really matters. Was there a reason behind setting the 'verbose' default to TRUE? That would be the only factor which would sway my opinion on this, I think.

R/group_apply.R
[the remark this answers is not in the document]
LewisAJones: They should already be cross-referenced! :-)

R/group_apply.R
LewisAJones: OK, let's leave it as it is. After checking tax_check, I think it is better to leave verbose set to TRUE purely because I don't see many users using tax_check with group_apply anyway...

willgearty:
I've made some changes to the handling of the supplemental arguments and the way groupings are reassigned to the output (now as separate columns if there were separate columns in the input). The latter actually seems to improve the performance of the function dramatically (nearly an order of magnitude) when multiple columns are specified.

Otherwise, everything looks good to me! Happy to address any reviews on the new code now that I'm back from GSA!

LewisAJones:
> I've made some changes to the handling of the supplemental arguments and the way groupings are reassigned to the output (now as separate columns if there were separate columns in the input). The latter actually seems to improve the performance of the function dramatically (nearly an order of magnitude) when multiple columns are specified.
> 
> Otherwise, everything looks good to me! Happy to address any reviews on the new code now that I'm back from GSA!

Hi Will, 

Thanks for the update, I think binding multiple groupings as separate columns is definitely a better way to go. I think your updates have caused an issue though... Several of the examples no longer seem to work and return NULL (ex1, ex4, ex5). From a quick look at the code, I think this is because you now force the users to pass additional arguments to `...` rather than updating the default ones when arguments are passed?

R/group_apply.R
```diff
@@ -0,0 +1,133 @@
...
+
+  supp_args <- list(...)
+  indx <- which(!(names(supp_args) %in% names(formals(fun))))
+  if (length(indx) > 1) {
```
LewisAJones, line 100:
```suggestion
  if (length(indx) > 0) {
```
Shouldn't this be 0 or >= 1 to detect any additional arguments?

R/group_apply.R
```diff
@@ -0,0 +1,133 @@
...
+  bin_codes <- as.formula(paste0("~ ", paste(group, collapse = " + ")))
+  # Split dataframe
+  lst <- split(occdf, f = bin_codes, drop = TRUE, sep = "&1*^$0%")
+  # Group names
+  nme <- names(lst)
+  # Split into individual columns
+  nme_df <- do.call(rbind.data.frame, strsplit(nme, "&1*^$0%", fixed = TRUE))
+  colnames(nme_df) <- group
+  # Apply function
+  output_lst <- lapply(lst, fun, ...)
```
LewisAJones, line 118: I think this is the problematic line for functions like `nrow`

R/group_apply.R
```diff
@@ -0,0 +1,133 @@
...
+  # Add groupings to output
+  output_df <- do.call(rbind.data.frame,
+                       lapply(seq_along(output_lst), FUN = function(i) {
+    df <- output_lst[[i]]
+    if (!is.null(nrow(df)) && nrow(df) > 0) {
```
LewisAJones, line 123: This could also be the problem as the function nrow does not return a data.frame?

R/group_apply.R
```diff
@@ -0,0 +1,133 @@
...
+  nme <- names(lst)
+  # Split into individual columns
+  nme_df <- do.call(rbind.data.frame, strsplit(nme, "&1*^$0%", fixed = TRUE))
+  colnames(nme_df) <- group
+  # Apply function
+  output_lst <- lapply(lst, fun, ...)
+  # Add groupings to output
+  output_df <- do.call(rbind.data.frame,
+                       lapply(seq_along(output_lst), FUN = function(i) {
+    df <- output_lst[[i]]
```
LewisAJones, line 122:
```suggestion
    df <- data.frame(output_lst[[I]])
```
This should fix things! But some updating of column names might be needed.

R/group_apply.R
```diff
@@ -0,0 +1,133 @@
...
+#' unique combination of values.
+#' @param fun \code{function}. The function you wish to apply to
+#' `occdf`. See details for compatible functions.
+#' @param ... Additional arguments available in the called
+#' function. These additional arguments may be required for function arguments
+#' without default values, or if you wish to overwrite the default argument
+#' value (see examples).
+#'
+#' @return A \code{dataframe} of the outputs from the selected function, with
+#' appended column(s) indicating the user-defined groups (`grouping`).
```
LewisAJones, line 25:
```suggestion
#' appended column(s) indicating the user-defined groups.
```
This can be removed now.

R/group_apply.R
```diff
@@ -0,0 +1,133 @@
...
+#' # Remove NA data
+#' occdf <- subset(occdf, !is.na(genus))
+#' # Count number of occurrences from each country
+#' ex1 <- group_apply(occdf = occdf, group = "cc", fun = nrow)
+#' # Temporal range of taxa per country with default arguments
+#' ex2 <- group_apply(occdf = occdf, group = "cc", fun = tax_range_time)
+#' # Temporal range of taxa per country with updated arguments
+#' ex3 <- group_apply(occdf = occdf,
+#'                    group = c("cc"),
+#'                    fun = tax_range_time,
```
LewisAJones, line 69: I think we should update `tax_range_time` to `tax_unique` in the examples. This will be more common usage?

R/group_apply.R
LewisAJones: I've put in a fix in for this, and added functionality that names the column based on the input function when a vector is returned. Let me know what you think!

willgearty: This is great, thanks @LewisAJones.

One thought about this functionality: if the user makes their own function within the `group_apply()` call, e.g.,:
```r
group_apply(occdf = occdf, group = c("cc"), fun = function(df) if (nrow(df) > 0) nrow(df))
```
then the entire text of the function is used for the column name:
image

Is this desired?

R/group_apply.R
willgearty: I don't have any issue with the examples using `tax_range_time`, but we should probably include an example with `tax_unique` (and/or add an example to `tax_unique` that uses `group_apply`)

R/group_apply.R
LewisAJones: I think that's fine, but maybe we should add something to the return section?

R/group_apply.R
LewisAJones: Adding this to my issue log to integrate next week!

willgearty:
This looks good to me! I'm happy to merge, unless there are any final comments?

PR #54, opened 2022-10-18
Review updates

LewisAJones:
Updates based on Google docs issue log.


R/bin_time.R
```diff
@@ -154,12 +154,12 @@ bin_time <- function(occdf, bins, method = "mid", reps = 100) {
 
     # Generate empty column for recording the number of bins an occurrence
     # appears in, and empty columns for the new bin allocation and midpoint.
-    occdf$n_bins <- NA
+    occdf$n_bin <- NA
```
willgearty, line 157: This can be kept as n_bins (since each individual value in the column refers to the number of bin**s**)

tests/testthat/test-palaeorotate.R
```diff
@@ -1,32 +1,24 @@
-test_that("palaeorotate() works", {
+test_that("palaeorotate() point method works", {
+  test <- try(exp = GET(
+    paste0("https://gws.gplates.org/reconstruct/reconstruct_points/",
+           "?points=95,54,142,-33&time=140&model=SETON2012")
+  ), silent = TRUE)
+  test <- grep("Error", test)
+  if (test == 1) {
+    skip("GPlates API not available.")
+  }
```
willgearty, line 9: Would you be able to just use `skip_if_offline()` instead? https://testthat.r-lib.org/reference/skip.html

willgearty (approved):

PR #66, opened 2022-12-02
Update rotation files and doc

### LINKED ISSUE #59: Reconstruction files
Currently, the reconstruction files are based on a 1º x 1º spatial grid. This should perhaps be updated to use a discrete equal-area grid. As implemented, points at high latitudes will be linked to reconstruction files at a higher geographic resolution than those at low latitudes.

LewisAJones:
This updates the palaeogeographic reconstruction files to use an equal-area grid of ~100 x 100 km, rather than a 1º x 1º grid.  These reconstruction files are now deposited on [Zenodo](https://zenodo.org/communities/palaeoverse/?page=1&size=20) on our new community data repository.

Fixes #59.


willgearty:
I'll take a more thorough look at the code later, but here are some benchmarks (using `microbenchmark`) for running paleorotate with the tetrapod data.
main:
```r
 expr                            min      lq      mean     median   uq       max         neval
 palaeorotate(occdf = tetrapods) 10.62838 11.8759 13.08635 12.71203 14.72765 16.24027    10
```
this PR:
```r
 expr                            min      lq       mean     median   uq       max         neval
 palaeorotate(occdf = tetrapods) 3.857538 4.083622 4.325168 4.244654 4.549429 5.004376    10
```

It's faster now!

R/palaeorotate.R
```diff
@@ -51,17 +51,17 @@
...
 #' rotation are required, or fine-scale spatial analyses are being conducted,
 #' use of the "point" `method` might be preferable for the user (particularly
 #' if occurrences are close to plate boundaries). As implemented, points within
 #' the same grid cell will be assigned equivalent palaeocoordinates due to
 #' spatial aggregation. The reconstruction files provide pre-generated
-#' palaeocoordinates for a grid of 1&deg; x 1&deg;, allowing the past
-#' distribution of fossil occurrences to be estimated efficiently. Access to
-#' the reconstruction files and documentation is available via the
-#' [palaeorotate](https://github.com/LewisAJones/palaeorotate) package.
+#' palaeocoordinates for a grid of ~100 km x 100 km, allowing the past
```
willgearty, line 61: I think this (and the zenodo documentation) needs to be updated to the hexagonal grid?

willgearty (approved):
Code and functionality are both great! Just one comment about the documentation!

R/palaeorotate.R
```diff
@@ -318,22 +320,32 @@ palaeorotate <- function(occdf, lng = "lng", lat = "lat", age = "age",
...
+                                     simple = TRUE)
+    model_cell <- h3jsr::point_to_cell(input = base_model_sf,
+                                       res = 3, # Grid resolution
+                                       simple = TRUE)
 
     # Generate row index
     pc_ind <- sapply(seq_len(nrow(occdf)), function(i) {
-      which(base_model[, c("lng")] == occdf[i, "rot_lng"] &
-              base_model[, c("lat")] == occdf[i, "rot_lat"])
+      which(model_cell == occ_cell[i])
```
willgearty, line 344: I'm not sure if it's faster, but I think you could use `match` here instead of `sapply` and `which`?

willgearty: Just checked, and it should be orders of magnitude faster.

willgearty: Another idea to speed things up would be to include the cell indices in the data files, so you don't need to do point_to_cell each time you use the function?

R/palaeorotate.R
```diff
@@ -318,22 +320,32 @@ palaeorotate <- function(occdf, lng = "lng", lat = "lat", age = "age",
...
       })]
 
-    # Search for matching longitude and  latitude
-    occdf$rot_lng <- sapply(seq_len(nrow(occdf)), function(i) {
-      # Extract closest longitude
-      base_model[which.min(abs(base_model[, c("lng")]  - occdf[i, lng])), 1]
-    }, simplify = TRUE)
+    # Set-up
+    # Convert to sf object and add CRS
+    occdf_sf <- sf::st_as_sf(x = occdf,
```
willgearty, line 325: st_as_sf and point_to_cell are now imported, so you don't need to specify the package and the double colon (but make sure to run document to add the imports to the NAMESPACE file).

LewisAJones: Done!

R/palaeorotate.R
LewisAJones: I've updated to use `match`, and it does seem to be a lot faster! Thanks! 

I agree that including the cell indices in the data files will likely speed things up. However, this would then require depositing another version of the data on Zenodo which I am a bit reluctant to do right now. Considering this function isn't really slowing things down a great deal, I'd prefer to keep it as it is. The addition of the cell indices in the data files is also likely to make them a bit heavier (increasing download time), so it probably doesn't make too much of a difference.

R/palaeorotate.R
LewisAJones: Done! Thanks for the catch.

R/palaeorotate.R
willgearty: Nice! (I updated the benchmarks above)

With regards to the cell indices, I agree it's fine to leave it as-is!

willgearty:
Updates look great, thanks!

PR #67, opened 2022-12-08
Fix look_up()

KEichenseer:
# Fix look_up()

## Description
The look_up function tries to look up intervals from the GTS tables and assign stages. This failed for intervals older than the Phanerozoic, as older stages are not defined in the GTS tables, resulting in an error. This error should be fixed with this pull request, and a test has been added.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,7 @@ Minor bug fixes
 * Fixed tests when suggested packages were not installed (#64)
 * Removed dependence of tests on divDyn and deeptime
 * Fixed tax_range_time example (#60)
+* Fixed look_up issue when handling pre-Cambrian occurrences.
 
 Added package-level documentation function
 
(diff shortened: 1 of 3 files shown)


LewisAJones (approved):
Hi Kilian,

This looks good to me and seems to do the job! A relatively minor change in the end. I think this can be integrated but try to clean up the new lint issues first.

Best,

Lewis

KEichenseer:
Thanks Lewis,

the lint issues are resolved now. 
A different change to fix this issue would have been to enable assigning Proterozoic periods with GTS, but I think it is better not to mix periods and stages in the output.

Kilian

LewisAJones:
> Thanks Lewis,
> 
> the lint issues are resolved now. A different change to fix this issue would have been to enable assigning Proterozoic periods with GTS, but I think it is better not to mix periods and stages in the output.
> 
> Kilian

I agree, but I think people will have different opinions on this. Let's see if we get any user feedback over the coming months. Unless there is anything else, I'll go ahead and merge this!

PR #69, opened 2022-12-20
Address #23: latitudinal binning and variable latitudinal range

### LINKED ISSUE #23: lat_bins() occurrence binning and variable latitudinal range
1. Currently, if occurrences fall on boundaries in lat_bins() they are assigned to the higher bin number. It could be useful to add functionality to allow occurrences to be binned into both bins.
2. Currently, lat_bins() only covers the complete latitudinal range (i.e. -90 to 90 latitude). It could be useful to add functionality to allow user input ranges (e.g. -60 to 60).

LewisAJones:
These updates enable users to:
- `lat_bins`: input a user-defined latitudinal range.
- `bin_lat`: bin occurrences into both latitudinal bins if the occurrence falls on the bin boundary. It does so by duplicating the occurrence (row).

Fixes #23.

### DIFF
--- a/R/bin_lat.R
+++ b/R/bin_lat.R
@@ -12,6 +12,11 @@
 #' @param lat \code{character}. The name of the column you wish to be treated
 #' as the input latitude (e.g. "lat" or "p_lat"). This column should contain
 #' numerical values. Defaults to "lat".
+#' @param boundary \code{logical}. If \code{TRUE}, occurrences
+#' falling on the boundaries of latitudinal bins will be duplicated and
+#' binded to the dataframe after being assigned to both bins.
+#' If \code{FALSE}, occurrences will be binned into the upper bin
+#' only (i.e. highest row number).
 #'
 #' @return A dataframe of the original input `occdf` with appended
 #' columns containing respective latitudinal bin information.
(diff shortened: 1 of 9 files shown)


SofiaGalv (commented):
Everything looks fine to me. I think allowing the user to define the bins is great!

R/bin_lat.R
```diff
@@ -12,6 +12,11 @@
 #' @param lat \code{character}. The name of the column you wish to be treated
 #' as the input latitude (e.g. "lat" or "p_lat"). This column should contain
 #' numerical values. Defaults to "lat".
+#' @param boundary \code{logical}. If \code{TRUE}, occurrences
+#' falling on the boundaries of latitudinal bins will be duplicated and
+#' binded to the dataframe after being assigned to both bins.
```
bethany-j-allen, line 17:
```suggestion
#' assigned to both bins.
```

bethany-j-allen (approved):
Looks great Lewis! I've made one grammatical suggestion but otherwise I think this is good to go.

LewisAJones:
Thanks both! You turned this around incredibly quickly. As these are not new functions and only expanding current functionality we can probably even get this into the next version :-)

PR #70, opened 2022-12-21
Enable user probability input

### LINKED ISSUE #20: User-defined probability function for bin_time()
Right now `bin_time()` uses a uniform distribution to assign point estimates for ages. However, I can see situations where users would prefer a normal distribution, logistic distribution, etc. It would be quite cool (although possibly complicated) to allow for this type of flexibility.

LewisAJones:
These changes enable users to input their own probability weights into the new argument `prob`. This allows point estimates to be generated following any probability distribution. I've tried to enable this flexibility in the cleanest way while also making it easy for the user. Please let me know if you have any thoughts or suggestions for further improving this as I guess it is a nice feature for the community.

Fixes #20.


R/bin_time.R
```diff
@@ -86,10 +100,12 @@
...
 #'
-#' #Assign point estimates based on fossil occurrence age range
-#' ex5 <- bin_time(occdf = occdf, bins = bins, method = "point", reps = 5)
+#' #Assign point estimates following a normal distribution
+#' ex5 <- bin_time(occdf = occdf, bins = bins, method = "point", reps = 5,
+#'                 fun = dnorm, mean = 0.5, sd = 0.25)
 #' @export
-bin_time <- function(occdf, bins, method = "mid", reps = 100) {
+bin_time <- function(occdf, bins, method = "mid", reps = 100,
+                     fun = NULL, ...) {
```
willgearty, line 108: Should we supply a default for `fun`? Maybe `dunif` to preserve the original default behavior of the function?

R/bin_time.R
```diff
@@ -211,15 +246,16 @@ bin_time <- function(occdf, bins, method = "mid", reps = 100) {
         #generate occurrence sequence for sampling
         occ_seq <- seq(from = occdf[i, "min_ma"],
                        to = occdf[i, "max_ma"],
-                       by = 0.01)
+                       by = 0.001)
+        #generate x for input probability function
+        x_prob <- seq(from = 0, to = 1, length.out = length(occ_seq))
+        # Generate probabilities
+        prob <- sapply(x_prob, fun, ...)
```
willgearty, line 253: Is there a reason we are using `sapply` here instead of just supplying `x_prob` as the first function to `fun`? I checked and it's an order of magnitude (or more) slower to use `sapply`.

NEWS.md
```diff
@@ -2,6 +2,7 @@
 Updates
 
 * Added package-level documentation function
+* bin_time now accepts user-defined probability weights
```
willgearty, line 5: This needs to be updated to reflect the new `fun` behavior.

willgearty (changes requested):
This appears to work well to me! I just have a few comments that should be addressed before merging.

R/bin_time.R
LewisAJones: Not really! Just a blunder from me. Updated!

LewisAJones:
Thanks for the review @willgearty! I've implemented your suggested changes, let me know if you have any other comments!

willgearty (approved):
Looks good, thanks!

PR #72, opened 2022-12-23
Enable user-input dataframe in time_bins

### LINKED ISSUE #19: User-defined timescale in time_bins()
It would be great to give more flexibility to the user to which timescale is being used for `time_bins()` (like `axis_geo()`). This could include incorporating other timescales through `deeptime::getTimeScale()` or by letting the user supply their own dataframe. I imagine this would be most useful for the equal-length binning aspect of `time_bins()`.

LewisAJones:
These updates allow users to input their own dataframe of a geological time scale into the `scale` argument, enabling near equal-length time bins to be generated. It is also compatible with time scales pulled from Macrostrat via `deeptime::getScaleData`.

Closes #19.

### DIFF
--- a/R/time_bins.R
+++ b/R/time_bins.R
@@ -32,7 +32,7 @@
 #'   geological timescale to be used "GTS2020" (default), "GTS2012" or a
 #'   user-input `data.frame`. If a `data.frame` is provided, it must contain
 #'   at least the following named columns: "interval_name", "max_ma", and
-#'   "min_ma". Column names "name", "max_age", and "min_ma" are also accepted,
+#'   "min_ma". Column names "name", "max_age", and "min_age" are also accepted,
 #'   but these are assumed to be equivalent to the aforementioned. As such,
 #'   age data should be provided in Ma.
 #' @param plot \code{logical}. Should a plot of time bins be generated?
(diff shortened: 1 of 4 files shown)


R/time_bins.R
```diff
@@ -130,68 +157,92 @@ time_bins <- function(interval = "Phanerozoic", rank = "stage",
...
+    }
+  } else {
+    # Assign input scale to df
+    df <- scale
+    # Update column names to work with getScaleData
+    colnames(df)[which(colnames(df) %in% c("name",
+                                           "max_age",
+                                           "min_age"))] <- c("interval_name",
+                                                             "max_ma",
+                                                             "min_ma")
```
willgearty, line 229: This assumes the data.frame comes from getScaleData and the columns are always in this order. The documentation implies that user-made data.frames can also have these column names, but they might not necessarily be in this order. Should we modify this bit of code to accommodate this?

R/time_bins.R
```diff
@@ -10,27 +10,31 @@
...
+#'   assign to requested bins. If assign is specified, a numeric vector is
+#'   returned of the midpoint age of the specified bins. Note this is the
+#'   simplified approach of assignment in `palaeoverse` included for data with
+#'   'known' point-age estimates. For a wider range of binning methods, see
+#'   \code{\link[palaeoverse:bin_time]{palaeoverse::bin_time()}}.
+#' @param scale \code{character} or \code{data.frame}. Specify the desired
+#'   geological timescale to be used "GTS2020" (default), "GTS2012" or a
+#'   user-input `data.frame`. If a `data.frame` is provided, it must contain
+#'   at least the following named columns: "interval_name", "max_ma", and
+#'   "min_ma". Column names "name", "max_age", and "min_ma" are also accepted,
```
willgearty, line 35:
```suggestion
#'   "min_ma". Column names "name", "max_age", and "min_age" are also accepted,
```

willgearty (changes requested):
Looks great, just a few comments!

R/time_bins.R
LewisAJones: Good catch, thanks!

R/time_bins.R
LewisAJones: Definitely, I've now implemented a fix for this. Let me know what you think!

willgearty (approved):

Issue #51, opened 2022-10-13
tax_unique() output which lists occurrences


bethany-j-allen:
**Is your feature request related to a problem? Please describe.**
tax_unique() currently provides a list of 'unique' taxon names, and does not utilise occurrence information.

**Describe the solution you'd like**
An alternative output format which provides occurrence information as nested lists within each 'unique' taxon might be useful.

willgearty:
Can't this be accomplished by using group_apply()?

willgearty:
Ah, I think I was misunderstanding the goal here. You could use `group_apply` for collections to get the different unique taxa within collections, but I agree, if you want the overall list of unique taxa and then the lists of occurrences for those taxa, `group_apply` would not work. I like the idea of just adding the unique name back on to the original df.

PR #73, opened 2023-01-06
tax_unique output with original occurrences

willgearty:
This provides the option to have `tax_unique` return the original occurrence dataframe with the unique names appended as a new column. I decided to go with @LewisAJones's suggestion in #51, but @bethany-j-allen's suggestion (returning the default df but with row numbers specified) is an intermediate product of the new code, so if the reviewers would prefer, I could make that the returned value (or make both possible return values by modifying how `orig` works).

It should be noted that this option results in the function taking about 1.5 times as long to process the tetrapods dataset, but ultimately it's only about 50 extra milliseconds, so not too bad of a performance hit. On smaller datasets the time increase isn't even noticeable.

Fixes https://github.com/palaeoverse-community/palaeoverse/issues/51.


R/tax_unique.R
```diff
@@ -29,11 +29,16 @@
 #' occurrence dataframe is of this type.
 #' @param resolution \code{character}. The taxonomic resolution at which to
 #' identify unique occurrences, either "species" (the default) or "genus".
+#' @param orig \code{logical}. Should the original dataframe be returned with
```
LewisAJones, line 32: Probably doesn't matter too much, but maybe we should change the name of this argument? Maybe something like bind? I only say this because we have a similar argument name for origin in another function.

LewisAJones, line 32: I can't seem to get the new functionality to work with different resolutions, for example:

```
occdf <- tetrapods[1:100, ]
genera_orig <- tax_unique(occdf = occdf, genus = "genus", family = "family",
                     order = "order", class = "class", resolution = "genus", orig = TRUE)
```

R/tax_unique.R
```diff
@@ -121,7 +126,7 @@
 #'
 tax_unique <- function(occdf = NULL, binomial = NULL, species = NULL,
                        genus = NULL, ..., name = NULL,
-                       resolution = "species") {
+                       resolution = "species", orig = FALSE) {
```
LewisAJones, line 129: Maybe we can add an example with the new functionality?

R/tax_unique.R
```diff
@@ -262,7 +267,24 @@ tax_unique <- function(occdf = NULL, binomial = NULL, species = NULL,
...
-  occurrences <- unique(occurrences)
+  if (orig) {
+    occurrences_with_dupes <- occurrences
+    # List the rows of the original dataframe that correspond to each unique row
+    # Hack to include NAs as groups
+    occurrences_with_dupes[is.na(occurrences_with_dupes)] <- "thisisanNAvalue"
+    occurrences_with_dupes$rows <- seq_len(nrow(occurrences_with_dupes))
+    form <- as.formula(
+      paste(". ~",
+            paste(c(rev(higher_names), "genus", "genus_species"),
```
LewisAJones, line 278: I think this is the line causing issues with resolution... "genus_species" will not always be present in the dataset?

LewisAJones (changes requested):
Hi @willgearty! This is looking pretty good. It seems to work well with species resolution but hits an issue when trying to use `resolution = "genus"`. See below for specific comments!

LewisAJones (approved):
Hi @willgearty, this is looking good to me now! The function works at different resolutions and append seems a better argument name IMO. Thanks for getting this together so quickly!

One last thing, I'm getting the following when running checks:

```
 tax_unique: no visible global function definition for ‘aggregate’
  Undefined global functions or variables:
    aggregate
  Consider adding
    importFrom("stats", "aggregate")
  to your NAMESPACE file.
```
So I think you need to add aggregate? Other than that, good to go!

R/tax_unique.R
```diff
@@ -29,11 +29,17 @@
...
+#' the unique names appended as a new column?
 #'
 #' @return A \code{dataframe} of taxa, with each row corresponding to a unique
 #' "species" or "genus" in the dataset (depending on the chosen resolution).
 #' The dataframe will include the taxonomic information provided into the
-#' function, as well as a column providing the 'unique' names of each taxon.
+#' function, as well as a column providing the 'unique' names of each taxon. If
+#' \code{append} is \code{TRUE}, the original dataframe (\code{occdf}) will be
+#' returned with these 'unique' names appended as a new column. Occurrences that
+#' are identified to a coarser taxonomic resolution which is already represented
```
bethany-j-allen, line 41:
```suggestion
#' are identified to a coarse taxonomic resolution, and belong to a clade which is already represented
```

R/tax_unique.R
```diff
@@ -29,11 +29,17 @@
...
+#' within the dataset will have \code{NA} for their 'unique' names.
```
bethany-j-allen, line 42:
```suggestion
#' within the dataset, will have their 'unique' names listed as \code{NA}.
```

R/tax_unique.R
```diff
@@ -90,6 +96,7 @@
 #' @section Reviewer(s):
```
bethany-j-allen, line 96: It won't let me select it, but
"The function matches taxonomic names at increasingly higher taxonomic levels, so homonyms may be falsely filtered out."
isn't true anymore. Perhaps change to
"The function matches taxonomic names at face value, so homonyms may be falsely filtered out."

bethany-j-allen (approved):
Thanks a lot for doing this Will! The outputs are looking good. I have a couple of suggestions for the documentation but otherwise I'm happy to approve this.

PR #75, opened 2023-02-04
update documentation and add MULLER2022

### LINKED ISSUE #74: Rotation (MULLER2019)
Documentation for the MULLER2019 model has been updated on [GPlates Web Service](https://gwsdoc.gplates.org/reconstruction-models#muller2019) stating that the MULLER2019 model covers 0--250 Ma (as the paper also states). However, the API service allows points to be reconstructed up to 540 Ma for this model. A little digging required... 

This should be addressed for v1.1.1.

As a side note, the MULLER2022 model is also now available and perhaps should be incorporated down the line.

LewisAJones:
This PR updates documentation to be consistent with [GPlates Web Service](https://gwsdoc.gplates.org/reconstruction-models) after recent documentation changes. The MULLER2022 model has also now been made available via `palaeorotate()`. Reconstruction files are currently being rebuilt and will be deposited on [Zenodo](https://zenodo.org/record/7390066#.Y940oy8w1QI) once finished. Until then, this PR will fail R CMD Checks.

Closes #74.


LewisAJones:
Hey @willgearty and @Buffan3369! 

I've now finished adding the MULLER2022 rotations and updated the documentation accordingly. I've also updated the `uncertainty` argument to take a vector of model names. This allows a user to compare palaeogeographic uncertainty between specific models. I've also added a warning message for when a user tries to use plate rotation models with a mantle reference frame which is not recommended for reconstructing palaeocoordinates (documentation has also been added for this).

Hopefully changes are quite clear!

Buffan3369 (commented):
Everything goes fine, I'm just wondering why you wrote the same `if` statement twice in your code (line 244 and 250). Probably a typo

R/palaeorotate.R
```diff
@@ -218,12 +236,34 @@ palaeorotate <- function(occdf, lng = "lng", lat = "lat", age = "age",
...
+  }
+
+  # Add warnings for use of mantle reference frame models
+  if (any(model %in% c("MULLER2022", "MULLER2019",
+                   "MATTHEWS2016_mantle_ref"))) {
+    warning(paste0("Selected model(s) use a mantle reference frame and are ",
+    "not recommended for reconstructing palaeocoordinates. See details."))
+  }
+
+  if (any(uncertainty %in% c("MULLER2022", "MULLER2019",
```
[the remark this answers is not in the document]
LewisAJones: The first if statement is checking the `model` input argument and the second is checking the `uncertainty` input argument. This could be collapsed into a single if statement but I am not sure it makes too much of a difference.

LewisAJones:
> Everything goes fine, I'm just wondering why you wrote the same `if` statement twice in your code (line 244 and 250). Probably a typo

Thanks Lucas! Addressed above.

R/palaeorotate.R
```diff
@@ -92,6 +96,11 @@
 #'   - 0--540 Ma (grid)
 #'   - 0--750 Ma (point)
 #'
+#' While access is provided to all Global Plate Models available via the
+#' GPlates API, the use of the following mantle reference frame models is not
+#' recommended for reconstructing coordinates:
+#' "MULLER2022", "MULLER2019", and "MATTHEWS2016_mantle_ref".
```
willgearty, line 102: Given this, would it make sense to put the other ones first in the above documentation? I realize it's currently chronological, but it might make more sense to try to bring the recommended models to the forefront?

R/palaeorotate.R
```diff
@@ -92,6 +96,11 @@
...
+#'
 #' @section References:
```
willgearty, line 104: Should these references be in the same order as the models?

R/palaeorotate.R
```diff
@@ -281,7 +318,8 @@ palaeorotate <- function(occdf, lng = "lng", lat = "lat", age = "age",
     }
     # Reconstruction files
     rot_files <- list(
-      BASE = "https://zenodo.org/record/7390066/files/",
+      BASE = "https://zenodo.org/record/7615203/files/",
```
willgearty, line 321: Is there a way to always link to the most recent version of the zenodo repo? (I know you can do this with the doi, but I don't know whether it's possible for files)

NEWS.md
```diff
@@ -1,7 +1,9 @@
-# palaeoverse 1.1.0.9...
+# palaeoverse 1.1.1
 
 * palaeoverse now requires deeptime (>= 1.0.0)
 * Added additional test to catch when Zenodo is offline (palaeorotate)
+* palaeorotate documentation updated and MULLER2022 model made available
+* uncertainty argument in palaeorotate updated to enable user-input
```
willgearty, line 6: Make sure to update the news accordingly with whatever changes are made based on my comments!
(and you should probably cite the issue that you are fixing)

willgearty (changes requested):
I've got a few comments, some of which may take more time than others. Let me know if you have any questions!

R/palaeorotate.R
LewisAJones: I don't think so. Each version of the data has a unique DOI so you can identify which version of the data you are working on/with.

R/palaeorotate.R
LewisAJones: This section has been updated. Let me know if you still have concerns!

R/palaeorotate.R
LewisAJones: This section has been updated. Let me know if you still have concerns!

R/palaeorotate.R
LewisAJones: I think keeping them in alphabetical order is fine (I guess this is just a style preference).

R/palaeorotate.R
```diff
@@ -22,12 +22,14 @@
...
-#' Circle Distance between palaeocoordinates (in km). This argument is only
-#' relevant if `method` is set to "grid".
+#' @param uncertainty \code{NULL} or \code{character}. The names of the plate
+#' rotation models to calculate palaeogeographic uncertainty for.
+#' Palaeocoordinates from all reconstruction files (models) will be returned,
+#' along with their respective palaeolatitudinal range and the maximum Great
+#' Circle Distance between palaeocoordinates (in km).
+#' By default, this argument is set to NULL (default) and palaeogographic
+#' uncertainty is not calculated. This argument is only relevant if `method`
+#' is set to "grid".
```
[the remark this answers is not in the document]
LewisAJones: I have implemented point 1 and 3. For point 2, I think `uncertainty` should remain as `FALSE` by default. The reason for this is that I imagine most people will only want to use one model at a time in the palaeorotate function. It would be tiresome to continuously set the argument to `FALSE`. Of course we could ignore this argument when only one model is provided, but I think this is unnecessary.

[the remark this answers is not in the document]
willgearty: My thought for #2 was indeed that we could just ignore the argument if only one model is specified.

R/palaeorotate.R
willgearty: Looks like `httr::GET("https://doi.org/10.5281/zenodo.7390065")$url` will always give you the most recent base URL for the files.

willgearty: `httr::GET("https://zenodo.org/record/7390065")$url` works too

R/palaeorotate.R
```diff
@@ -143,7 +157,7 @@
 #' @section Developer(s):
 #' Lewis A. Jones
 #' @section Reviewer(s):
-#' Kilian Eichenseer & Lucas Buffan
+#' Kilian Eichenseer, Lucas Buffan & Will Gearty
 #' @importFrom geosphere distm distHaversine
```
willgearty, line 161: I think you need to update this line?

willgearty, line 161: I think you need to update this line?

R/palaeorotate.R
LewisAJones: Implemented! Thanks!

R/palaeorotate.R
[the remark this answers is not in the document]
LewisAJones: Implemented!

willgearty (changes requested):
The new changes look great @LewisAJones! I just have a couple very minor comments:

R/palaeorotate.R
```diff
@@ -454,98 +423,151 @@ palaeorotate <- function(occdf, lng = "lng", lat = "lat", age = "age",
...
+    cnames <- c("p_lng", "p_lat")
+  }
+  if (any(is.na(occdf[, unlist(cnames)]))) {
+    warning(
+      paste0(
+      "Palaeocoordinates could not be reconstructed for all points.",
+        "\n",
+        "Either assigned plate does not exist at time of ",
+        "reconstruction or the plate rotation model(s) does not cover ",
+        "the age of reconstruction."
```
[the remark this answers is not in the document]
willgearty: Looks like it might be the second test with PALEOMAP and SETON2012?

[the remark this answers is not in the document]
LewisAJones: This was intentionally for code coverage. However, now you've flagged it, it does seem undesirable to have such behaviour. I've removed this test.

LewisAJones:
> The new changes look great @LewisAJones! I just have a couple very minor comments:

Thanks @willgearty! Changes implemented!

willgearty (approved):

PR #76, opened 2023-02-14
add coords argument

LewisAJones:
This PR adds an additional argument to `tax_range_space` which allows the user to control whether coordinates are returned for the "gcd" and "con" `method`.


tests/testthat/test-tax_range_space.R
```diff
@@ -19,19 +19,19 @@ test_that("tax_range_soace() works", {
     unique_taxa)
 
   # Expect greater than
```
willgearty, line 21: This comment should be adjusted

willgearty (changes requested):
Looks good! Just one very minor comment, then it can be merged!

tests/testthat/test-tax_range_space.R
LewisAJones: Done! Thanks!

willgearty (approved):

PR #77, opened 2023-02-17
Remove deeptime dependence

willgearty:
In playing around with the [pkgndep package](https://cran.rstudio.com/web/packages/pkgndep/index.html), I realized that deeptime has quite a few dependencies (due to its interactions with ggplot and various phylogenetics packages). We currently depend on deeptime for the `axis_geo` function, but it's not really necessary (and certainly not worth the 39 extra dependencies). 

Here I've copied over `get_scale_data` (called `get_timescale_data` in palaeoverse) and adapted it for our needs. It uses the built-in GTS2020 and GTS2012 instead of the deeptime datasets, can filter to specific ranks, and still can pull from Macrostrat. I've updated `axis_geo` to use this new function and have also made sure that it remains backwards compatible with our examples and users' existing code.

While I was at it, I updated axis_geo to follow the variable naming scheme and British styling of the rest of the package. deeptime-formatted data is still acceptable, but is now the alternative rather than the expected.

Finally, I noticed that we weren't using the `font` column of the GTS datasets, so that is now fixed (in addition to fixing some of the `font` values in both datasets).

Also, after implementing all of this, I realized that a bit of this functionality (extracting GTS2012/2020 data) overlaps with some of the functionality of `time_bins`. Should I remove this functionality from `get_timescale_data` in favor of `time_bins`?

Comment: our next largest dependency hog is h3jsr (26 additional dependencies)


R/get_timescale_data.R
```diff
@@ -0,0 +1,107 @@
...
+#' @param name \code{character}. The name of the desired timescale.
+#' @param rank \code{character}. The name of the desired rank (if `name`
+#'   specifies a built-in timescale).
+#' @return A `data.frame` with at least the following columns:
+#'   \item{name}{the names of the time intervals.}
+#'   \item{max_age}{the oldest boundaries of the time intervals, in millions of
+#'     years.}
+#'   \item{min_age}{the youngest boundaries of the time intervals, in millions
+#'     of years.}
+#'   \item{abbr}{either traditional abbreviations of the names of the time
```
LewisAJones, line 22: I think this is currently only true for Macrostrat timescales?

R/get_timescale_data.R
```diff
@@ -0,0 +1,107 @@
+#' Get geological timescale data
+#'
+#' This function takes a name of a geological timescale and returns data for the
+#' timescale. Valid names include those of built-in `data.frames` ([GTS2020()]
```
LewisAJones, line 4:
```suggestion
#' timescale. Valid names include those of built-in `data.frame`(s) ([GTS2020()]
```
Maybe? I guess we normally put backticks just around classes/functions/etc?

R/get_timescale_data.R
```diff
@@ -0,0 +1,107 @@
+#' Get geological timescale data
+#'
+#' This function takes a name of a geological timescale and returns data for the
```
LewisAJones, line 3: There are a few minor lint issues (e.g. line length) to tidy up

R/get_timescale_data.R
```diff
@@ -0,0 +1,107 @@
...
+#' @return A `data.frame` with at least the following columns:
+#'   \item{name}{the names of the time intervals.}
+#'   \item{max_age}{the oldest boundaries of the time intervals, in millions of
+#'     years.}
+#'   \item{min_age}{the youngest boundaries of the time intervals, in millions
+#'     of years.}
+#'   \item{abbr}{either traditional abbreviations of the names of the time
+#'     intervals (if they exist) or custom abbreviations created with R.}
+#'   \item{color}{hex color codes associated with the time intervals (if
+#'     applicable).}
```
LewisAJones, line 25: Is it worth including a details section stating which timescales are available via Macrostrat? I guess in comparison to the GTS timescales we also don't have much information about what these timescales are based on or where they are coming from?

R/get_timescale_data.R
```diff
@@ -0,0 +1,107 @@
...
+#' @importFrom utils read.csv
+#' @importFrom curl nslookup
+#' @importFrom grDevices col2rgb
+#' @export
```
LewisAJones, line 29: We should include a couple of examples here

R/get_timescale_data.R
```diff
@@ -0,0 +1,107 @@
...
+get_timescale_data <- function(name = "GTS2020", rank = NULL) {
+  possible_names <- c("GTS2020", "GTS2012")
+  name_match <- charmatch(name, possible_names)
```
LewisAJones, line 32: I generally like using charmatch but it seems overdoing it here... it's only really going to catch "GTS201" and "GTS202"?

R/get_timescale_data.R
```diff
@@ -0,0 +1,107 @@
...
+  if (!is.na(name_match)) {
+    if (name_match == 0) {
+      stop("'name' matches both built-in scales. Please be more specific.",
```
LewisAJones, line 35:
```suggestion
      stop("`name` matches both built-in scales. Please be more specific.",
```
Elsewhere we use backticks for arguments

R/get_timescale_data.R
```diff
@@ -0,0 +1,107 @@
...
+      name <- possible_names[name_match]
+      if (name == "GTS2020") {
+        dat <- palaeoverse::GTS2020
+      } else if (name == "GTS2012") {
+        dat <- palaeoverse::GTS2012
+      }
+    }
+    if (!is.null(rank)) {
+      possible_ranks <- c("period", "epoch", "stage", "era", "eon")
+      rank_match <- charmatch(rank, possible_ranks)
```
LewisAJones, line 48: Definitely more appropriate here! Though I do have mix feelings about this. I guess one of the things we want to achieve with palaeoverse is to make code more readable. Do we enable users to be less specific/spell incorrectly or do we reinforce readability?

R/get_timescale_data.R
```diff
@@ -0,0 +1,107 @@
...
+    )
+    macrostrat_url <- url(paste0("https://macrostrat.org/api/v2/defs/intervals",
+                                 "?format=csv&timescale=",
+                                 gsub(" ", "%20", name)))
+    raw_dat <- tryCatch(
+      {
+        read.csv(macrostrat_url, header = TRUE, stringsAsFactors = FALSE)
+      },
+      error = function(e) {
+        stop("'name' does not match a built-in or Macrostrat timescale.",
```
LewisAJones, line 83:
```suggestion
        stop("`name` does not match a built-in or Macrostrat timescale.",
```

R/get_timescale_data.R
```diff
@@ -0,0 +1,107 @@
+#' Get geological timescale data
+#'
+#' This function takes a name of a geological timescale and returns data for the
+#' timescale. Valid names include those of built-in `data.frames` ([GTS2020()]
+#' and [GTS2012()]) and exact matches to those hosted by Macrostrat (see full
+#' list here: <https://macrostrat.org/api/defs/timescales?all>). If a built-in
```
LewisAJones, line 6: Maybe in the details section we can mention that GTS in Macrostrat is always the most recent? I am also wondering if users should also just be able to enter "Stages", "Periods", etc in name?

R/axis_geo.R
```diff
@@ -3,37 +3,35 @@
 #' \code{axis_geo} behaves similarly to \code{\link[graphics]{axis}} in that it
```
LewisAJones, line 3: I think there are some lint issues with line length, etc?

R/axis_geo.R
```diff
@@ -3,37 +3,35 @@
...
-#' also available from the \code{deeptime} package (see
-#' \code{\link[deeptime]{get_scale_data}}). A custom interval dataset can also
-#' be provided (see Details below). The appearance of the axis is highly
-#' customizable (see Usage below), with the intent that plots will be
+#' The default scale includes international epochs from the the Geological
+#' Timescale 2020 (\code{\link{GTS2020}}). However, international periods,
+#' stages, eons, and eras are also available. Interval data hosted by Macrostrat
+#' are also available (see \code{\link{get_timescale_data}}). A custom interval
+#' dataset can also be used (see Details below). The appearance of the axis is
+#' highly customizable (see Usage below), with the intent that plots will be
```
LewisAJones, line 11:
```suggestion
#' highly customisable (see Usage below), with the intent that plots will be
```

R/axis_geo.R
```diff
@@ -49,33 +47,39 @@
...
-#'   \code{data.frame} (see \code{\link[deeptime]{get_scale_data}}), or B) a
+#'   a \code{character} string indicating a rank of intervals from the built-in
+#'   \code{\link{GTS2020}}, B) a \code{character} string indicating a remotely
+#'   hosted \code{data.frame} (see \code{\link{get_timescale_data}}), or C) a
 #'   custom \code{data.frame} of time interval boundaries (see Details).
 #' @param height \code{numeric}. The relative height (or width if \code{side} is
 #'   \code{2} or \code{4}) of the scale. This is relative to the height (if
 #'   \code{side} is \code{1} or \code{3}) or width (if \code{side} is \code{2}
 #'   or \code{4}) of the plot.
 #' @param fill \code{character}. The fill color of the boxes. The default is to
```
LewisAJones, line 68:
```suggestion
#' @param fill \code{character}. The fill colour of the boxes. The default is to
```
I think you're trying to standardise?

R/axis_geo.R
```diff
@@ -49,33 +47,39 @@
...
+#'   use the \code{font} or \code{lab_color} column included in
 #'   \code{intervals}. If a custom dataset is provided with \code{intervals}
-#'   without a \code{lab_color} or \code{lab_colour} column and without
-#'   specifying \code{lab_col}, all labels will be black. Custom label colors
-#'   can be provided with this option (overriding the \code{lab_color} or
-#'   \code{lab_colour} column) and will be recycled if/as necessary.
+#'   without a \code{font} or \code{lab_color} column and without
+#'   specifying \code{lab_col}, all labels will be black. Custom label colours
+#'   can be provided with this option (overriding the \code{font} or
+#'   \code{lab_color} column) and will be recycled if/as necessary.
```
LewisAJones, line 82: Does "color" need changing to "colour" here too?

R/axis_geo.R
```diff
@@ -404,77 +407,86 @@ axis_geo <- function(
...
-      "name"
-    colnames(scale_intervals)[colnames(scale_intervals) == "max_ma"] <-
-      "max_age"
-    colnames(scale_intervals)[colnames(scale_intervals) == "min_ma"] <-
-      "min_age"
-    colnames(scale_intervals)[colnames(scale_intervals) == "colour"] <-
-      "color"
-    colnames(scale_intervals)[colnames(scale_intervals) == "lab_colour"] <-
-      "lab_color"
+    # fix column names if using deeptime data
```
LewisAJones, line 420: Is this required here? Isn't this handled in `get_timescale_data`?

tests/testthat/test-get_timescale_data.R
```diff
@@ -0,0 +1,22 @@
+test_that("get_timescale_data works", {
+  expect_named(get_timescale_data(rank = "stage"),
```
LewisAJones, line 2: Do we want a skip if macrostrat is offline here?

tests/testthat/test-get_timescale_data.R
```diff
@@ -0,0 +1,22 @@
...
+  expect_named(
+    get_timescale_data("North American Land Mammal Ages"),
+    c("interval_name", "max_ma", "min_ma", "abbr", "colour", "mid_ma",
+      "duration_myr", "font")
+  )
+
+  expect_error(get_timescale_data(name = "GTS"))
+  expect_error(get_timescale_data(rank = "pickles"))
+  expect_error(get_timescale_data(rank = "e"))
+  expect_error(get_timescale_data("international house of pancakes"))
```
LewisAJones, line 21: Now I'm hungry...

LewisAJones (changes requested):
Hey Will! Thanks for putting this together. I think it is important that we have periodic 'house cleaning' to remove dependencies where possible and this is definitely a good function to start with. 

I have quite a few comments but my main concern is that `get_timescale_data` and `time_bins` now seem very similar to me! It makes me think that we should perhaps just merge these two functions?

R/get_timescale_data.R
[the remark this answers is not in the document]
willgearty: It looks like raw URLs are [not terribly accessible for screen readers](http://www.nysed.gov/webaccess/create-accessible-hypertext-links), so we might want to do a pass over the full package at some point.

R/get_timescale_data.R
willgearty: I'm not sure what we'd really say about them. We have no control over which ones are there, and as you say, we don't have much information about them beyond what is accessible in the API. We could reach out to the Macrostrat folks and try to get references, but ultimately, this seems like it would be a pretty fragile bit of documentation.

R/get_timescale_data.R
willgearty: Hmm...I think it's fine as is. I think most people will use the full words anyways?

R/get_timescale_data.R
willgearty: Hmm...maybe...I'd be fine listing a few Macrostrat timescales that we think are useful, but I don't think we should get too detailed because things can change without much notice.

R/axis_geo.R
willgearty: No, because that's referring to the deeptime standards

willgearty: "font" is the palaeoverse standard

R/axis_geo.R
willgearty: This handles deeptime-formatted timescales

LewisAJones:
`time_bins` has now been updated to incorporate functionality for accessing time scales from Macrostrat. During these updates I've also spent a bit of time tidying up and streamlining the function. Let me know if you have any comments @willgearty. Once you're happy with `time_bins`, I can go over `axis_geo` again. If you're short on time, perhaps @KEichenseer can also review the most recent updates. As one of the other original reviewers, he is familiar with the function too.

I haven't removed`get_timescale_data` for now in case you were wanting to salvage anything from it.

tests/testthat/test-time_bins.R
```diff
@@ -1,5 +1,6 @@
 test_that("time_bins() works", {
-
+  # Skip if offline
+  skip_if_offline(host = "macrostrat.org")
```
willgearty, line 3: We should either put the macrostrat related tests at the very end and only skip those or put them all within a separate `test_that()`.

R/time_bins.R
```diff
@@ -29,37 +31,36 @@
...
 #' @importFrom stats sd
+#' @importFrom curl nslookup
 #'
 #' @return A \code{dataframe} of time bins for the specified intervals or a
-#' list with a \code{dataframe} of time bins and a named \code{numeric}
-#' vector (bin number) of binned age estimates (midpoint of specified bins) if
-#' assign is specified.
+#'   list with a \code{dataframe} of time bins and a named \code{numeric}
+#'   vector (bin number) of binned age estimates (midpoint of specified bins)
+#'   if assign is specified.
```
willgearty, line 48:
```suggestion
#'   if `assign` is specified.
```

R/time_bins.R
```diff
@@ -29,37 +31,36 @@
...
+#'   user-input `data.frame` (see `scale` argument) to generate time bins.
+#'   Additional information on included Geological Time Scales and source can
+#'   be accessed via:
+#'   - \code{\link{GTS2020}}
+#'   - \code{\link{GTS2012}}
 #'
-#' Available intervals names are accessible via the `interval_name`
-#' column in `GTS2012` and `GTS2020`. Data of the Geological Timescale 2020
-#' and 2012 were compiled by Lewis A. Jones (2022-07-02).
+#'   Available intervals names are accessible via the `interval_name` column
```
willgearty, line 59:
```suggestion
#'   Available interval names are accessible via the `interval_name` column
```

R/time_bins.R
```diff
@@ -29,37 +31,36 @@
...
-#'   "min_ma". Column names "name", "max_age", and "min_age" are also accepted,
-#'   but these are assumed to be equivalent to the aforementioned. As such,
-#'   age data should be provided in Ma.
+#'   "min_ma". As such, age data should be provided in Ma.
 #' @param plot \code{logical}. Should a plot of time bins be generated?
 #' @importFrom graphics polygon title
 #' @importFrom stats sd
+#' @importFrom curl nslookup
 #'
 #' @return A \code{dataframe} of time bins for the specified intervals or a
```
willgearty, line 45: Should we include information about the columns that can be expected?

R/time_bins.R
```diff
@@ -94,161 +95,185 @@
...
-    # Order by mid_ma
-    df <- df[order(df$mid_ma), ]
+    # Add duration
+    df$duration_myr <- (df$max_ma - df$min_ma)
+    # Add bin number
+    df$bin <- seq_len(nrow(df))
+    # Reorder dataframe
+    df <- df[, c("bin", "interval_name", "max_ma", "mid_ma",
+                 "min_ma", "duration_myr", "abbr", "colour")]
+    # Add colours based on https://stackoverflow.com/a/1855903/4660582
```
willgearty, line 259: I think we could do this for any df that has a "colour" column, including user-provided dfs?

willgearty (commented):
Thanks for the changes @LewisAJones, `time_bins()` looks great! I have a few comments on the code, but overall it looks clean and straightforward. As far as I can tell, `time_bins()` now has all of the functionality of `get_timescale_data()`, so `get_timescale_data()` can probably be removed. @KEichenseer, if you have a chance, it would probably be useful to have you review this whole PR (since it has a bunch of `axis_geo` changes, too).

LewisAJones:
> Thanks for the changes @LewisAJones, `time_bins()` looks great! I have a few comments on the code, but overall it looks clean and straightforward. As far as I can tell, `time_bins()` now has all of the functionality of `get_timescale_data()`, so `get_timescale_data()` can probably be removed. @KEichenseer, if you have a chance, it would probably be useful to have you review this whole PR (since it has a bunch of `axis_geo` changes, too).

Hey @willgearty, thanks for the quick review! I've now addressed your comments in my latest commit and made some changes to standardise output. I think this will help with any changes to `axis_geo()`. We still need to remove `get_timescale_data()`, but I believe `axis_geo` needs updating with `time_bins()`?

KEichenseer:
> Thanks for the changes @LewisAJones, `time_bins()` looks great! I have a few comments on the code, but overall it looks clean and straightforward. As far as I can tell, `time_bins()` now has all of the functionality of `get_timescale_data()`, so `get_timescale_data()` can probably be removed. @KEichenseer, if you have a chance, it would probably be useful to have you review this whole PR (since it has a bunch of `axis_geo` changes, too).

Excellent, I will review that by the end of the week.

KEichenseer (commented):
Hi Will and Lewis,

it all seems to work flawlessly, great job!

Here are my comments:
  axis_geo():
  line 178: why does it say "scale with old GTS intervals"? In that case, time_bins(rank = "period") needs to be changed to time_bins(rank = "period",  scale = "GTS2012"), as the default is "GTS2020".
  line 221: `intervals` should default to "epoch" now, to be consistent with the rank names used throughout the package.
    Similarly, in the examples, "period" should be used rather than "periods".

This hasn't changed compared to the previous version, but I realised that for the lty argument I couldn't really predict which lines would change with changing the lty arguments - maybe that can use some extra explanation. 

  time_bins():
  uses "Phanerozoic" as default for `interval`, which contrasts with its use in `axis_geo()`, where `interval` is specified as c("Archaean","Phanerozoic"). 
This results in different abbreviations: e.g., "Cretaceous" is abbreviated to "Cre" when using Precambrian and Phanerozoic periods, but is abbreviated to "Cr" when using only Phanerozoic periods. This is seen in the examples of `axis_geo()`, when intervals is specified with `time_bins` VS when intervals is specified as "periods"

A minor point regarding GTS2012 and GTS2020, the font colours do not always match the font colours in the actual tables on stratigraphy.org. Is that on purpose?

Apart from that, it all looks good, thanks!

Kilian

LewisAJones:
Hi @KEichenseer @willgearty, thanks for the feedback and comments. 

Regarding updating the default behaviour for `time_bins` to use `c("Archaean","Phanerozoic")` instead of `c("Phanerozoic")`, I don't think it quite works here. For `time_bins`, we return stage-level bins by default (this is what most users will want for binning data). For the Precambrian we don't actually have stages defined yet in the GTS. I don't think there is anything wrong with having different defaults for `axis_geo` and `time_bins` because when adding the geological timescale to plots, we typically want period data, but for binning data we want stage data. 

In general, I like the approach of using luminance to define the colour that should be used for the font. However, I think we might want to revert it as it doesn't seem to always work well. For instance, I find white on red (Permian) a little hard to look at. What does everyone else think?

For abbreviations, I believe the approach we've included works well. However, I think I'd prefer that we use the standard abbreviations for the GTS dataframes. I think this mainly relates to periods? This will be a lot more familiar to the community. @willgearty do you want to revert your chances and then maybe you/I can go through and add the appropriate abbreviations? Once that's done I can update `time_bins` to reflect these changes. I'd also like to have one final look at `axis_geo` once we've made these changes. 

Thanks both!

LewisAJones:
Hi @willgearty @KEichenseer, I've gone ahead and added the standard abbreviations in for GTS2012 and GTS2020. Abbreviations are also now only added if NAs are present in the abbreviation column... so, when calling Macrostrat datasets, abbreviations will only be updated if none were present (i.e. skipped for periods but not for stages). 

As I mentioned to Will privately, probably the colour of text labels doesn't matter that much... it was only really the white on red (Permian) where I was a bit dazzled by the contrast. However, it's true that others might struggle to read black on red, so I think this is probably fine too.

Just a side note really, I noticed that the colours in Macrostrat datasets seem to be wrong(?) for the Neo-proterozoic and Meso-proterozoic, as in they don't align with [ICS](https://stratigraphy.org/ICSchart/ChronostratChart2022-02.pdf). This means that plots will look visually different when using GTS2020 and Macrostrats datasets for these time intervals.

R/axis_geo.R
LewisAJones: I guess my point with this comment was whether we would still expect deeptime-formatted timescales as input intervals? Presumably users will not get their data via deeptime to use axis_geo when they could use time_bins in palaeoverse? It's fine though if it allows a wider handling.

LewisAJones (approved):
Just a couple more comments on `axis_geo` @willgearty, otherwise I think this PR is good to go once you and Kilian are satisfied.

willgearty:
Suggested documentation edits made! I've kept the deeptime dataframe support in because deeptime does still have the most recent ICS built-in, so users may want to use that with axis_geo.

> Just a side note really, I noticed that the colours in Macrostrat datasets seem to be wrong(?) for the Neo-proterozoic and Meso-proterozoic, as in they don't align with [ICS](https://stratigraphy.org/ICSchart/ChronostratChart2022-02.pdf). This means that plots will look visually different when using GTS2020 and Macrostrats datasets for these time intervals.

We should reach out to the Macrostrat folks if this is indeed the case. They are usually very responsive to updating the API if something is wrong.

LewisAJones:
Thanks Will, a couple of comments, but I think we're good to go!

> Suggested documentation edits made! I've kept the deeptime dataframe support in because deeptime does still have the most recent ICS built-in, so users may want to use that with axis_geo.
> 

We can leave this in for now, but I think this should probably be removed in the next version. Ultimately, this data can now be accessed via `time_bins`. The only time this will be an issue is if Macrostrat is down or the user is offline? Which if `deeptime` is already installed, it isn't really an issue? I think it becomes quite hard to justify why we provide support for `deeptime` but not for other packages which also have in-built dataframes such as `divDyn`. 

> > Just a side note really, I noticed that the colours in Macrostrat datasets seem to be wrong(?) for the Neo-proterozoic and Meso-proterozoic, as in they don't align with [ICS](https://stratigraphy.org/ICSchart/ChronostratChart2022-02.pdf). This means that plots will look visually different when using GTS2020 and Macrostrats datasets for these time intervals.
> 
> We should reach out to the Macrostrat folks if this is indeed the case. They are usually very responsive to updating the API if something is wrong.

We can mention it in our meeting with Shanan :thumbsup:

PR #79, opened 2023-02-21
Fix unnecessary column output (palaeorotate)

### LINKED ISSUE #78: Unnecessary columns generated in palaeorotate
rot_age, rot_lng and rot_lat columns are generated by default in the palaeorotate function. However, these are not used by the "point" method and should therefore only be generated if the "grid" method is specified. Small bug, but annoying.

LewisAJones:
This PR fixes the unnecessary column output (rot_age, rot_lng and rot_lat) in the palaeorotate function when using the "point" method. It is a small 'bug', but annoying none the less. 

Closes #78.

### DIFF
--- a/R/palaeorotate.R
+++ b/R/palaeorotate.R
@@ -283,9 +283,11 @@ palaeorotate <- function(occdf, lng = "lng", lat = "lat", age = "age",
   }
 
   # Create info columns for "grid" method
-  occdf$rot_age <- NA
-  occdf$rot_lng <- NA
-  occdf$rot_lat <- NA
+  if (method == "grid") {
+    occdf$rot_age <- NA
+    occdf$rot_lng <- NA
+    occdf$rot_lat <- NA
+  }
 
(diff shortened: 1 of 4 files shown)


R/palaeorotate.R
```diff
@@ -437,6 +439,8 @@ palaeorotate <- function(occdf, lng = "lng", lat = "lat", age = "age",
       chk <- seq(from = 0, to = nr + chunks, by = chunks)
```
willgearty, line 439: I think everything from line 426 down should be indented once? Then everything from line 445 should be indented an extra time?

willgearty (changes requested):
A couple comments for now, but I know you are making some other changes! Let's make sure to add a test that covers the special edge case that you found (and fixed). Also, I noticed that we are still using the `::` notation within the code. I think that's technically fine, but it might be better to remove it since we are already importing all of those functions?

willgearty (approved):

PR #80, opened 2023-02-23
Fix binding issue

LewisAJones:
On playing with some more datasets, I noticed there were some binding issues with palaeocoordinates. This PR should fix this.


willgearty (approved):
Looks good to me, thanks for catching another weird edge case!

LewisAJones:
Thanks @willgearty!

PR #81, opened 2023-02-27
update description and example of axis_geo

KEichenseer:
# Function name

## Description

Slightly expand the documentation and one example of `axis_geo()` to make it easier to find that multiple scales can be added to the same side, which can all be modified by using lists for other function arguments.

## Motivation and Context

A user has emailed me, asking how he can modify the labels to have only one set of labels abbreviated. Although this is stated in the _Details_, I think it will be easier to find if this case is included in one of the examples. I've also added the possibility of list inputs to the _interval_ argument, for similar reasons.


willgearty (approved):
Looks good to me, thanks @KEichenseer!

KEichenseer:
Thanks Will!

PR #83, opened 2023-03-02
Fix multi-model call bug in palaeorotate (#82)

### LINKED ISSUE #82: Multi-model call bug in palaeorotate
Multi-model call in palaeorotate for the "point" method does not return all requested model coordinates, only the last called model.

LewisAJones:
Multi-model call in palaeorotate for the "point" method does not return all requested model coordinates, only the last called model. I've implemented a fix for this and added additional tests to catch this in future.

Closes #82.


willgearty (approved):
I checked before and after, and this does fix the indicated bug. The code looks good, and I appreciate the additional test!

PR #85, opened 2023-03-11
fix interval key data

### LINKED ISSUE #84: Atdabanian/Botomian in interval_key incorrect
**Describe the bug**
For the interval name Atdabanian/Botomian in `interval_key` the early stage and late stage are incorrect. Currently, the early stage is "Stage 3" and the late stage "Stage 2". This should be the opposite.

**To Reproduce**
```
interval_key[which(interval_key$interval_name == "Atdabanian/Botomian"), ]
```

LewisAJones:
This PR is a bug fix for #84.

> Describe the bug
For the interval name Atdabanian/Botomian in interval_key the early stage and late stage are incorrect. Currently, the early stage is "Stage 3" and the late stage "Stage 2". This should be the opposite.

Not a lot to say here other than this has now been corrected.

Closes #84.


SofiaGalv (commented):
Ok!

SofiaGalv (approved):
Everything ok!

PR #89, opened 2023-03-31
Vignettes

willgearty:
This adds two vignettes, one on latitudinal trends in reefs and one on alpha diversity and geographic range size in tetrapods.


ChristopherDavidDean (approved):
Both tutorials look good - cover a good range of the current Palaeoverse functions in a clear and accessible format. I'm happy for this to be merged.

willgearty:
I'm investigating why the tetrapods vignette is failing to compile...I think it has to do with the now-fixed palaeorotate problems on CRAN

willgearty:
OK, I think that should fix it

vignettes/tetrapod-biodiversity.Rmd
```diff
@@ -0,0 +1,384 @@
...
+# take a look at the new dataframe
+head(coll_taxa)
+```
+
+To plot an alpha diversity (local richness) plot, we will need the ages of intervals in millions of years. Since the numeric ages may have changed since we downloaded the data from the PBDB, we'll use `look_up` to get the numeric ages for the occurrences within `tetrapods` dataset based on their assigned interval names:
+
+```{r}
+# get new numeric ages for named intervals
+# the default is to use the Geological Time Scale 2020
+occdf <- look_up(maj_tetrapods)
```
LewisAJones, line 271: I think this should probably be introduced before binning data earlier on? Note that not all intervals can be matched using just GTS2020. You might want to also include the built-in interval key?

vignettes/tetrapod-biodiversity.Rmd
```diff
@@ -0,0 +1,384 @@
...
+                               name = "genus",
+                               lng = "p_lng",
+                               lat = "p_lat",
+                               method = "con")
+
+# Have a look at the dataset
+head(space_tetrapods)
+
+```
+As you can see, the combination of `group_apply` and `tax_range_space` produces a dataset with the area of geographic range for each taxonomic unit in kilometers squared for each time interval. From this information we can now plot how the average range size of tetrapods changed throughout the Carboniferous and Permian:
```
LewisAJones, line 345: Maybe worth mentioning that an area of 0 here indicates that there is only one occurrence of that taxon?

vignettes/tetrapod-biodiversity.Rmd
```diff
@@ -0,0 +1,384 @@
...
+# Find the average geographic range size for each time interval
+space_tetrapods_mean <- space_tetrapods %>%
+  group_by(bin_midpoint) %>%
+  summarise(mean_area = mean(area))
+
+# Create a plot of average range size through time
+plot(space_tetrapods_mean$bin_midpoint, space_tetrapods_mean$mean_area, # add the points
+     axes = FALSE,
+     xlim = rev(range(as.numeric(space_tetrapods_mean$bin_midpoint), na.rm = TRUE)), # reverse the x-axis
+     xlab = " ", ylab = "Average range size of tetrapod genera (km2)", # axes labels
```
LewisAJones, line 358: We should probably use superscript here for the 2 to give the proper units

LewisAJones (commented):
Hey everyone!

Thanks for putting together the tetrapod tutorial @ChristopherDavidDean (and @willgearty?). I think it is coming together very nicely. I have a few general comments but I think it is almost there. It's probably worth having a couple more pairs of eyes taking a look over these though. I would also suggest knitting the document and just making sure you're happy with the way the vignette appears in different formats (pdf/html). 

One concern I have is the build time of the vignettes. I think this is causing the R CMD Checks to take a while to run. This is most likely due to accessing the GPlates API for `palaeorotate`. However, functions like `bin_time` with the 'majority' method are probably also slowing it down. I am wondering if there is way to skip building locally each time and during CRAN checks.  Any idea @willgearty? 

Also thanks for the minor bug fix @willgearty!

Lewis

willgearty:
Putting these here for now, because we don't want to do it until we've finished editing the Rmd files: https://www.kloppenborg.ca/2021/06/long-running-vignettes/
https://ropensci.org/blog/2019/12/08/precompute-vignettes/

willgearty:
OK, I've addressed almost all of your comments @LewisAJones. Maybe @ChristopherDavidDean has some thoughts about the location of `look_up()`.

Also, for the record, it only took me ~50 seconds to knit the tetrapods vignette just now. I wonder if the gplates API was just being realllllly slow the other day?

LewisAJones:
> OK, I've addressed almost all of your comments @LewisAJones. Maybe @ChristopherDavidDean has some thoughts about the location of `look_up()`.
> 

Thanks @willgearty! The changes look good to me... @ChristopherDavidDean let us know if you have any other thoughts/comments e.g. regarding `look_up()`.

> Also, for the record, it only took me ~50 seconds to knit the tetrapods vignette just now. I wonder if the gplates API was just being realllllly slow the other day?
> 
Yeah, I think this can be fairly quick locally, but I think it is slower when run via GitHub actions? This is obviously a pain and not ideal. However, it looks like you've found some solutions to this. I was thinking one option could also be not to include the vignettes directly in the package but as a resource on our website (if we do not rebuild the vignettes each time then this basically seems the equivalent to me).

> Edit: Ah, it's the reefs one that takes a long time. It just took me about 10 minutes to knit.

Yes, this is for two reasons, (1) we use multiple plate models in the reef vignette and (2) some models return coordinates much quicker. The GOLONKA and PALEOMAP model are rather quick because they are not full-plate models (i.e. plate boundaries do not evolve over time) whereas the MERDITH2021 model is.

willgearty:
> However, it looks like you've found some solutions to this. I was thinking one option could also be not to include the vignettes directly in the package but as a resource on our website (if we do not rebuild the vignettes each time then this basically seems the equivalent to me).
I think I would prefer that we just knit them and include them in the package as described in the pages I linked above. That way we are consistent with the way other pkgdown sites work and they are also bundled with the package so users can read them locally.

ChristopherDavidDean:
Hey both, 

Made some final changes to the vignette, including moving 'look_up' higher in the document and making changes to accommodate for this, as well as some final checks and minor formatting. Let me know what you think.

LewisAJones:
> Hey both,
> 
> Made some final changes to the vignette, including moving 'look_up' higher in the document and making changes to accommodate for this, as well as some final checks and minor formatting. Let me know what you think.

Thanks @ChristopherDavidDean! This is looking great now and I think we're almost ready to merge. @willgearty I've just setup the framework for pre-building vignettes following the guide you shared. Do you want have a final look to see if all is in order before we go ahead and merge?

PR #91, opened 2023-07-13
Fix custom ticks behavior for phylogenies

willgearty:
This fixes an issue reported on the google group that occurred when trying to add an axis with just tick marks to a phylogeny. This should make this now possible. In addition to that, I've also made it so `exact` works with phylogenies as well. Finally, I've made it so that `exact` now overrides `tick_at`/`tick_labels`. I've updated the documentation to make all of this behavior a bit clearer, and I've modified some tests to now cover these possibilities.


bethany-j-allen:
I'm trying to run the examples but keep getting an `undefined columns selected` error. I think the problem might be with `time_bins` rather than `axis_geo`, any idea what I'm doing wrong?

LewisAJones:
> I'm trying to run the examples but keep getting an `undefined columns selected` error. I think the problem might be with `time_bins` rather than `axis_geo`, any idea what I'm doing wrong?

Hmmm odd. I've just tried from my end and the examples run through no problem, so not sure what is going on? Just to check, are you using the direct examples from the `axis_geo` documentation? If not, can you provide a reprex? Is the error happening on both standard and phylogeny plots? I guess the error you're seeing could be related to either function as `axis_geo` uses the output from `time_bins`, but I can't reproduce it.

bethany-j-allen (approved):
No idea why but it's working fine now :) guess I failed to run some part of the code somehow! All looks great to me, and definitely a nice option to have.

willgearty:
Weird, thanks for the quick review @bethany-j-allen!

PR #93, opened 2023-07-17
Fix handling of multiple models in palaeorotate

### LINKED ISSUE #92: Multimodel binding issue in palaeorotate
When calling multiple GPMs at once, there seems to be a binding issue with the palaeocoordinates. Note, this is not an issue if palaeorotations are generated iteratively and must be related to chunk handling. This should be resolved quickly.

LewisAJones:
This PR gives a much needed update to `palaeorotate`. It fixes the issue with multimodel binding of the palaeocoordinates and significantly improves how API requests are handled. This has required a major rewrite of the code. However, it now allows for models to be easily added/removed when changes are made to the GPlates Web Service (expected soon).

Fixes #92.

### DIFF
--- a/R/palaeorotate.R
+++ b/R/palaeorotate.R
@@ -419,7 +419,8 @@ palaeorotate <- function(occdf, lng = "lng", lat = "lat", age = "age",
       },
       error = function(e) {
         stop(paste("GPlates Web Service is not available.",
-                   "Either the website is down or you are not connected to the internet."),
+                   "Either the website is down or you are not",
+                   "connected to the internet."),
              call. = FALSE)
       })
 
@@ -488,14 +489,12 @@ palaeorotate <- function(occdf, lng = "lng", lat = "lat", age = "age",
           req <- query[[m]]
           req$time <- i
           req$model <- m
-          pts <- paste(x[, lng], x[, lat], sep = ",")
-          pts <- toString(pts)
-          pts <- gsub(" ", "", pts)
+          pts <- paste(x[, lng], x[, lat], sep = ",", collapse = ",")
           req$points <- pts
           return(req)
         })
         # Call API
-        rots <- lapply(request, function(x) {
+        api_req <- lapply(request, function(x) {
           RETRY(verb = "GET",
                 url = pbase,
                 query = x,
@@ -505,10 +504,10 @@ palaeorotate <- function(occdf, lng = "lng", lat = "lat", age = "age",
                 pause_cap = 10)
         })
         # Extract coordinates
-        rots <- lapply(rots, function(x) {
+        rots <- lapply(api_req, function(x) {
           coords <- content(x = x, as = "parsed")$coordinates
           # Replace NULL values
-          rpl <- unlist(lapply(coords, is.null))
+          rpl <- sapply(coords, is.null)
           if (sum(rpl) != 0) {
             rpl <- which(rpl == TRUE)
             for (r in rpl) coords[[r]] <- list(NA, NA)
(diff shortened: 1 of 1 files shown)


R/palaeorotate.R
```diff
@@ -419,108 +419,130 @@ palaeorotate <- function(occdf, lng = "lng", lat = "lat", age = "age",
...
+                times = 5,
+                pause_min = 1,
+                pause_base = 1,
+                pause_cap = 10)
+        })
+        # Extract coordinates
+        rots <- lapply(rots, function(x) {
+          coords <- content(x = x, as = "parsed")$coordinates
+          # Replace NULL values
+          rpl <- unlist(lapply(coords, is.null))
```
willgearty, line 511: sapply?

R/palaeorotate.R
```diff
@@ -419,108 +419,130 @@ palaeorotate <- function(occdf, lng = "lng", lat = "lat", age = "age",
       },
       error = function(e) {
         stop(paste("GPlates Web Service is not available.",
-      "Either the website is down or you are not connected to the internet."),
+                   "Either the website is down or you are not connected to the internet."),
```
willgearty, line 422: line is too long

R/palaeorotate.R
```diff
@@ -419,108 +419,130 @@ palaeorotate <- function(occdf, lng = "lng", lat = "lat", age = "age",
...
              call. = FALSE)
       })
+
+    # Set-up for point method -------
+    # Define root
+    pbase <- "https://gws.gplates.org/reconstruct/reconstruct_points/"
+    # Setup query structure
+    query <- list(points = NULL, time = NULL,
+                  model = NULL, anchor_plate_id = 0,
```
willgearty, line 431: Does this anchor plate id matter? Should we document that we are using `0`?

R/palaeorotate.R
```diff
@@ -419,108 +419,130 @@ palaeorotate <- function(occdf, lng = "lng", lat = "lat", age = "age",
...
+                url = pbase,
+                query = x,
+                times = 5,
+                pause_min = 1,
+                pause_base = 1,
+                pause_cap = 10)
+        })
+        # Extract coordinates
+        rots <- lapply(rots, function(x) {
+          coords <- content(x = x, as = "parsed")$coordinates
```
willgearty, line 509: I'd be careful using the same variable name as a global variable (seems to work fine here, but it definitely makes reading the code confusing)

willgearty (changes requested):
Just a few comments. Otherwise, code seems to look good and work as desired!

R/palaeorotate.R
```diff
@@ -419,108 +419,130 @@ palaeorotate <- function(occdf, lng = "lng", lat = "lat", age = "age",
...
-              rots[[r]] <- list(NA, NA)
-            }
+        # Build requests
+        request <- lapply(tmp, function(x) {
+          req <- query[[m]]
+          req$time <- i
+          req$model <- m
+          pts <- paste(x[, lng], x[, lat], sep = ",")
+          pts <- toString(pts)
+          pts <- gsub(" ", "", pts)
```
willgearty, line 493: I think this can be replaced with just the following:
`paste(x[, lng], x[, lat], sep = ",", collapse = ",")`

R/palaeorotate.R
LewisAJones: This basically indicates using their default reference frame (palaeomagnetic/mantle). For the models currently included it doesn't really matter. However, some of the new models being added are in a mantle reference frame (e.g. TorsvikCocks2017) by default and will need specifying as 1 to shift them to their palaeomagnetic reference frame. I don't think we need to update this now as it is the default behaviour. However, when we implement the major update coming, we will specify that all models are in their palaeomagnetic reference frame.

LewisAJones:
Thanks for the quick review @willgearty! I've implemented all your suggestions so I think we're good to go?

willgearty (approved):
Looks good to me, thanks!

PR #95, opened 2023-08-14
Added ability to use custom bins for tax_expand_time

willgearty:
This adds the ability to use a custom set of temporal bins for the `tax_expand_time` function. It now behaves just like `tax_expand_lat` and `bin_time`. The old arguments are maintained for backwards-compatibility (and are just passed to `time_bins` for convenience).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,7 @@
 
 * Fixed handling of multiple models in palaeorotate (#92)
 * Fixed custom ticks behavior for axis_geo_phylo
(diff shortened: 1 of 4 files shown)


LewisAJones (approved):
Hi @willgearty, thanks for this update, updated documentation and code looks great! My only comment is whether we want the output dataframe sorted in a certain way. Currently, it is essentially sorted by the time bins/age data. Would it be better to be sorted by taxa (genuine question)?

Issue #97, opened 2023-11-12
Range plot for stratigraphic sections


bethany-j-allen:
**Is your feature request related to a problem? Please describe.**
This problem was raised by Meghan Jenkinson (via X): an R-based way to plot stratigraphic range and occurrence data for an individual section.

**Describe the solution you'd like**
An extension of tax_range_time which plots ranges across beds within a section, including points indicating specific sampled levels, and ideally with open points for uncertain identifications.

**Describe alternatives you've considered**
These figures are common but typically made by hand for manuscripts, but it should be possible to generate them automatically from the input data.

**Additional context**
I already have some code from Alex Dunhill; I will be working on a full draft over the next couple of weeks.

PR #98, opened 2023-11-15
Stratigraphic section plot

bethany-j-allen:
This is a draft of a function addressing Issue #97.

The function appears to work, but still needs some final polishing. I wanted some feedback on the total functionality we want before doing this though.

Some thoughts:
- At the moment, stratigraphic levels can only be specified as numbers, either a height (in metres) or numbered beds. Do we also want people to be able to use stratigraphic names, and specify the order in which those names should appear?
- The y-axis currently appears as continuous numbers across the range of values given, but we could also add an option where only the sampled beds are labelled on the axis. We could also consider shading alternate beds, I'm not sure whether that would make the plots more or less readable.
- "Certainty" is currently expressed as a separate binary variable. I think this is a nice option and should be available, but I wonder whether we should also be able to read a list of taxonomic names and automatically identify the "?", "cf.", "aff." occurrences.
- When certainty is used, ranges are currently drawn across the full range, regardless of whether the end occurrences are certain or not. Do we want to e.g. used dashed lines for where an end of a range is being extended by an uncertain occurrence? I think this would be nice but will probably be fiddly to code.
- Plot sizing for the x-axis seems like painful trial-and-error, do we have a better solution?


R/plot_section.R
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  #Determine y-axis label
+  if (units == "height") {
+    y_axis <- "Stratigraphic height (m)"
+  } else
+    y_axis <- "Bed number"
+
+  #Create plot
+  plot(x = NA, y = NA,
```
willgearty, line 135: I find it really hard to balance/design functions that make a plot for you. How much of the plotting functionality should you expose to the user? In this case, I would probably make "xlab" and "ylab" additional arguments of the new function (with ylab replacing the "units" argument). I would probably also add a "..." argument which you would then pass directly to `plot()`.

R/plot_section.R
```diff
@@ -0,0 +1,154 @@
...
+       main = label)
+  segments(y0 = ranges$min_bin, y1 = ranges$max_bin,
+           x0 = ranges$ID,
+           col = "black")
+  points(y = certain$bed, x = certain$ID, pch = 19,
+         col = "black")
+  if (certainty != FALSE){
+    points(y = uncertain$bed, x = uncertain$ID, pch = 1,
+          col = "black")
+  }
```
willgearty, line 150: This might be personal preference, but I really don't like seeing the lines peak out from behind the circles. I'd recommend having filled circles (with white fills). But again, how much of this should be left to the user to define?

willgearty (changes requested):
This is a great draft, thanks @bethany-j-allen! I've added some comments regarding the code itself. I'll also address the great discussion points you brought up:

1. In an ideal world, we would probably find some very nifty way to make this function work with `axis_geo()` (as in you would be able to add the geological/stratigraphic column/axis to the y-axis), which means we would also support stratigraphic names like you've suggested.
2. I think all beds should be listed, not just those with occurrences, otherwise I think it could be quite misleading. If we're able to make this function work with axis_geo, I think users could just add their own alternating bed shading?
3. Hmm...I originally read "certainty" as whether the identification of the bed was certain (and had not even thought about taxonomic certainty). Given this, it seems like there are lots of types of certainty, so I would suggest just keeping it vague as it is now (or somehow allowing for more than one type of certainty).
4. As I started reading this, I immediately thought of dashed lines, before I even read your suggestion, so yes, I'd definitely go with dashed lines. I think the easiest way would be to calculate the "uncertain" range of each taxon, then plot those as dashed lines, then calculate the "certain" range of each taxon, then plot those as solid lines. The solid lines should cover up the dashed lines where we are certain and you would only see solid lines where the entire range is certain, I think?
5. Ugh, base R. If you're talking about the margin below the plot, I'd say we could just leave this to the user? It should be fairly easy for someone to change the margins with `par()` beforehand and then call our function. We'd just want to include this in the example in the docs.

Happy to help with anything or provide more feedback once you get around to making changes!

R/plot_section.R
```diff
@@ -0,0 +1,153 @@
+#' Generate stratigraphic section plot
+#'
+#' A function to plot the stratigraphic ranges of fossil taxa from occurrence
+#' data.
+#'
+#' @param occdf \code{dataframe}. A dataframe of fossil occurrences containing
+#' at least two columns: names of taxa, and their stratigraphic position
+#' (see `name` and `level` arguments).
+#' @param name \code{character}. The name of the column you wish to be treated
+#' as the input names, e.g. "genus" (default).
```
LewisAJones, line 10:
```suggestion
#' as the input names, e.g. "genus" (default).
```
"taxon" seems to be the current default?

R/plot_section.R
```diff
@@ -0,0 +1,153 @@
...
+#' @param certainty \code{character}. The name of the column you wish to be
+#' treated as the information on whether an identification is certain ("1") or
+#' uncertain ("0"). By default, no column name is provided, and all occurrences
+#' are assumed to be certain. In the plot, any occurrence labelled as certain
+#' will be plotted with a black circle, while any occurrence labelled as
+#' uncertain will be plotted with a white circle.
+#' @param by \code{character}. How should the output be sorted?
+#' Either: "FAD" (first appearance; default), "LAD" (last appearance), or
+#' "name" (alphabetically by taxon names).
+#' @param units \code{character}. Should the y-axis be labelled as "height"
```
LewisAJones, line 23: I am not sure we really need this parameter (or label), we could just use the default plot parameters available (main, ylab, etc) which could be passed via `...`.

R/plot_section.R
```diff
@@ -0,0 +1,153 @@
...
+#' with levels at which the taxon was sampled indicated with a point.
+#'
+#' @section Developer(s):
+#' Bethany Allen & Alexander Dunhill
+#' @section Reviewer(s):
+#' William Gearty & Lewis A. Jones
+#'
+#' @examples
+#' # Sample tetrapod occurrences
+#' tetrapod_names <- tetrapods$accepted_name[1:50]
```
LewisAJones, line 40: It's not really required, but in general I guess we should demonstrate that tetrapods is coming from `palaeoverse`.

R/plot_section.R
```diff
@@ -0,0 +1,153 @@
...
+#' # Simulate bed numbers
+#' beds_sampled <- sample.int(n = 10, size = 50, replace = TRUE)
+#' # Simulate certainty values
+#' certainty_sampled <- sample(x = 0:1, size = 50, replace = TRUE)
+#' #Combine into data frame
+#' occdf <- data.frame(taxon = tetrapod_names, bed = beds_sampled,
+#' certainty = certainty_sampled)
+#' # Plot stratigraphic ranges
+#' par(mar = c(12, 5, 2, 2))
+#' tax_range_strat(occdf)
```
LewisAJones, line 50: I think this needs updating? The function is currently called `plot_section`, though I think I prefer `tax_range_strat`.

R/plot_section.R
```diff
@@ -0,0 +1,153 @@
...
+  if (is.data.frame(occdf) == FALSE) {
+    stop("`occdf` should be a dataframe")
+  }
+
+  if (!is.numeric(occdf[, level])) {
+    stop("`level` must be of class numeric.")
+  }
+
+  if (any(c(name, level) %in% colnames(occdf) == FALSE)) {
+    stop("Either `name` or `level` is not a named column in\n`occdf`")
```
LewisAJones, line 69:
```suggestion
    stop("Either `name` or `level` is not a named column in `occdf`")
```

R/plot_section.R
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  if (certainty != FALSE && certainty %in% colnames(occdf) == FALSE) {
+    stop("`certainty` is not a named column in\n`occdf`")
```
LewisAJones, line 73:
```suggestion
    stop("`certainty` is not a named column in `occdf`")
```

LewisAJones (changes requested):
Hi @bethany-j-allen, thanks for getting this together. It looks like you've already made a lot of progress on it! I have a few general comments before going through the whole code:
- There is a lot of overlapping functionality here with `tax_range_time`, I wonder if it is completely unreasonably to somehow merge these? 
- As this function is currently focused on just plotting data, I think it would be great to make more customisable options available. For example, being able to italicise species names on the x-axis.
- I agree with @willgearty that it would be nice to be able to integrate `axis_geo` in some way.

Relating to your points:
- I think it would be great to be able to plot stratigraphic names. Maybe most of the work for this can be done by providing a custom dataframe to `axis_geo`. This can then add the names but also draw the respective border polygons.  
- I guess a lot of customisable functionality is already available in `axis_geo`, I wonder if it would almost be worth implementing some changes there related to shading? I guess we could do something like controlling the alpha of colours?
- I personally think it is fine to keep it as a binary state, users should be able to relatively easily generate such a column (we could also provide an example in the example section).
- I like the idea of a dashed line! Should this be reserved for just uncertain taxon assignments or also for bins where the taxon is inferred to be present, but no occurrence is observed?
- I guess sizing will be different for everyone, but users can set up their plot space themselves as Will suggested.

bethany-j-allen:
Thanks for your comments @willgearty and @LewisAJones!

So far I have:
- Updated the `return` information in the documentation
- Changed the uncertain points to a white fill, so you can't see the line underneath anymore
- Changed the uncertain ends of a range to a dashed line which is then overplotted by the certain ranges, as suggested by Will
- Changed the axis label options to `xlab` and `ylab` which are passed directly to the plot, and removed the title in favour of using `...`

To respond to the remaining points:
- On using 'stratigraphic names', I meant something different - instead of using bed numbers, at a larger scale people may wish to label their stratigraphic levels as a series of e.g. members or formations. This would require the user to provide an order for these variables, and a little extra framework for handling both numeric and character data for the y-axis. I could see this potentially being useful, and perhaps would not be too much work (i.e. worth doing now rather than as an enhancement later). Would you agree?
- On using `axis_geo()`, I do see that it would be great to be able to add the geological age information as an extra facet on the y-axis. But as I see it, this would be painful given the infrastructure we currently have. `axis_geo()` takes in age information, and often direct ages are not available for individual beds in a stratigraphic column. Even if we asked people to provide a vector of 'artificial' ages which represented the correct geological interval for each bed, we would still need to figure out how to get this information to plot on the same axis as the bed/height information. I think this would be difficult, but perhaps if one of you can see a way for it to be less difficult (e.g. allow people to provide the geological interval for each bed rather than an age) then perhaps we could collaborate on adding that (at a later date?).
- I agree with the idea of all _beds_ being listed on the y-axis, but if the y-axis represents _height_, perhaps it would be good for people to be able to label the axis with only the sampled heights - what do you think?
- It's in the code already that sorting by FAD/LAD already involves sorting by the inverse first, then the desired order, because it looks a lot neater. I don't think we need to add other options, personally.
- I think this function should stand alone from `tax_range_time()` because, as Will points out, it does not carry out the computation that the latter function does, only the plotting. Putting it into `tax_range_time()` could also 'hide' it from potential users. On a related point, I changed the name to `plot_section()` as Will requested, but I am open to discussing the name some more.
- The option to italicise the species names is a great idea but is currently a faff (the `...` is passed to `plot` but the axis options are added via a later `axis` call, maybe one of you knows these functions better than me and can come up with an alternative?). We could also choose to make italic names on the x-axis labels automatic/default, what do you think?
- On using dashed lines "for bins where the taxon is inferred to be present, but no occurrence is observed" - what we currently have is the convention for stratigraphic range plots, and I think most readers of these plots would understand that a bed which is ranged through, but does not have a point, means that the taxon was not sampled in that bed.

As a final note: my intention is to add some `expect_doppelganger` tests to bolster the test coverage once we are happy with the function.

willgearty:
This is all great, thanks @bethany-j-allen! I've got some ideas but probably won't be able to provide a real response here until sometime next week.

willgearty:
Apologies for how long it's taken me to get back to this. Even then, I still haven't figured out solutions for everything. I took a little time to address a few things:

1. Fixed up some broken code (certainty is now NULL by default, fixed seq_along usage, uncertain point shape type, etc)
2. Split out `...` to three separate arguments for plot and axis arguments
3. Made the x-axis tick labels italicized by default
4. Because I was lazy in my debugging, I made it so the function invisibly returns the `ranges` data.frame (but that seems useful to me, so I kept that functionality). This is not intended to replace `tax_range_time()`, but to supply the user with the data they would need to make their own plot, if they so wished.

Now for some thoughts about the other points you brought up @bethany-j-allen and @LewisAJones:

1. I think `axis_geo()` is the key to adding stratigraphic (e.g., formation) information. All we need is for `axis_geo()` to be able to handle custom discrete data like [`coord_geo()`](https://williamgearty.com/deeptime/articles/coord_geo.html#custom-discrete-scales). The user would then be able to add a custom geological/stratigraphic axis to the right side of the plot (or to even cover the left axis and then make their own numeric axis further to the left). If my thinking is correct, this would mean that no further work would need to be done in this PR with respect to the integration of `axis_geo()`. In fact, it looks like this already works (at least for some cases):

``` r
library(palaeoverse)
tetrapod_names <- tetrapods$accepted_name[1:50]
beds_sampled <- sample.int(n = 10, size = 50, replace = TRUE)
certainty_sampled <- sample(x = 0:1, size = 50, replace = TRUE)
occdf <- data.frame(taxon = tetrapod_names, bed = beds_sampled,
                    certainty = certainty_sampled)
par(mar = c(12, 5, 2, 6))
plot_section(occdf, certainty = "certainty", ylab = "Stratigraphic height (m)")
eras_custom <- data.frame(
  name = c("Mesozoic", "Cenozoic"), max_age = c(0.5, 3.5),
  min_age = c(3.5, 9.5), color = c("#67C5CA", "#F2F91D")
)
eras_custom2 <- data.frame(
  name = c("Bethany's\nFormation", "Will's\nFormation"), max_age = c(0.5, 6),
  min_age = c(6, 10), color = c("purple", "pink")
)
axis_geo(side = 4, intervals = list(eras_custom, eras_custom2))
```



<sup>Created on 2024-01-05 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

2. With regards to the y-axis labels, I think labels should always be standardized in some way. Only labelling some heights seems confusing to me (unless you can supply an existing example figure where this is done) if the axis is linear/log in nature. I suppose if the heights are treated as discrete data it could work, but I'm not sure I like that either. I think if users really want to label things, they could add their own height labels to the post afterwards?

bethany-j-allen:
Thanks a lot for the fixes @willgearty! I also agree that amending `axis_geo()` is probably the best thing to do - and I guess there may be other plots for which being able to add a similar extra axis might be useful.

For the y-axis labels, I did find an example of a section plot with just the sampled levels labelled, Fig. 2 [here](https://doi.org/10.1111/j.1475-4983.2008.00810.x). But it took me a while to find one, so maybe it's becoming recognised as poor practice. I am happy to leave this if you are.

willgearty (approved):

willgearty:
Ah, now I see with that example figure. I think I'm fine with the options that are currently implemented, and then folks can add those kind of details on their own if they like. Looks like (to me) all that's left is to add some more tests to increase the code coverage.

R/plot_section.R
```diff
@@ -0,0 +1,207 @@
...
+#'              ylab = "Stratigraphic height (m)")
+#' plot_section(occdf, certainty = "certainty", by = "LAD",
+#'              ylab = "Bed number")
+#' plot_section(occdf, certainty = "certainty", by = "name",
+#'              ylab = "Bed number")
+#' plot_section(occdf, certainty = "certainty", ylab = "Bed number",
+#'              plot_args = list(main = "Section A"))
+#'
+#' @export
+plot_section <- function(occdf, name = "taxon", level = "bed",
```
LewisAJones, line 70: Elsewhere in the package, the default for `name` is generally "genus", do we want to stick with this?

R/plot_section.R
```diff
@@ -0,0 +1,207 @@
...
+                         certainty = NULL, by = "FAD",
+                         xlab = "", ylab = "", plot_args = list(),
+                         x_args = list(font = 3, las = 2), y_args = list()) {
+
+  if (is.data.frame(occdf) == FALSE) {
+    stop("`occdf` should be a dataframe")
```
LewisAJones, line 76:
```suggestion
    stop("`occdf` should be a dataframe.")
```

R/plot_section.R
```diff
@@ -0,0 +1,207 @@
...
+  }
+
+  if (!is.numeric(occdf[, level])) {
+    stop("`level` must be of class numeric.")
+  }
+
+  if (any(c(name, level) %in% colnames(occdf) == FALSE)) {
+    stop("Either `name` or `level` is not a named column in `occdf`")
```
LewisAJones, line 84:
```suggestion
    stop("Either `name` or `level` is not a named column in `occdf`.")
```

R/plot_section.R
```diff
@@ -0,0 +1,207 @@
...
+  }
+
+  if (!is.null(certainty)) {
+    if (!is.character(certainty)) {
+      stop("`certainty` must either be of class character or NULL")
```
LewisAJones, line 89:
```suggestion
      stop("`certainty` must either be of class character or NULL.")
```

R/plot_section.R
```diff
@@ -0,0 +1,207 @@
...
+    }
+    if (certainty %in% colnames(occdf) == FALSE) {
+      stop("`certainty` is not a named column in `occdf`")
```
LewisAJones, line 92:
```suggestion
      stop("`certainty` is not a named column in `occdf`.")
```

LewisAJones: Here and elsewhere we should add `.`, or just remove them in all messaging?

LewisAJones (changes requested):
Hi @bethany-j-allen and @willgearty, sorry it has taken me so long to get around to this! It seems there has been a lot of progress since last time I took a look, great job! I think I'm caught up with the comments and I am in general agreement with everything. The function is looking good! 

I agree with using `axis_geo` as a follow up call for users to add GTS/other time bins after calling this function. As you note @bethany-j-allen, the updates we might make to `axis_geo` could be useful elsewhere in the future!

My main comment is that I would really prefer this to be a function used for more than just plotting data. I think by renaming it to `tax_range_strat` and returning the actual stratigraphic ranges aligns with the rest of the palaeoverse a lot better.

Anyway, I've been through the code again and I have a few additional (minor!) comments/suggestions.  Let me know if anything is unclear.

Thanks for your patience! 🙈

R/plot_section.R
```diff
@@ -0,0 +1,207 @@
...
+  do.call(axis, args = c(list(side = 2, at = unique(occdf$bed)), y_args))
+  # plot x-axis
+  dump <- c("side", "at", "labels")
+  if (any(names(x_args) %in% dump)) {
+    x_args <- x_args[-which(names(x_args) %in% dump)]
+  }
+  do.call(axis, args = c(list(side = 1, at = ranges$ID, labels = ranges$taxon),
+                         x_args))
+  box()
+  invisible(ranges)
```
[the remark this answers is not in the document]
willgearty: invisible() does return the object?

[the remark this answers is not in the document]
willgearty: (unless you would rather just use return() instead of invisible())

R/plot_section.R
```diff
@@ -0,0 +1,207 @@
...
+#'
+#'   Function is usually used for its side effect, which is to create a plot
+#'   showing the stratigraphic ranges of taxa in a section, with levels at which
+#'   the taxon was sampled indicated with a point.
+#'
+#' @section Developer(s): Bethany Allen & Alexander Dunhill
+#' @section Reviewer(s): William Gearty & Lewis A. Jones
+#' @importFrom graphics axis par segments plot points box
+#'
+#' @examples
```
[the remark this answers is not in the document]
willgearty: I thought it might be better to showcase this in one of the vignettes, if possible?

R/plot_section.R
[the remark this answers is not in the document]
LewisAJones: Yes, I mean to use `return()` so that it is also printed.

R/plot_section.R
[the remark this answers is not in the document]
LewisAJones: We could do that too, but I suspect a lot of people don't go to the vignettes. If we provide something simple in the examples, I think this would be most useful. Are you concerned about the length of the examples?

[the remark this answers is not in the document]
willgearty: Yeah, I was mostly worried about the length of the examples

R/plot_section.R
```diff
@@ -0,0 +1,207 @@
...
+#'   occurrence labelled as certain will be plotted with a black circle, while
+#'   any occurrence labelled as uncertain will be plotted with a white circle.
+#' @param by \code{character}. How should the output be sorted? Either: "FAD"
+#'   (first appearance; default), "LAD" (last appearance), or "name"
+#'   (alphabetically by taxon names).
+#' @param xlab \code{character}. The x-axis label (over taxa) passed directly to
+#'   \code{plot}.
+#' @param ylab \code{character}. The y-axis label (over strata) passed directly
+#'   to \code{plot}.
+#' @param plot_args A list of optional arguments that are passed directly to
```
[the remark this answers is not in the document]
willgearty: I'm fine with that

R/plot_section.R
```diff
@@ -0,0 +1,207 @@
...
+  labels <- ranges[, c("taxon", "ID")]
+  occdf <- merge(occdf, labels, by.x = name, by.y = "taxon")
+
+  #Obtain uncertain occurrences
+  if (!is.null(certainty)) {
+    uncertain <- occdf[(occdf[, certainty] == 0), ]
+  }
+
+  #Create plot
+  dump <- c("x", "y", "xlim", "ylim", "axes", "xlab", "ylab")
```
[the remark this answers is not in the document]
willgearty: I haven't tested it, but it's possible we can just leave xlim and ylim blank by default and then allow them as plot_args, but it's possible we will need to set them if they are not set by the user.

tax_range_strat_font.svg
LewisAJones, line 1: I think this was unintended, and should be removed?

R/tax_range_strat.R
```diff
@@ -0,0 +1,225 @@
...
+#' @examples
+#' #Load tetrapod dataset
+#' data(tetrapods)
+#' # Sample tetrapod occurrences
+#' tetrapod_names <- tetrapods$accepted_name[1:50]
+#' # Simulate bed numbers
+#' beds_sampled <- sample.int(n = 10, size = 50, replace = TRUE)
+#' # Simulate certainty values
+#' certainty_sampled <- sample(x = 0:1, size = 50, replace = TRUE)
+#' #Combine into data frame
```
LewisAJones, line 56:
```suggestion
#' # Combine into data frame
```

R/tax_range_strat.R
```diff
@@ -0,0 +1,225 @@
...
+#'   The function is usually used for its side effect, which is to create a plot
+#'   showing the stratigraphic ranges of taxa in a section, with levels at which
+#'   the taxon was sampled indicated with a point.
+#'
+#' @section Developer(s): Bethany Allen, William Gearty & Alexander Dunhill
+#' @section Reviewer(s): William Gearty & Lewis A. Jones
+#' @importFrom graphics axis par segments plot points box
+#'
+#' @examples
+#' #Load tetrapod dataset
```
LewisAJones, line 48:
```suggestion
#' # Load tetrapod dataset
```
Just standardising...

R/tax_range_strat.R
```diff
@@ -0,0 +1,225 @@
...
+#' # Plot stratigraphic ranges
+#' par(mar = c(12, 5, 2, 2))
+#' tax_range_strat(occdf, name = "taxon")
+#' tax_range_strat(occdf, name = "taxon", certainty = "certainty",
+#'                 plot_args = list(ylab = "Stratigraphic height (m)"))
+#' # Plot stratigraphic ranges with more labelling
+#' tax_range_strat(occdf, name = "taxon", certainty = "certainty", by = "name",
+#'                 plot_args = list(main = "Section A",
+#'                                  ylab = "Stratigraphic height (m)"))
+#' eras_custom <- data.frame(name = c("Mesozoic", "Cenozoic"),
```
LewisAJones, line 69: Do you think it is worth having a comment here saying that you can add a geological time scale and that the max_age/min_age refer to the height?

R/tax_range_strat.R
```diff
@@ -0,0 +1,225 @@
...
+#'                           max_age = c(0.5, 3.5),
+#'                           min_age = c(3.5, 10.5),
+#'                           color = c("#67C5CA", "#F2F91D"))
+#' axis_geo(side = 4, intervals = eras_custom, tick_labels = FALSE)
+#'
+#' @export
+tax_range_strat <- function(occdf, name = "genus", level = "bed",
+                            certainty = NULL, by = "FAD",
+                            plot_args = list(xlab = "", ylab = "Bed number"),
```
LewisAJones, line 78: Something to be aware of is that you can't really use xlab currently as taxonomic names are plotted on top of it. I don't think this really matters here because it is rare(?) to see an xlab/title on range charts? However something you might want to consider updating in the future.

R/tax_range_strat.R
```diff
@@ -0,0 +1,225 @@
...
+#'   solid lines, while uncertain occurrences will be plotted with a white
+#'   circle and joined with dashed lines.
+#' @param by \code{character}. How should the output be sorted? Either: "FAD"
+#'   (first appearance; default), "LAD" (last appearance), or "name"
+#'   (alphabetically by taxon names).
+#' @param plot_args A list of optional arguments that are passed directly to
+#'   [graphics::plot()]. Useful arguments include `xlab` (the x-axis label),
+#'   `ylab` (the y-axis label, default is "Bed number"), `main` (the plot
+#'   title), `xlim` (the x-axis limits), and `ylim` (the y-axis limits). The
+#'   `axes` and `type` arguments are not supported and will be overridden.
```
LewisAJones, line 28: I think there are quite a lot of additional arguments that are not supported, for example, `pch`, `cex`, `col`, etc? If the purpose of this function is to just generate a quick plot, I think it is fine if we just clarify this. However, if the point is to have something highly customisable that can be used for publications, we probably want to re-look at this. Based on #99, this might be something that users request.

LewisAJones (commented):
Hi @bethany-j-allen and @willgearty, this is looking great! I have a few final comments, but I think the function is basically there. 

Thanks again for all the work you've put in with this! I suspect it will be very handy for a lot of users.

tax_range_strat_font.svg
bethany-j-allen: I wanted to add a test for the plot customisability, and just accidentally saved this in the wrong place. But I don't think it will actually hit any of the "untested" code anyway so I removed it.

LewisAJones: Yeah, I only meant that it was in the top-level directory, good to have tests for such things!

R/tax_range_strat.R
bethany-j-allen: No, I think this would be appropriate for the vignette but is too detailed for the simple example here.

R/tax_range_strat.R
bethany-j-allen: Surely you could use it if you really wanted to, you would just need to be careful about the placement?

R/tax_range_strat.R
bethany-j-allen: I think this is something we can revisit at a later date, if/when someone raises it.

LewisAJones: Sure, I am happy with increasing customisability later. However, I think we still need to at least state which arguments can or cannot be used. Currently, the documentation suggests arguments available via `plot()` are compatible. If we don't do this, a lot of users will end up banging their head against the wall asking why very common arguments don't work (e.g. `col`). We might also get quite a few issues raised if this is not clear.

R/tax_range_strat.R
LewisAJones: I think it is a pain within the plot function to modify the location of the x-axis title. However, you could update the function by always adding the xlab after you've added the x-axis labels, e.g.:

```title(xlab = "Taxa", line = 4)```

willgearty: Yeah, it doesn't look very straightforward to modify the x-axis title via `plot()` (the `mgp` parameter seems closest, but it affects all of the titles, not just the x-axis title), and `axis()` doesn't have the option to plot the axis title, which is frustrating. If we call `title()` within the function, should we have yet another argument for the list of arguments that should be passed to that (e.g., title_args)? Another option would be to just show in the example how a user might use the `title()` function after calling `tax_range_strat()`?

R/tax_range_strat.R
willgearty: I think the only arguments that are not currently supported are those related to the styling of the points and segments. It looks like these are all hard-coded at the moment. We could make these individual arguments for the function (although would we want different col and bg arguments for the certain and uncertain points?). Another option is to just have a list of arguments that would be passed to `points()` (again, would we want two arguments, one for the certain `points()` call, one for the uncertain `points()` call?). Yet another option is to have an argument that is a list of graphical parameters. In the function we could then save the old `par()`, set the new values with this argument, plot the points, then reload the old `par()`. And of course, a final option is to just report in the documentation that these things are fixed and can not be changed.

R/tax_range_strat.R
LewisAJones: I think the most straight forward thing would be just to use `title()` and calculate a value for line argument based on the  `nchar()` of the longest species names on x? Unless there is a way to extract the exact coordinates of plotted labels? I don't think we necessarily need to make all arguments available from `title()` as we are just treating it as our "xlab". However, given that most users might not even want an xlab, I think your suggestion of showing how one might use `title()` is probably OK too?

R/tax_range_strat.R
willgearty: I'm not sure that solves the problem of how to handle the certain vs uncertain point and line types? But if we're not terribly worried about that, then I'm fine pulling these arguments from `plot_args` if they are specified.

R/tax_range_strat.R
willgearty: I don't know much about base R and getting the coordinates of plotted labels, but it sounds complicated. I'm in favor of just including an example showing how to use `title()` (and maybe adding some text in the `x_args` documentation that says that users might want to use `title()` instead of supplying `xlab`).

R/tax_range_strat.R
LewisAJones: What if by default the `col`, `bg`, `lty`, etc arguments are a vector of length 2, with the first element relating to certain occurrences and the second uncertain occurrences?

willgearty:
OK, that should address both of those points @LewisAJones. Let me know what you think.

LewisAJones:
Thanks for the updates @bethany-j-allen and @willgearty, I think this is good to go now! We just need to update the `NEWs.md` and increase test coverage (currently less than the desired 90–95%). I've approved the PR but please implement before merging. Great job and thanks for putting up with my picky review! 😅

LewisAJones (approved):

PR #105, opened 2023-12-21
Add warning to bin_time

### LINKED ISSUE #104: Midpoint age falling right on boundary
When running bin_time(method = "mid"), occurrences with a mid_ma that fall right on the boundary of two stages get an NA for their bin assignment. Is there a reasonable way to address this? If not, should the user be alerted to the issue?

image

LewisAJones:
This PR adds a warning to address #104:

> "When running bin_time(method = "mid"), occurrences with a mid_ma that fall right on the boundary of two stages get an NA for their bin assignment. Is there a reasonable way to address this? If not, should the user be alerted to the issue?"

I've opted for a warning here to let the user know what is going on. Another option would be to always bin to the younger or older interval (or both). However, I think this is undesirable and the user should take this decision for themselves. 

Closes #104.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,8 @@
+# palaeoverse (development version)
+
+* Added warning to bin_time (#104)
+* Improved link accessibility (#88)
+
 # palaeoverse 1.2.1
 
 * Fixed handling of multiple models in palaeorotate (#92)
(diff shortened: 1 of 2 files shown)


ChristopherDavidDean (approved):
Happy for this inclusion; this is the right way to deal with the problem (e.g. warning the user to data issues rather than making decisions for them). A simple solution too.

LewisAJones:
> Happy for this inclusion; this is the right way to deal with the problem (e.g. warning the user to data issues rather than making decisions for them). A simple solution too.

Thanks Chris!!!

PR #107, opened 2024-02-12
add customisability functionality

### LINKED ISSUE #99: Greater plot customisability in tax_range_time
Kateryn Pino on our Google Group requested additional customisability of plots in `tax_range_time`. This might be desirable for some users to refine plots for publications.

Basically, we should allow greater customisability. I guess we could pass standard plot arguments using `...` - it seems it would be the cleanest way.

An alternative would be to set up a number of expected customisable arguments (e.g. colour, title, etc), but this seems unnecessary and a lot more work.

**Additional context**
Nothing to add.

@KEichenseer @ChristopherDavidDean happy to review if I put a PR together?

LewisAJones:
This PR adds additional functionality to `tax_range_time` for customising the plot appearance and resolving #99.

The function can now be supplied with a list of `plot_args` for customising plotted points and lines (colour, size, type, etc). I have also now allowed users to specify the type of `intervals` they want to add via `axis_geo` to the plot. For example, users can now choose to add stages and periods.

As the original reviewers @bethany-j-allen and @ChristopherDavidDean, would you be willing to review these changes?

Closes #99.


R/tax_range_time.R
```diff
@@ -18,6 +18,17 @@
 #' Either: "FAD" (first-appearance date; default), "LAD" (last-appearance data),
 #' or "name" (alphabetically by taxon names).
 #' @param plot \code{logical}. Should a plot of the ranges be generated?
+#' @param plot_args \code{list}. A list of optional arguments that are used
```
bethany-j-allen, line 21: Maybe `arguments relevant to plotting`?

R/tax_range_time.R
```diff
@@ -18,6 +18,17 @@
...
+#'   for plotting. See Details for options.
+#' @param intervals \code{character}. The interval information to use to plot
```
bethany-j-allen, line 23:
```suggestion
#' @param intervals \code{character}. The time interval information used to plot
```

R/tax_range_time.R
```diff
@@ -18,6 +18,17 @@
...
+#'   the axis: either A) a \code{character} string indicating a rank of
```
bethany-j-allen, line 24:
```suggestion
#'   the x-axis: either A) a \code{character} string indicating a rank of
```

R/tax_range_time.R
```diff
@@ -26,15 +37,25 @@
...
 #'
 #' @details The temporal range(s) of taxa are calculated by extracting all
-#' unique taxa (`name` column) from the input `occdf`, and checking their first
-#' and last appearance. The temporal duration of each taxon is also calculated.
-#' A plot of the temporal range of each taxon is also returned if `plot = TRUE`.
-#' If the input data columns contain NAs, these should be removed prior to
-#' function call.
+#'   unique taxa (`name` column) from the input `occdf`, and checking their
+#'   first and last appearance. The temporal duration of each taxon is also
+#'   calculated. If the input data columns contain NAs, these should be
```
bethany-j-allen, line 42:
```suggestion
#'   calculated. If the input data columns contain NAs, these must be
```

R/tax_range_time.R
```diff
@@ -87,6 +115,10 @@ tax_range_time <- function(occdf,
     stop('`by` must be either "FAD", "LAD", or "name"')
   }
 
+  if (!is.null(plot_args) && !is.list(plot_args)) {
+    stop("`plot_args` must be either NULL, or a list")
```
bethany-j-allen, line 119: Maybe add a test here - it should be easy to do.

bethany-j-allen (approved):
Hi Lewis, it seems to work well! I've suggested some wording changes to the documentation but otherwise it looks great.

LewisAJones:
Thanks @bethany-j-allen, I’ve gone ahead and implemented all your suggestions. @ChristopherDavidDean will you have chance to take a look over this soon? We could maybe get this into the new version which I hope to release early next week.

KEichenseer (approved):
Great job on adding that functionality! All seems to work as intended. 

A few comments: 

- an argument `name` is required for the input, but in the output this column is called "taxon", irrespective of the content of `name` - should this be changed to either be always "name" or "taxon", or, probably best, the content of the `name` argument?
- a slight simplification could be to allow the graphical parameters being directly specified via `...`, rather than being contained in a list for `plot_args`. I realise however that this has disadvantages as well, so maybe the current approach is fine.
- currently, some graphical parameters such as `xlim` or `xpd` are not supported. Perhaps a future expansion could be to allow any graphical parameters being used by using e.g. `do.call` for `plot` and `points` to pass any list of argument from `...` or `plot_args` to these function. But, that requires some fiddling with setting custom defaults so maybe not worth it.

LewisAJones:
> Great job on adding that functionality! All seems to work as intended.

Hi Kilian, thanks for the quick review on this!

> 
> A few comments:
> 
> * an argument `name` is required for the input, but in the output this column is called "taxon", irrespective of the content of `name` - should this be changed to either be always "name" or "taxon", or, probably best, the content of the `name` argument?

Hmmm, I kind of agree that it would be a great addition to extract the column name of `name`, and that we are currently only accounting for taxonomic units. However, this is currently consistent in functions throughout palaeoverse (e.g. `tax_range_strat`) and is maybe something best thinking about updating throughout the package in a later version?

> * a slight simplification could be to allow the graphical parameters being directly specified via `...`, rather than being contained in a list for `plot_args`. I realise however that this has disadvantages as well, so maybe the current approach is fine.

This is currently implemented as `plot_args` to be consistent with `tax_range_strat`. There was some discussion on that PR about standardising plot input arguments across functions going forward and this is what we decided on. I do agree it would be more convenient to pass it directly to `...` but will leave it as it is for now.

> * currently, some graphical parameters such as `xlim` or `xpd` are not supported. Perhaps a future expansion could be to allow any graphical parameters being used by using e.g. `do.call` for `plot` and `points` to pass any list of argument from `...` or `plot_args` to these function. But, that requires some fiddling with setting custom defaults so maybe not worth it.

Currently such graphical parameters are not supported as the data to be plotted are driving these limits. I've tried to make it clear in the Details section which arguments are supported, but let me know if anything is unclear!

Thanks again for the quick review!

PR #108, opened 2024-02-13
Add age arguments to bin_time

### LINKED ISSUE #106: Age arguments for bin_time
@ChristopherDavidDean suggested that we should add additional arguments for allowing different name columns for age information in `occdf` for `bin_time()`. This could be something along the lines of two arguments `max_ma` and `min_ma` as in `tax_expand_time()` or `max_age` and `min_age` with defaults set to "max_ma" and "min_ma". I am in favour of sticking with `max_ma` and `min_ma` as in the rest of the package as it puts emphasis on creating a standard unit throughout the package. However, I can also understand the desire for being more flexible here.

LewisAJones:
This PR adds age arguments (`max_ma` and `min_ma`) to the `bin_time` function. This allows users to easily specify different column names for their age data. Rather than having many new additional arguments, I have updated the function so that these `max_ma` and `min_ma` also relate to `bins`. Currently, users will still need to have a "bin" column in `bins`. Do we also want to add an argument for that to allow modification? 

As one of the original developers and reviewers, would you be willing to review @ChristopherDavidDean and @willgearty?

Closes #106.

### DIFF
--- a/R/bin_time.R
+++ b/R/bin_time.R
@@ -8,11 +8,18 @@
 #'   "max_ma" and "min_ma". These columns should contain `numeric` values. If
 #'   required, `numeric` ages can be generated from interval names via the
 #'   \code{\link[palaeoverse:look_up]{look_up()}} function.
+#' @param min_ma \code{character}. The name of the column you wish to be
+#'   treated as the minimum age for `occdf` and `bins`, e.g. "min_ma"
+#'   (default).
+#' @param max_ma \code{character}. The name of the column you wish to be
+#'   treated as the maximum age for `occdf` and `bins`, e.g. "max_ma"
+#'   (default).
 #' @param bins \code{dataframe}. A dataframe of the bins that you wish to
 #'   allocate fossil occurrences to such as that returned by
 #'   \code{\link[palaeoverse:time_bins]{time_bins()}}. This dataframe must
-#'   contain at least the following named columns: "bin", "max_ma" and
-#'   "min_ma". Columns "max_ma" and "min_ma" must be `numeric` values.
+#'   contain at least the following named columns: "bin" and those specified
+#'   to `max_ma` (default: "max_ma") and `min_ma` (default: "min_ma).
+#'   Columns `max_ma` and `min_ma` must be `numeric` values.
 #' @param method \code{character}. The method desired for binning fossil
 #'   occurrences. Currently, five methods exist in this function: "mid",
 #'   "majority", "all", "random", and "point". See Details for a description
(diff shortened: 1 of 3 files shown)


R/bin_time.R
```diff
@@ -302,7 +304,7 @@ bin_time <- function(occdf, bins, method = "mid", reps = 100,
         occdf$bin_assignment[id_vec] <- bin_list[[i]]
       }
       # Add bin midpoints to dataframe
-      for (i in seq_len(nrow(occdf))) {
+      for (i in seq_len(nrow(bins))) {
```
ChristopherDavidDean, line 307: Any particular reason this has been changed? Sorry, can't figure it out immediately.

ChristopherDavidDean (approved):
Happy with these changes, all seem simple and elegant. Updated descriptions are clear too. I've left a comment in the only place I'm unsure of, but otherwise I'm happy for this to go ahead.

R/bin_time.R
LewisAJones: This seems to have been a bug? Previously we were using the row number of the `occdf` to index through bins, but shouldn't it be the bin?

ChristopherDavidDean: Yes, you're correct; done a few tests and changing this increases speed by a large amount. Nice catch!

willgearty: What terrible reviewer let that bug through?! /s

R/bin_time.R
```diff
@@ -114,8 +123,8 @@ bin_time <- function(occdf, bins, method = "mid", reps = 100,
     if (is.data.frame(bins) == FALSE) {
       stop("`bins` should be a dataframe.")
     }
-    if (any(is.na(occdf$max_ma)) || any(is.na(occdf$min_ma))) {
-      stop("NA values detected in occdf$max_ma or occdf$min_ma.")
+    if (any(is.na(occdf[, max_ma])) || any(is.na(occdf[, min_ma]))) {
+      stop(paste("NA values detected in", max_ma, "or", min_ma))
     }
```
willgearty, line 128: Perhaps outside the scope of this PR, but I wonder if it would be more user friendly to have an `na.action` argument that would specify how to proceed in these cases? For example, "na.omit" would just remove these rows and continue, whereas "na.fail" would raise this error.

R/bin_time.R
LewisAJones: Unfortunately, we're all human 😅

LewisAJones:
Thanks @ChristopherDavidDean and @willgearty for the quick review! Looks like we can ✅ another issue off 😄

PR #113, opened 2024-03-25
Fix for GPlates API response

### LINKED ISSUE #112: GPlates API now returns original coordinates instead of NA when outside age scope
When a user submits coordinates to `palaeorotate()` with an age outside of the scope of a given model, it appears the GPlates API now returns the original coordinates instead of NA:

``` r
library(palaeoverse)
occdf <- data.frame(lng = c(2, -103, -66),
                    lat = c(46, 35, -7),
                    age = c(88, 125, 500))
palaeorotate(occdf = occdf, method = "point", model = c("PALEOMAP", "MULLER2016", "SETON2012"))
#> PALEOMAP
#> MULLER2016
#> SETON2012
#>    lng lat age p_lng_PALEOMAP p_lat_PALEOMAP p_lng_MULLER2016 p_lat_MULLER2016
#> 1    2  46  88         3.3803        36.0574           1.5367          38.9498
#> 2 -103  35 125       -54.3464        34.2684         -57.6315          28.7241
#> 3  -66  -7 500      -115.9991       -63.7982         -66.0000          -7.0000
#>   p_lng_SETON2012 p_lat_SETON2012 range_p_lat max_dist
#> 1          5.7160         40.8531      4.7957  569.989
#> 2        -47.2065         29.6023      5.5443 1018.511
#> 3        -66.0000         -7.0000     56.7982 7441.814
```

<sup>Created on 2024-03-11 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

We throw a warning when the returned coordinates are NAs, but right now we do not throw any warning when the returned coordinates are the same as the submitted coordinates.

LewisAJones:
This is a quick fix for the new GPlates API response (see #112) which now returns 400 Bad Request when age of rotation exceeds the model age. As part of this fix, I have had to remove `RETRY` which was automatically retrying `GET` requests when there was a bad request. We might want to include `RETRY` again in future but this would basically require subsetting all points with an age of rotation that exceeds the specified model first. However, after doing some tests, it doesn't seem `RETRY` is strictly necessary, and we haven't hit any problems with `rphylopic` and using just `GET` yet neither... something to keep in mind though.

Fixes #112.


tests/testthat/test-palaeorotate.R
```diff
@@ -44,6 +44,29 @@ test_that("palaeorotate() point method works", {
...
+  )
+
+  expect_true(
+    all(
+      is.na(
+        palaeorotate(occdf = occdf,
+                     method = "point",
+                     model = c("SETON2012", "MULLER2016"),
+                     uncertainty = TRUE)$max_dist
+      )
```
willgearty, line 67: Hmm...is this the desired behavior? Or should we use `na.rm = T` in our calculations?

R/palaeorotate.R
```diff
@@ -428,7 +428,7 @@ palaeorotate <- function(occdf, lng = "lng", lat = "lat", age = "age",
     # Define root
     pbase <- "https://gws.gplates.org/reconstruct/reconstruct_points/"
     # Setup query structure
-    query <- list(points = NULL, time = NULL,
+    query <- list(time = NULL, lons = NULL, lats = NULL,
```
willgearty, line 431: Is this due to an API change?

willgearty (commented):
This seems great, thanks for putting this together so quickly @LewisAJones! Just a few comments/questions before approval...

tests/testthat/test-palaeorotate.R
LewisAJones: None of these models extend back to the opted age of rotation which is why NAs are returned here... in the function we only return NAs if one or no models are available for comparison

R/palaeorotate.R
LewisAJones: Not really, the `points` argument is still available but I think this is cleaner for specifying coordinates and we don't need to mess around with generating the combined string....

tests/testthat/test-palaeorotate.R
willgearty: Ah, I had only tried two models (one of which had an invalid age)! Upon further investigation this seems to work well for 2+ models with valid gplates numbers.

willgearty (approved):

PR #116, opened 2024-05-06
Fix point bug in tax_range_strat

bethany-j-allen:
This branch fixes a bug in tax_range_strat().
At present, the code for plotting the points references a column named `bed`, and if an alternative name is used (e.g. `height`) then the column is not identified. This fix instead pulls the column name based on the input `level`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# palaeoverse 1.3.1
+
+* Fixed point plotting bug in tax_range_strat
+
 # palaeoverse 1.3.0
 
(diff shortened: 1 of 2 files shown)


willgearty (approved):
Looks good to me, thanks for the quick fix @bethany-j-allen!

willgearty:
Don't forget to add documentation to NEWS.md!

LewisAJones:
Looks fine from my end! It really was just three lines of code 😆

LewisAJones (approved):

PR #122, opened 2024-09-18
tax_range_time/strat grouping

bethany-j-allen:
This is a draft of new code corresponding to the request we received from @RussellGarwood. A grouping variable can now be supplied to either tax_range_time or tax_range_strat and used to sort the taxa.
I realised one issue which I'm not sure how to resolve, namely conflicts between the `name` and `group`. For example, what group does "cat" belong in, if the first row with a "cat" in says the cat belongs to "mammals", but the second row with a "cat" in says the cat belongs to "vertebrates". I set up the code to take the grouping from the first instance (smallest row number) of the taxon, with the grouping given in any other row ignored. This seems reasonable to me, but I think we should add a warning in the documentation if this is the approach we decide to take.

### DIFF
--- a/R/tax_range_strat.R
+++ b/R/tax_range_strat.R
@@ -8,6 +8,8 @@
 #'   `name` and `level` arguments).
 #' @param name \code{character}. The name of the column you wish to be treated
 #'   as the input names, e.g. "genus" (default).
+#' @param group \code{character}. In the case of using "by = "group"", the name
+#' of the column you wish to be treated as the grouping variable, e.g. "family".
 #' @param level \code{character}. The name of the column you wish to be treated
 #'   as the stratigraphic levels associated with each occurrence, e.g. "bed"
 #'   (default) or "height". Stratigraphic levels must be \code{numeric}.
@@ -19,8 +21,8 @@
 #'   solid lines, while uncertain occurrences will be plotted with a white
 #'   circle and joined with dashed lines.
 #' @param by \code{character}. How should the output be sorted? Either: "FAD"
-#'   (first appearance; default), "LAD" (last appearance), or "name"
-#'   (alphabetically by taxon names).
+#'   (first appearance; default), "LAD" (last appearance), "name"
+#'   (alphabetically by taxon names), or "group"(an additional variable).
 #' @param plot_args A list of optional arguments that are passed directly to
 #'   [graphics::plot()]. Subsets of these arguments are also passed to
 #'   [graphics::segments()] and [graphics::points()] (see Details). Useful
@@ -104,7 +106,7 @@
 #' title(xlab = "Taxon", line = 10.5)
 #'
 #' @export
-tax_range_strat <- function(occdf, name = "genus", level = "bed",
+tax_range_strat <- function(occdf, name = "genus", group = NULL, level = "bed",
                             certainty = NULL, by = "FAD", plot_args = NULL,
                             x_args = NULL, y_args = NULL) {
 
@@ -140,8 +142,16 @@ tax_range_strat <- function(occdf, name = "genus", level = "bed",
     stop("The `level` column contains NA values")
   }
 
-  if (!by %in% c("name", "FAD", "LAD")) {
-    stop("`by` must be either \"FAD\", \"LAD\", or \"name\"")
+  if (!by %in% c("name", "FAD", "LAD", "group")) {
+    stop("`by` must be either \"FAD\", \"LAD\", \"name\" or \"group\"")
+  }
+
+  if (by == "group" && is.null(group)) {
+    stop('`group` variable must be provided to enable filtering by group')
+  }
+
+  if (by == "group" && (group %in% colnames(occdf) == FALSE)) {
+    stop('`group` is not a named column in `occdf`')
   }
 
   #List and order unique taxa
@@ -150,9 +160,10 @@ tax_range_strat <- function(occdf, name = "genus", level = "bed",
 
   #Create object to hold information
   if (is.null(certainty)) {
-    ranges <- data.frame(taxon = unique_taxa, min_bin = NA, max_bin = NA)
+    ranges <- data.frame(taxon = unique_taxa, group = NA, min_bin = NA,
+                         max_bin = NA)
   } else {
-    ranges <- data.frame(taxon = unique_taxa, min_bin = NA,
+    ranges <- data.frame(taxon = unique_taxa, group = NA, min_bin = NA,
                          max_bin = NA, min_bin_certain = NA,
                          max_bin_certain = NA)
   }
@@ -160,16 +171,17 @@ tax_range_strat <- function(occdf, name = "genus", level = "bed",
   #Populate nested list
   for (i in seq_along(unique_taxa)) {
     occ_filter <- occdf[(occdf[, name] == unique_taxa[i]), ]
-    ranges[i, 2] <- min(occ_filter[level])
-    ranges[i, 3] <- max(occ_filter[level])
+    ranges[i, 2] <- occ_filter[1, group]
+    ranges[i, 3] <- min(occ_filter[level])
+    ranges[i, 4] <- max(occ_filter[level])
     #If uncertainty is used, fill second set of columns for certain IDs
     if (!is.null(certainty)) {
       occ_filter <- occ_filter[(occ_filter[, certainty] == 1), ]
       if (nrow(occ_filter) == 0) {
         occ_filter[1, ] <- NA
       }
-      ranges[i, 4] <- min(occ_filter[level])
-      ranges[i, 5] <- max(occ_filter[level])
+      ranges[i, 5] <- min(occ_filter[level])
+      ranges[i, 6] <- max(occ_filter[level])
     }
   }
 
@@ -182,6 +194,10 @@ tax_range_strat <- function(occdf, name = "genus", level = "bed",
   if (by == "FAD") {
     ranges <- ranges[order(ranges$max_bin), ]
     ranges <- ranges[order(ranges$min_bin), ]
+  }
+
+  if (by == "group") {
+    ranges <- ranges[order(ranges$group), ]
   }
 
   #Add ID numbers
--- a/R/tax_range_time.R
+++ b/R/tax_range_time.R
@@ -14,9 +14,11 @@
 #' as the minimum limit of the age range, e.g. "min_ma" (default).
 #' @param max_ma \code{character}. The name of the column you wish to be treated
 #' as the maximum limit of the age range, e.g. "max_ma" (default).
+#' @param group \code{character}. In the case of using "by = "group"", the name
+#' of the column you wish to be treated as the grouping variable, e.g. "family".
 #' @param by \code{character}. How should the output be sorted?
 #' Either: "FAD" (first-appearance date; default), "LAD" (last-appearance data),
-#' or "name" (alphabetically by taxon names).
+#' "name" (alphabetically by taxon names), or "group"(an additional variable).
 #' @param plot \code{logical}. Should a plot of the ranges be generated?
 #' @param plot_args \code{list}. A list of optional arguments relevant to
 #'   plotting. See Details for options.
@@ -69,8 +71,12 @@
 #' occdf <- subset(occdf, !is.na(order) & order != "NO_ORDER_SPECIFIED")
 #' # Temporal range
 #' ex <- tax_range_time(occdf = occdf, name = "order", plot = TRUE)
+#' # Temporal range ordered by class
+#' ex <- tax_range_time(occdf = occdf, name = "order", group = "class",
+#'                      by = "group", plot = TRUE)
 #' # Customise appearance
-#' ex <- tax_range_time(occdf = occdf, name = "order", plot = TRUE,
+#' ex <- tax_range_time(occdf = occdf, name = "order", group = "class",
+#'                      by = "group", plot = TRUE,
 #'                      plot_args = list(ylab = "Orders",
 #'                                       pch = 21, col = "black", bg = "blue",
 #'                                       lty = 2),
@@ -80,6 +86,7 @@ tax_range_time <- function(occdf,
                            name = "genus",
                            min_ma = "min_ma",
                            max_ma = "max_ma",
+                           group = NULL,
                            by = "FAD",
                            plot = FALSE,
                            plot_args = NULL,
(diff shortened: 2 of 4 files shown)


LewisAJones:
Hi @bethany-j-allen, thanks for putting this together! I probably won't have time to check over this properly until early next week. However, a quick couple of conceptual points:
- Does "group" need to be an option in `by`? Could we group and sort? Is there value in that?
- The issue you note is a little tricky. I agree with adding a warning in such instances. For the handling of this, I think it probably makes most sense to either assign it to the most common group, or better still, treat them as separate entities belonging to different groups to avoid potential issues with homonyms?

R/tax_range_strat.R
```diff
@@ -8,6 +8,8 @@
 #'   `name` and `level` arguments).
 #' @param name \code{character}. The name of the column you wish to be treated
 #'   as the input names, e.g. "genus" (default).
+#' @param group \code{character}. In the case of using "by = "group"", the name
+#' of the column you wish to be treated as the grouping variable, e.g. "family".
```
willgearty, line 12: I think this should be right after the definition of "by". Also, is this argument necessary? If the user supplies something other than "FAD", "LAD", or "name", we could just assume that is the name of the other variable?

bethany-j-allen:
I have separated 'group' and 'by' as requested - I didn't do this with `group_apply()` because it seemed like extra work to make that plot-compatible, so I just adapted the previous code.
The question about how to allocate groups is outstanding - evaluating the most common grouping applied to each taxon would require much more code than the current (first) option. There might be a few circumstances where you would want to genuinely allocate taxa to multiple groupings, but I think mistakes are probably our primary concern, with people wanting to do something more complex able to write their own code (or come up with a more sophisticated taxon/grouping system).
I'm happy to make additional changes but will also double-check everything and improve the test coverage regardless.

R/tax_range_time.R
```diff
@@ -157,7 +173,12 @@ tax_range_time <- function(occdf,
     if (by == "LAD") {
       temp_df <- temp_df[order(temp_df$min_ma), ]
       temp_df$taxon_id <- seq_len(nrow(temp_df))
-      }
+    }
+
+    if (is.null(group)) {
```
LewisAJones, line 178:
```suggestion
    if (!is.null(group)) {
```
Is the "!" missing here?

R/tax_range_time.R
```diff
@@ -132,6 +144,7 @@ tax_range_time <- function(occdf,
                         taxon_id = seq(1, length(unique_taxa), 1),
                         max_ma = rep(NA, length(unique_taxa)),
                         min_ma = rep(NA, length(unique_taxa)),
+                        group = rep(NA, length(unique_taxa)),
```
LewisAJones, line 147: Should we add some cleanup at the end to remove this column if `group = NULL`? Otherwise we're just getting a column of NAs.

R/tax_range_time.R
```diff
@@ -157,7 +173,12 @@ tax_range_time <- function(occdf,
     if (by == "LAD") {
       temp_df <- temp_df[order(temp_df$min_ma), ]
       temp_df$taxon_id <- seq_len(nrow(temp_df))
-      }
+    }
+
+    if (is.null(group)) {
+      temp_df <- temp_df[order(temp_df$group), ]
```
LewisAJones, line 179: Currently, we're ordering by `group` after ordering `name` by how the user specified. I think when group is supplied, we should be ordering using `by` for `name` and `group` (presumably `group` first), e.g. `group` are all organised by FAD, and then name within `group` are organised by FAD too.

R/tax_range_strat.R
```diff
@@ -184,6 +197,10 @@ tax_range_strat <- function(occdf, name = "genus", level = "bed",
     ranges <- ranges[order(ranges$min_bin), ]
   }
 
+  if (!is.null(group)) {
+    ranges <- ranges[order(ranges$group), ]
```
LewisAJones, line 201: Same comment as with `tax_range_time()`

Currently, we're ordering by `group` after ordering `name` by how the user specified. I think when group is supplied, we should be ordering using `by` for `name` and `group` (presumably `group` first), e.g. `group` are all organised by FAD, and then name within `group` are organised by FAD too.

R/tax_range_strat.R
```diff
@@ -104,14 +107,18 @@
 #' title(xlab = "Taxon", line = 10.5)
 #'
 #' @export
-tax_range_strat <- function(occdf, name = "genus", level = "bed",
+tax_range_strat <- function(occdf, name = "genus", group = NULL, level = "bed",
```
LewisAJones, line 110: We have a lot of code for examples already for this function - is it worth having an example using group?

R/tax_range_strat.R
```diff
@@ -150,26 +157,32 @@ tax_range_strat <- function(occdf, name = "genus", level = "bed",
 
   #Create object to hold information
   if (is.null(certainty)) {
-    ranges <- data.frame(taxon = unique_taxa, min_bin = NA, max_bin = NA)
+    ranges <- data.frame(taxon = unique_taxa, group = NA, min_bin = NA,
```
LewisAJones, line 160: Should we add some cleanup at the end to remove this column if group = NULL? Otherwise we're just getting a column of NAs.

R/tax_range_time.R
```diff
@@ -157,7 +173,12 @@ tax_range_time <- function(occdf,
...
+      temp_df$taxon_id <- seq_len(nrow(temp_df))
+    }
 
     # Plot data?
     if (plot == TRUE) {
```
LewisAJones, line 184: When `group` is supplied, is it worth plotting different groups by different colours?

LewisAJones (changes requested):
Hi @bethany-j-allen, thanks for implementing these changes. I definitely think it is an improvement having group as a separate argument. I have a few small comments and suggested changes (see below). 

Relating to the question about how to allocate groups... I think this should really just be based on unique instances. If you have the same taxon names belonging to two different groups (as provided/specified by the user), I would allocate the associated occurrences to the two different groups. This might even be a way for users to spot mistakes in their data.

Perhaps out of the scope of this PR, but I was also wondering whether we eventually want to update `tax_range_time` to also plot the names of taxa rather than the numbers, as in `tax_range_strat`. Originally, I didn't implement the plotting like this because it quickly becomes difficult to see the names when many (> 10) are plotted. However, I guess users could play with `par()` to overcome this issue.

We should also update NEWS.md!

R/tax_range_time.R
willgearty: Ah, that would explain why the `group` variable wasn't changing the plot for me.

willgearty (changes requested):
I agree with Lewis, these changes are definitely a step in the right direction, thanks @bethany-j-allen! My main two concerns right now are as follows:

1. We aren't really visually indicating the groups in any way. @LewisAJones has suggested using different colors, whereas I suggest using some sort of brackets. Either way, I think showing where groups start and stop is important, but we also need to somehow respect the custom styling that users can now supply. I know we've discussed before whether users should expect these plots to be hyper customizable and/or publication quality, but I think this would be useful even just for data exploration.
2. We are currently "allocating" group names to taxa based on the first group that they are found in. I think for simple situations this is fine (if each taxon only ever appears in one group this is perfect). However, as you mention @bethany-j-allen this wouldn't work so great if the same name appears in multiple groups. If `group` is a taxonomic variable, it's definitely concerning that the same name appears in multiple groups (but I agree with @LewisAJones that this could be a way for users to find errors). However, I think there are plenty of reasons why the same name could appear in multiple groups (e.g., if `group` is different localities or geographic regions). I know you are against using `group_apply()` @bethany-j-allen, but I think by performing our current sorting and calculations within each group (instead of on the whole group then assigning group names), that should resolve this whole issue. I'd be happy to implement `group_apply()` in both functions if that would help!

Here's a quick mockup of something for `tax_range_time()` which is the same number of lines as the current code and gives roughly the same output format:
```r
group_apply(occdf, c("class"), function(occdf, name, min_ma, max_ma) {
    unique_taxa <- unique(occdf[, name, drop = TRUE])
    # Order taxa
    unique_taxa <- unique_taxa[order(unique_taxa)]
    
    #=== Temporal range ===
    # Generate dataframe for population
    temp_df <- data.frame(taxon = unique_taxa,
                          taxon_id = seq(1, length(unique_taxa), 1),
                          max_ma = rep(NA, length(unique_taxa)),
                          min_ma = rep(NA, length(unique_taxa)),
                          range_myr = rep(NA, length(unique_taxa)),
                          n_occ = rep(NA, length(unique_taxa)))
    # Run for loop across unique taxa
    for (i in seq_along(unique_taxa)) {
        vec <- which(occdf[, name, drop = TRUE] == unique_taxa[i])
        temp_df$max_ma[i] <- max(occdf[vec, max_ma])
        temp_df$min_ma[i] <- min(occdf[vec, min_ma])
        temp_df$range_myr[i] <- temp_df$max_ma[i] - temp_df$min_ma[i]
        temp_df$n_occ[i] <- length(vec)
    }
    row.names(temp_df) <- NULL
    temp_df
}, "order", "min_ma", "max_ma")
```

R/tax_range_strat.R
LewisAJones: Yes, precisely

LewisAJones:
Hi @bethany-j-allen and @willgearty!

In the interest of trying to push this forward, I've done a little experimenting this morning with plotting options (only currently implemented in `tax_range_time`). What do you both think of something like this?:

```
# Grab internal data
occdf <- tetrapods
# Remove NAs
occdf <- subset(occdf, !is.na(order) & order != "NO_ORDER_SPECIFIED")
# Customise appearance
ex <- tax_range_time(occdf = occdf, name = "order", group = "class",
                     plot = TRUE,
                     plot_args = list(ylab = "Orders",
                                      pch = 21, col = "black", bg = "blue",
                                      lty = 2),
                     intervals = list("periods", "eras"))
```
image

I need to clean up the code a little more but I think it is a good start for what we've discussed here? I'm not sure how much customisability we want to allow but I think alternating light/dark grey works well in addition to the text labels.

willgearty:
This looks great, thanks for implementing my group_apply code and adding some other nice functionality! I think the colored rectangles look good, although I do worry that the dark grey is a bit too dark. Maybe "grey90" and "grey95" would be a better pair?

And I imagine these changes should also be applied to tax_range_strat? (I think the visualization changes would definitely be more useful there since you can actually see the taxon names in the plot)

LewisAJones:
> This looks great, thanks for implementing my group_apply code and adding some other nice functionality! I think the colored rectangles look good, although I do worry that the dark grey is a bit too dark. Maybe "grey90" and "grey95" would be a better pair?

Thanks for taking a look @willgearty! Sure, I can get behind that!

> And I imagine these changes should also be applied to tax_range_strat? (I think the visualization changes would definitely be more useful there since you can actually see the taxon names in the plot)

Sure! I just wanted to get thoughts before going ahead. Should `tax_range_time` also have taxon names? I can imagine it getting quickly unwieldy. 🤔

willgearty:
> Sure! I just wanted to get thoughts before going ahead. Should `tax_range_time` also have taxon names? I can imagine it getting quickly unwieldy. 🤔

Yeah, I originally thought about recommending this (and was actually surprised to see that this isn't the current behavior), but I think we did this because the plot for `tax_range_time` was supposed to be just a side-effect, and the important output is the data.frame that is returned, and vice versa for `tax_range_strat`?

R/tax_range_time.R
```diff
@@ -164,61 +168,74 @@ tax_range_time <- function(occdf,
...
     }
-    # Define lims
+    # Collect usr par for resetting
+    usrpar <- par(no.readonly = TRUE)
+    # Estimate max label width
+    max_label_width <- max(strwidth(temp_df$taxon, units = "inches"))
+    # Convert inches to lines (approximate conversion factor: 0.2)
+    extra_margin <- max_label_width / 0.2
+    # Set left margin (add extra space, default is 4)
+    par(mar = c(5, 4 + extra_margin, 4, 2))
```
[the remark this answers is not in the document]
LewisAJones: Yes. Yes we should!

LewisAJones:
Hi both, I think at this point I'm happy with `tax_range_time`. @bethany-j-allen are you OK with finishing up your changes to `tax_range_strat`, e.g. implementing `group_apply` and plotting by groups? Perhaps we can then do a cross-review!

LewisAJones:
Just bumping this, @bethany-j-allen do you think you will have chance to look into this soon?

bethany-j-allen:
Sorry for the delay, I'll try to get round to it this week!

bethany-j-allen:
I've removed the group column when there isn't one, and added an example. I think the last step will be merging the plotting code Lewis wrote for `tax_range_time` into what we already have for `tax_range_strat` - splitting the table into groups should also enable the sorting by group that you asked for above. I'm struggling to get my head around the plot code and how to do the merge though, I tried and failed a couple of times! If one of you has the time I think you could manage it more easily than me - otherwise I can have another go, but probably won't get round to this for another couple of weeks.

LewisAJones:
Thanks @bethany-j-allen - I've gone ahead and integrated `group_apply` for `tax_range_strat` as discussed and I've made an attempt to integrate plotting of groups. As both you and @willgearty put together this function initially, a good review of the changes I've made are probably needed to make sure I've not accidentally fudged something up 😆 . Let me know if you want to discuss anything (here or on Slack) - one thing we should consider is the placement of the labels for groups.

image

LewisAJones:
**Bumpity bump** 😄 
 
@bethany-j-allen or @willgearty would you be able to give this a once over in the next week or two?

R/tax_range_strat.R
```diff
@@ -102,20 +106,31 @@
...
+    stop("`occdf` should be a data.frame")
   }
 
   if (!is.numeric(occdf[, level, drop = TRUE])) {
     stop("`level` must be of class numeric")
   }
 
+  if (!is.null(group) && (group %in% colnames(occdf) == FALSE)) {
+    stop('`group` is not a named column in `occdf`')
+  }
```
willgearty, line 132: As @bethany-j-allen mentioned, this line breaks if more than one grouping column is supplied (e.g., `group = c("class", "order")`. I think it's fine to only allow a single grouping variable, but we should do a check to make sure that's a case.

willgearty: Same for the same code in `tax_range_time.R`

bethany-j-allen:
This looks fantastic to me, happy to merge - and will be glad to see this one finished!

PR #123, opened 2024-09-18
Add lat_bins_equal

### LINKED ISSUE #114: Equal area latitudinal bins
We were previously working on developing functionality to enable the creation of equal-area latitudinal bins. This should be implemented in the next version of the package.

LewisAJones:
# Function name

`lat_bins_equal()`

## Description

This function generates approximately equal-area latitudinal bins for a user-specified given number of bins and latitudinal range. This approach is based on calculating the curved surface area of spherical segments bounded by two parallel disks.

## Motivation and Context

Standard degree latitudinal bins are inherently uneven in area. This can result in more or less fossil occurrences (or samples) being assigned to a bin purely because it is larger or smaller in areal extent than other bins. 

I have opted to implement this as a separate function. However, I am open to discussion about this and keen to hear others' thoughts. I had originally planned to add this to `lat_bins()` but it would require reworking it quite a lot and after drafting the docs a few times, I felt it could be potentially confusing for users to have so many options/arguments in one function.

This PR addresses #114.

Closes #114.


KEichenseer (changes requested):
Hi @LewisAJones, thanks for this, it's a very useful function and works as intended. 

I agree with the choice of adding it as a separate function to keep it simple. We can link to `lat_bins()` in the documentation (and vice versa).

I suggest those changes, implemented in branch https://github.com/palaeoverse/palaeoverse/tree/equal-area-lat-bins-ke:
- I've changed the bin generation to generate bins that have precisely equal areas.
- A matter of preference but I've tripped over that when testing the function: I'd suggest ordering the arguments such that `min` appears before `max` (as e.g. in `runif()` and similar base R functions). I have changed also the `min` `max` order in the output and for the `lat_bins()` function.  - I just noticed that other functions in palaeoverse such as `tax_expand_lat`, too, have a `max` `min` order, so feel free to reject that.

LewisAJones:
Hi @KEichenseer - those were some great updates! Thanks a lot - the function is much faster now. I've merged the two branches and I agree with both your changes.

One outstanding thing... I previously had `r` specified in metres, but after looking through our other functions, distances are given in kilometres (e.g. `bin_space`). Do we want to standardise this here, or should everything else in fact be in metres too?

Note, tests seem to be failing at present for `palaeorotate()` due to a problem with the GPlates API.

KEichenseer:
> Hi @KEichenseer - those were some great updates! Thanks a lot - the function is much faster now. I've merged the two branches and I agree with both your changes.
> 
> One outstanding thing... I previously had `r` specified in metres, but after looking through our other functions, distances are given in kilometres (e.g. `bin_space`). Do we want to standardise this here, or should everything else in fact be in metres too?
> 
> Note, tests seem to be failing at present for `palaeorotate()` due to a problem with the GPlates API.

As far as I can see, "kilometres" are only used here and in `bin_space`, and "meters" are used nowhere. Both functions default to large km values (`100` and `6371`), so I think it's nicer to stick with kilometres rather than meters.

KEichenseer (approved):

R/lat_bins.R
```diff
@@ -20,6 +20,8 @@
 #' @param plot \code{logical}. Should a plot of the latitudinal bins be
 #' generated?
 #' @return A \code{dataframe} of latitudinal bins of user-defined size.
+#' @seealso
+#' For equal-area latitudinal bins, see \link{lat_bins_equal}.
```
bethany-j-allen, line 24: I wonder whether we could make the distinction between the two functions more clear in the names. Really `lat_bins()` is equal by degrees and `lat_bins_equal()` is equal by area - would `lat_bins_degrees()` and `lat_bins_area()` be better? Maybe you have a better idea.

R/lat_bins_equal.R
```diff
@@ -0,0 +1,114 @@
+#' Generate equal-area latitudinal bins
+#'
+#' A function to generate approximately equal-area latitudinal bins for a
+#' user-specified given number of bins and latitudinal range. This approach is
```
bethany-j-allen, line 4:
```suggestion
#' user-specified number of bins and latitudinal range. This approach is
```

R/lat_bins_equal.R
```diff
@@ -0,0 +1,114 @@
...
+#' based on calculating the curved surface area of spherical segments bounded
+#' by two parallel disks.
```
bethany-j-allen, line 6:
```suggestion
#' by two parallel discs.
```

R/lat_bins_equal.R
```diff
@@ -0,0 +1,114 @@
...
+#'
+#' @param n \code{numeric}. A single numeric value defining the number of
```
bethany-j-allen, line 8: `n` should be an integer rather than just numeric, right?

R/lat_bins_equal.R
```diff
@@ -0,0 +1,114 @@
...
+#'   equal-area latitudinal bins to split the latitudinal range by (as defined
```
bethany-j-allen, line 9:
```suggestion
#'   equal-area latitudinal bins to split the latitudinal range into (as defined
```

R/lat_bins_equal.R
```diff
@@ -0,0 +1,114 @@
...
+#'   by `min` and `max`).
+#' @param min \code{numeric}. A single numeric value defining the lower limit
+#'   of the latitudinal range (defaults to -90).
+#' @param max \code{numeric}. A single numeric value defining the upper limit
+#'   of the latitudinal range (defaults to 90).
+#' @param r \code{numeric}. The radius of the Earth in kilometres. Defaults to
+#'   the the mean radius of the Earth (6371 km).
```
bethany-j-allen, line 16: Do we need to expose `r`? I can't imagine a scenario where a user of our package would want to change this.

R/lat_bins_equal.R
```diff
@@ -0,0 +1,114 @@
...
+#' @param plot \code{logical}. Should a plot of the latitudinal bins be
+#'   generated?
+#' @return A \code{data.frame} of user-defined number of latitudinal bins.
```
bethany-j-allen, line 19: It would be good to add the column names here, I think. For completeness should we also add e.g. "if plot = TRUE, a plot is also returned." ?

bethany-j-allen (commented):
Hi both, this looks great and will no doubt be very useful. I have tested out the function and it seems to work well - I haven't checked the maths but I trust that you have already done this :) I have made some mostly cosmetic suggestions, happy to discuss any of them in further detail.

R/lat_bins_equal.R
LewisAJones: This is something I've thought about on a few occasions too (e.g. for `reps` in `bin_time()`). It is true that this number needs to be an integer (a whole number), but in terms of classes in R, this means the expected input would need to be the value followed by a L, e.g.
```
> is.integer(3)
[1] FALSE
> is.integer(3L)
[1] TRUE
```
Do we want to enforce this throughout the package or some class handling?

R/lat_bins_equal.R
LewisAJones: I don't think users will change this either tbh, but when I went on a bit of a deep dive of this, there are a few different values that individuals use. I think by giving a common default, we're being transparent about it and allowing users to change this if they really want to.

R/lat_bins.R
LewisAJones: I completely agree with you... I was just in two minds about how we treat the current `lat_bins()`... but for now, I've gone with:

- Updating `lat_bins_equal()` to `lat_bins_area()`
- Adding a `lat_bins_degrees()` alias for `lat_bins()`  
- Added a lifecycle badge and note for `lat_bins()` to say that the function has been deprecated and will be removed in the future

What do you think @bethany-j-allen - happy with those changes? @KEichenseer any additional thoughts here?

LewisAJones:
> Hi both, this looks great and will no doubt be very useful. I have tested out the function and it seems to work well - I haven't checked the maths but I trust that you have already done this :) I have made some mostly cosmetic suggestions, happy to discuss any of them in further detail.

Thanks for all the helpful comments @bethany-j-allen ! I've essentially implemented all your suggestions and commented where needed. 

Note: currently R CMD checks seem to be failing because the GPlates Web Service is having issues again. I'll re-run the checks tomorrow.

R/lat_bins.R
bethany-j-allen: This seems like a good solution to me!

R/lat_bins_equal.R
bethany-j-allen: Hmm. For me really this falls down to what the resultant behaviour is if someone tries to enter a non-integer number, so as long as the function handles this in a sensible way (that is clearly conveyed to the user), or throws an error saying the number should be an integer, it doesn't need to be formally handled as an integer by R.

R/lat_bins_equal.R
bethany-j-allen: Sure, so in theory someone could use a value, for example, with more decimal places. In that case, can we add some sort of citation to the parameter description for this value and/or state clearly that this allows numbers from other sources to be used?

bethany-j-allen (approved):
Thanks @LewisAJones!

R/lat_bins_equal.R
LewisAJones: Agreed – I'll add error handling for this.

R/lat_bins_area.R
```diff
@@ -0,0 +1,120 @@
...
+#'
+#' @param n \code{numeric}. A single numeric value defining the number of
+#'   equal-area latitudinal bins to split the latitudinal range into (as
+#'   defined by `min` and `max`).
+#' @param min \code{numeric}. A single numeric value defining the lower limit
+#'   of the latitudinal range (defaults to -90).
+#' @param max \code{numeric}. A single numeric value defining the upper limit
+#'   of the latitudinal range (defaults to 90).
+#' @param r \code{numeric}. The radius of the Earth in kilometres. Defaults to
+#'   the the mean radius of the Earth (6371 km).
```
willgearty, line 16:
```suggestion
#'   the mean radius of the Earth (6371 km).
```

R/lat_bins_degrees.R
```diff
@@ -0,0 +1,119 @@
+#' Generate latitudinal bins
```
willgearty, line 1: Should this be more specific now? Equal-width, maybe?

willgearty (approved):

PR #124, opened 2024-09-20
Updates to palaeorotate()

LewisAJones:
This is a first draft of an update to the `palaeorotate()` function. The aim of this PR is to:

- Remove mantle reference frame/hybrid Global Plate Models as they are not recommended for reconstructing 'true' palaeocoordinates
- Make the TorsvikCocks2017 Global Plate Model available to users
- Enable users to make use of the new reconstruction grids which have a temporal resolution of 1 million years (equivalent to the GPlates API) and are fully peer-reviewed and documented

Alongside doing this, I've tried to improve the function code and the (horribly written) tests.

@willgearty and @Buffan3369 I think you're both best positioned to review this, but let me know if you don't have time and I can try to find another reviewer.

Note: the GPlates Web Service seems to currently be having some issues, so this PR will likely fail R CMD Checks on occasion. To try to prevent the API being overloaded with requests, I've added a rest between the GET requests (this seems to be helping).


R/palaeorotate.R
```diff
@@ -1,41 +1,43 @@
 #' Palaeorotate fossil occurrences
 #'
-#' A function to generate palaeocoordinates for fossil occurrence data
+#' A function to estimate palaeocoordinates for fossil occurrence data
```
willgearty, line 3: 👍

R/palaeorotate.R
```diff
@@ -423,43 +362,34 @@ palaeorotate <- function(occdf, lng = "lng", lat = "lat", age = "age",
...
+                  GOLONKA = query)
     # Define GPlates model names
-    query[["MULLER2022"]]$model <- "MULLER2022"
     query[["MERDITH2021"]]$model <- "MERDITH2021"
-    query[["MULLER2019"]]$model <- "MULLER2019"
-    query[["MULLER2016"]]$model <- "MULLER2016"
     query[["PALEOMAP"]]$model <- "PALEOMAP"
+    query[["TorsvikCocks2017"]]$model <- "TorsvikCocks2017"
+    # anchor plate id 1 required for palaeomagnetic reference frame
+    query$TorsvikCocks2017$anchor_plate_id <- 1
```
willgearty, line 381: Not required for MATTHEWS2016_pmag_ref?

willgearty (approved):
Just some very minor comments/questions, looks great to me otherwise!

R/palaeorotate.R
LewisAJones: Oddly, MATTHEWS2016_pmag_ref and MATTHEWS2016_mantle_ref are implemented as two separate endpoints, and the anchor_plate_id does not need defining based on the [GWS docs](https://gwsdoc.gplates.org/models#matthews2016_pmag_ref).

R/palaeorotate.R
```diff
@@ -211,60 +173,53 @@ palaeorotate <- function(occdf, lng = "lng", lat = "lat", age = "age",
...
-  # Add warnings for use of mantle reference frame models
-  if (any(model %in% c("MULLER2022", "MULLER2019",
-                   "MATTHEWS2016_mantle_ref"))) {
-    warning(paste0("Selected model(s) use a mantle reference frame and are ",
-    "not recommended for reconstructing palaeocoordinates. See details."))
+  # Add stop for removed models
+  if (any(model %in% c("MULLER2022", "MULLER2019", "MULLER2016",
+                       "MATTHEWS2016_mantle_ref", "SETON2012"))) {
+    stop(paste0("Selected model(s) have recently been removed as they are not",
+    " in a palaeomagnetic reference frame. See details for available models."))
```
Buffan3369, line 217: I would maybe suggest displaying the problematic models. One possibility:
```
deprecated_models <- c("MULLER2022", "MULLER2019", "MATTHEWS2016_mantle_ref", "SETON2012")
  if (any(model %in% deprecated_models)) {
paste0("Selected model(s) have recently been removed as they are not",
       " in a palaeomagnetic reference frame: ", 
       paste(deprecated_models[which(model %in% deprecated_models)], collapse = ", "),
       ". See details for available models.")
}
```

R/palaeorotate.R
```diff
@@ -353,17 +291,10 @@ palaeorotate <- function(occdf, lng = "lng", lat = "lat", age = "age",
...
-    # Get available rotation ages
-    rot_age <- colnames(ref_model)[4:ncol(ref_model)]
-    rot_age <- unique(as.numeric(sub(".*_", "", rot_age)))
-
     # Calculate rotation ages for data
-    occdf$rot_age <- rot_age[sapply(seq_len(nrow(occdf)), function(i) {
-        which.min(abs(occdf[i, age] - rot_age))
-      })]
+    occdf$rot_age <- round(occdf[, age], digits = 0)
+    # Set 0 to 1
```
Buffan3369: doesn't work with zeros?

Buffan3369 (approved):
Nice job Lewis, just like Will, just a minor comment, everything works perfectly otherwise!

R/palaeorotate.R
LewisAJones: Yeah, that's fair enough! Thanks!

R/palaeorotate.R
LewisAJones: Sorry 😬 can you be a bit more specific? What doesn't work with zeros? The round function?

LewisAJones: Ahh it looks like you might have been reviewing an earlier commit and not the most recent one?

willgearty: Just checked, and this seems to work properly for zeros.

LewisAJones:
Thanks for the review @willgearty and @Buffan3369!

R/palaeorotate.R
Buffan3369: Yes sorry Lewis, just checked and in worked properely ^^'.

PR #125, opened 2024-09-21
Convert id certainty into binary vector

### LINKED ISSUE #109: Convert occurrence ID certainty into binary vector
Taxonomic uncertainty is usually expressed using specific punctuation or abbreviations within the written name. This makes uncertainty difficult to process computationally.

A function which reads a vector of taxon names and screens for e.g. ?, "", cf., aff., then creates a vector indicating whether each identification is certain or uncertain.

I already have some code to do this but it's fiddly and is likely a broadly useful tool.

**Additional context**
Ideally the vector produced should take the same format as "certainty" in tax_range_strat.

BrunaFarina:
# Function name
"tax_certainty"

## Description
The function screens a vector of taxa names and convert it into a binary vector – 1 is assigned to "certain" taxa (genus + especies name) and 0 is assigned to "uncertain" (genus only or genus + aff./cf./indet/?/etc).

## Motivation and Context
It addresses issue #109. The function deals with taxonomic uncertainty in written names returning to the user a vector of zeros and ones as the "certainty" argument in tax_range_strat.

- [x] I have provided sufficient examples of implementation (I hope so).
- [x] If it is a core feature, I have added thorough tests (I tried to run all tests).


LewisAJones:
Hi @BrunaFarina, this is just a quick note to say that I've just merged the main branch into this branch and updated `_pkgdown.yml` to stop the tests failing for reasons unrelated to your function!

R/tax_certainty.R
```diff
@@ -0,0 +1,33 @@
+#' Check if a taxon is certain or uncertain.
+#'
+#' Check whether a given taxon in a vector is certain or uncertain.
+#' Binomial names (genus and species) are considered "certain",
+#' whereas names containing only the genus,
+#' or genus and not determined species is considered "uncertain".
+#'
+#' @param taxon_names Character vector containing the species names to be
```
LewisAJones, line 8:
```suggestion
#' @param taxon_names \code{character}. A vector containing the species names to be
```
This is just a documentation style we tend to follow (which should probably be documented somewhere)

R/tax_certainty.R
```diff
@@ -0,0 +1,33 @@
...
+#' @export
+tax_certainty <- function(taxon_names) {
+
+  if (is.null(taxon_names) || !is.character(taxon_names) ||
+        length(taxon_names) == 0 || any(is.na(taxon_names))) {
+    stop("'taxon_names' must be a non-empty character vector, cannot be NULL,
+         and must not contain NA values")
+  }
+  match_patterns <- c("cf.", "aff.", "\\?", "^[A-Za-z]*$", "indet.", "sp.")
+  certainty <- apply(sapply(match_patterns, grepl, taxon_names), 1, any)
```
LewisAJones, line 29: This is a great use of grepl, but because we are using pattern matching here, it could lead to some problems such as any species names beginning with "sp" being caught, e.g. 
```
taxon_names <- c("Metaxytherium indet.", "Wether spoons")
tax_certainty(taxon_names)
```
For this reason, I would suggest using exact matching, or some hybrid approach.

R/tax_certainty.R
```diff
@@ -0,0 +1,33 @@
...
+#' (0).
+#' @examples
+#' sirenia_sp <- c("Metaxytherium indet.", "Felsinotherium_serresii",
+#'                  "Felsinotherium gervaisi", "Potamosiren indet.",
+#'                  "Potamosiren", "Metaxytherium sp.")
+#'
+#' certainty <- tax_certainty(sirenia_sp)
+#'
+#' @export
+tax_certainty <- function(taxon_names) {
```
LewisAJones, line 21: I am sure this wouldn't happen frequently, but I noticed that if for some reason, a user supplies a vector of just length 1, the function fails. e.g. 
```
taxon_names <- c("Metaxytherium indet.")
tax_certainty(taxon_names)
```

R/tax_certainty.R
```diff
@@ -0,0 +1,33 @@
...
+#' checked.
+#' It allows taxon names separated by underscore or space.
+#' @return Numeric vector classifying each species as certain (1) or uncertain
+#' (0).
+#' @examples
+#' sirenia_sp <- c("Metaxytherium indet.", "Felsinotherium_serresii",
+#'                  "Felsinotherium gervaisi", "Potamosiren indet.",
+#'                  "Potamosiren", "Metaxytherium sp.")
+#'
+#' certainty <- tax_certainty(sirenia_sp)
```
LewisAJones, line 18:
```suggestion
#' certainty <- tax_certainty(taxon_names = sirenia_sp)
```

R/tax_certainty.R
```diff
@@ -0,0 +1,33 @@
...
+#'
+#' Check whether a given taxon in a vector is certain or uncertain.
+#' Binomial names (genus and species) are considered "certain",
+#' whereas names containing only the genus,
+#' or genus and not determined species is considered "uncertain".
+#'
+#' @param taxon_names Character vector containing the species names to be
+#' checked.
+#' It allows taxon names separated by underscore or space.
+#' @return Numeric vector classifying each species as certain (1) or uncertain
```
LewisAJones, line 11: I think @bethany-j-allen envisioned this function as such to feed into `tax_range_strat` but I am wondering if as a standalone function it makes more sense to be logical (TRUE/FALSE)

R/tax_certainty.R
```diff
@@ -0,0 +1,33 @@
...
+#' (0).
+#' @examples
```
LewisAJones, line 13:
```suggestion
#' @section Developer(s):
#'     Bruna Farina
#' @section Reviewer(s):
#'     Lewis A. Jones
#' @examples
```
We normally try to keep track of this in the documentation to acknowledge one another and serve as a point of contact for any future updates.

R/tax_certainty.R
```diff
@@ -0,0 +1,33 @@
+#' Check if a taxon is certain or uncertain.
+#'
+#' Check whether a given taxon in a vector is certain or uncertain.
+#' Binomial names (genus and species) are considered "certain",
+#' whereas names containing only the genus,
```
LewisAJones, line 5: I guess more broadly, it is currently any name which is not binomial (so it could include family or order for example, not just genus?)

LewisAJones (changes requested):
Hi @BrunaFarina, I've just given the function a once over. I think you've done a great job getting all this together - you've got all the core components required (including tests!).

I've left some comments and suggested changes but I'd recommend waiting for @bethany-j-allen's input too before actioning anything (in case we disagree on anything). 

Once we've discussed here and you have integrated any changes that Bethany and I consider required, I'll give the function another review (this time including the tests and more extensively the documentation). 

Thanks for all your great work!

R/tax_certainty.R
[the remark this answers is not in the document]
willgearty: I'm going to veto this suggested change (not necessarily the comment) because by not assigning a value here, R automatically throws a useful error if an argument isn't provided.

R/tax_certainty.R
```diff
@@ -0,0 +1,33 @@
+#' Check if a taxon is certain or uncertain.
+#'
+#' Check whether a given taxon in a vector is certain or uncertain.
+#' Binomial names (genus and species) are considered "certain",
```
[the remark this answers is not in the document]
willgearty: Doesn't the "indet" regex catch higher level uncertainties already (e.g., Cetacea indet.)?

R/tax_certainty.R
willgearty: I think just changing the regex to "\\ssp\\.?$" should fix this (could make the same change to the other elements of the vector too).

R/tax_certainty.R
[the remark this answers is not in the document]
LewisAJones: That's true, but the current error handling is checking for a NULL input argument anyway which gives a more explicit message of what the input should be. That said, I am not going to lose any sleep over this being implemented or not - I can't imagine many users will forget to specify an input.

R/tax_certainty.R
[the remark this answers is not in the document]
LewisAJones: I don't think it is that common to have data structured as "Cetacea indet." in an occurrence dataframe. If we take the PBDB as a common example, and we were conducting some genus-level analyses, we'd probably either be using the genus column, or the accepted_name column and filtering for genus and species in the accepted_rank column. If using the genus column, you'd most likely have "" in the rows of occurrences that have only been identified to family-level or above. I guess my point, which is actually a question, is if this function should be compatible with working with different levels of taxonomic identification, i.e. can an input vector of genus names be given and we return a logical vector identifying whether we are certain or uncertain about those genus-level names? If we think this is worthwhile, we could have a secondary argument which controls the taxonomic level we're interested in.

R/tax_certainty.R
```diff
@@ -0,0 +1,33 @@
...
+#'
+#' @export
+tax_certainty <- function(taxon_names) {
+
+  if (is.null(taxon_names) || !is.character(taxon_names) ||
+        length(taxon_names) == 0 || any(is.na(taxon_names))) {
+    stop("'taxon_names' must be a non-empty character vector, cannot be NULL,
+         and must not contain NA values")
+  }
+  match_patterns <- c("cf.", "aff.", "\\?", "^[A-Za-z]*$", "indet.", "sp.")
```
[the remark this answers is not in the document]
bethany-j-allen: This structure is neat though, because adding a switch to include or not include genus level identifications could be quite easy, an if/else statement here with and without "sp." should do most of the hard work.

bethany-j-allen (commented):
Hi @BrunaFarina, I think you've made a great start here. I've added my thoughts to the points already raised, mostly focused on the philosophy as I think it will be easier to go through the exact wording and code once we have agreed on these broad-scale points. Happy to discuss anything here in more detail.

R/tax_certainty.R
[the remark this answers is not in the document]
LewisAJones: As the function is currently implemented, I think the main advantages of this would be:
- Keeping data together and more organised
- Environment "tidier", i.e. less objects
- In line with a lot of the package

Any others?

R/tax_certainty.R
```diff
@@ -0,0 +1,33 @@
+#' Check if a taxon is certain or uncertain.
```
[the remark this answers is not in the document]
LewisAJones: Agreed - that's better. Perhaps "Classify" rather than "Summarise" though? With summarise, I'd expect an output of how many are certain and how many are uncertain.

R/tax_certainty.R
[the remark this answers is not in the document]
LewisAJones: How do we think this should be implemented? I see a few potential options.
- Do we go with a switch where we identify whether genus names (or other taxonomic ranks) are present e.g. `tax_certainty(name, genus = TRUE)`? 
- Do we have a second input vector of ranks `tax_certainty(name, rank)`? 
- Do we have multiple potential input arguments (with only one allowed) and carry out the relevant checks e.g. `tax_certainty(species = NULL, genus = NULL, family = NULL)`?

R/tax_certainty.R
LewisAJones: I don't think it is confusing if you frame it in the docs as "the taxonomic identification is certain", TRUE or FALSE. That said, I like your second option. It could be envisioned as something like:
- `tax_certainty(certain = TRUE, uncertain = FALSE)` or 
- `tax_certainty(def = list(certain = 1, uncertain = 0)`

BrunaFarina:
Hi everyone, I made some changes in the code, but haven’t tested it thoroughly yet. I tried to address all the comments above. So, if (or when) the function works, it should:
  - allow the user to use a data frame as input
  - allow for NA, and they should be treated as “uncertain”
   - allow to choose taxonomic rank (and match the correct qualifiers against the taxa names - correctly place uncertainty)
- user have 3 options of how certainty should be displayed (TRUE/FALSE, 1/0, certain/uncertain)
- option to add "certainty" as a new column of df
- function should allow 1 column data frame

In this new code we check for 'cf.', 'aff.', '?', 'indet.', 'sp.', 'ex gr.', 'incertae sedis', 'problematica', 'informal' (not sure this is a taxonomic qualifier for uncertainty but I saw it in some names in pbdb), and NA (Should we check for names between quotation marks? ).
I created two vectors of qualifiers, one for species and other for the other taxonomic ranks. The only difference is that for species the uncertainty should be placed between genus name and species name (so a hypothetical case like “? genus especies”  would be treated as certain - not sure if this is possible, but it’s just an example. Whereas “genus ? especies” will be uncertain.). Species case also check for sp. (Besides all other qualifiers) - in this case in the end of the string. 
For other taxonomic ranks 'cf.’/‘aff.’/'?' should be detected on the beginning of the string (e.g., ? genus or cf. genus). I removed “indet.” from this case otherwise someone working, let’s say at family level, would have “family indet.” and then it would be treated as uncertain, even though the family is identified. 

I will work on the tests asap, but let me know if all this makes sense or if I missed something

LewisAJones:
Hi @BrunaFarina! 

First, I want to say thank you for your patience with this PR and give my apologies for the **EXTREMELY** long time it has taken me to get around to reviewing this PR again. Your last changes went a long way in improving the function and streamlining the code!

Second, in an effort to streamline getting this PR integrated, I’ve made some direct suggested changes to the documentation, function, and tests. The changes are mainly to bring it in line with the rest of the palaeoverse package (e.g. changing tax_df to taxdf). Please let me know your thoughts on the changes! If anything looks off, we can revert them.

One potential addition did come to mind when I was going through the code again - often with the PBDB we have occurrences with values such as “NO_FAMILY_SPECIFIED”, I wonder if we should screen for this kind of thing too by using `rank`/`name`? e.g.

```paste0("NO_", toupper(rank), "_SPECIFIED")```

Any thoughts?

Once you are happy with the changes, perhaps @bethany-j-allen or @willgearty can give it a once over and make sure they are happy too!

**Review by willgearty — APPROVED** [review]
Sorry for the terrible delay, this looks great! I just had a couple of comments, take them or leave them!

And don't forget to add an item to NEWS.md!

File: R/tax_certainty.R [hunk 11]
```diff
@@ -0,0 +1,114 @@
...
+  # Calculate row sums (1 = match/uncertain, 0 = no match/certain)
+  if (is.null(nrow(matches))) {
+    matches <- sum(matches)
+  } else {
+    matches <- rowSums(as.matrix.data.frame(matches))
+  }
+  # Match user definitions
+  # first element of `certainty` is certain (0 + 1), second uncertain (1 + 1)
+  matches <- matches + 1
+  classif <- certainty[matches]
```
bethany-j-allen, line 106: This method of converting the uncertainty match sums needs improving. At present, T/F values are converted to 0/1 then summed, then 1 is added to match the positions in the certainty vector. However, this means certainty is matched with a value of 1 and uncertainty with a value of 2, but when more than one uncertainty match occurs (3 or above), the returned value is NA. Maybe the most efficient way is to sum the uncertainty hits then classify based on "0" and "not 0"?

bethany-j-allen (commented):
Hello all, this is quite an extensive review but that was the intention, I went through everything except the tests (which I think will need revisiting after we've remedied these comments). The biggest thing here is that the mechanism for converting the uncertainties into a binary vector does not yet work, as it does not handle the scenario of two or more uncertainty expressions.
I'm happy to work on this more actively (i.e. respond to my own comments) if Bruna does not have time to do so.

R/tax_certainty.R
```diff
@@ -0,0 +1,114 @@
...
+#' @section Reviewer(s):
+#'     Lewis A. Jones, Bethany J. Allen, & William Gearty
+#'
+#' @examples
+#' # Get internal data
+#' data(tetrapods)
+#' # Summarise taxonomic certainty
+#' certainty <- tax_certainty(taxdf = tetrapods, name = "accepted_name",
+#'                            certainty = c("certain", "uncertain"),
+#'                            rank = "species", add_df = TRUE)
```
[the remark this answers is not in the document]
LewisAJones: Agree with all of the above, but I think (3) should ultimately go into `tax_range_strat` as it is preparing data for that function...

R/tax_certainty.R
```diff
@@ -0,0 +1,114 @@
+#' Classify the certainty of taxonomic identifications
+#'
+#' Check whether a given taxonomic name is certain or uncertain by screening
+#' for common substitutes, abbreviations, qualifiers, and notations for
+#' denoting the certainty of taxonomic identifications (see Details for
+#' screening values).
+#'
+#' @param taxdf \code{data.frame}. A dataframe with named columns
```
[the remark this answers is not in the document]
LewisAJones: True! You could in theory have a single column dataframe.

R/tax_certainty.R
```diff
@@ -0,0 +1,114 @@
...
+#'   containing taxonomic names to be checked (e.g. "species"). Names
+#'   separated by underscores or spaces are allowed.
+#' @param name \code{character}. The column name of the taxonomic names you
+#'   wish to check (e.g. "genus").
+#' @param rank \code{character}. Taxonomic rank to be considered when
+#'   screening for certainty in taxonomic assignment (e.g. "species", "genus",
+#'   "family", "order", "class").
```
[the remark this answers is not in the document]
bethany-j-allen: Maybe it's the phrase "to be considered" that I find vague. Agree that we can keep the nitty-gritty to `Details` and signpost there.
How about e.g. "Taxonomic rank at which to evaluate certainty (e.g. ...). Determines which symbols/abbreviations are considered to indicate uncertainty, with respect to the provided rank. See `Details` for more information."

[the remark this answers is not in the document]
bethany-j-allen: What you have added to the `Details` is in a good direction but I think should be comprehensive/specific, i.e. rather than "For example..." I think we should list and justify each of the differences.
On that note, I think we should revisit these lists to make sure we are completely in agreement. E.g. at the moment "indet." is not flagged as uncertainty for higher taxa, but it is almost exclusively used for higher taxa (for genera "Example sp." is more common whereas family or above tends to be "Example indet.").

R/tax_certainty.R
```diff
@@ -0,0 +1,114 @@
...
+  # Check for valid rank input
+  if (is.null(rank) ||
+      !rank %in% c("species", "genus", "family", "order", "class")) {
+    stop("Invalid `rank`. It must be either: 'species', 'genus', 'family',
+         'order', or 'class'")
+  }
+  # Common PBDB entry (NO_RANK_SPECIFIED)
+  not_specified <- paste0("NO_", toupper(rank), "_SPECIFIED")
+  # Assign rank for handling
+  rank <- ifelse(rank == "species", "species", "other")
```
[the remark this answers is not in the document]
bethany-j-allen: Ah yes, I see you incorporated my suggestion of using a wildcard on this - in which case I think it makes even more sense to follow through with a binary switch here. I think `species` with T/F is clear enough as long as we explain it properly in the documentation.
I don't feel strongly about what the default should be as I could see the function getting thorough use both ways - our example has species-level data so perhaps makes sense to make this the default, although as I suggested above re: the expanded examples, I think we should demonstrate both and compare/contrast anyway.

bethany-j-allen:
Thanks a lot @LewisAJones! I think I've replied to everything outstanding - happy to have a quick meeting if this would be the most efficient way to clear up any confusion or disagreement.

bethany-j-allen:
@willgearty I also suggested something like this solution, I agree that it could be fiddly for us to implement but that the payoff in terms of meeting both novice and expert user needs could be worth it

willgearty:
If you're worried about the number of functions, we could just have two arguments, one for a custom list of filters (perhaps with filters that make sense for any rank by default) and one for rank (which would then add a curated set of filters to the custom list depending on the rank).

LewisAJones:
I've been working on this a little bit this morning. My idea is to update this function, retaining the `rank` field which will be used to control which values are screened for, and allow users to add `custom` values to screen for.

For the default values to screen for, I've been following this [paper](https://besjournals.onlinelibrary.wiley.com/doi/10.1111/2041-210X.12594). Which has led to me putting together this list. 

```
  screen <- list(subspecies = c("ssp.", "subsp."),
                 species = c("sp.", "spp."),
                 genus = "gen.",
                 family = "fam.",
                 indeterminable = c("indeterminabilis", "indeterminata",
                                    "indet.", "ind."),
                 uncertain = c("incerta", "?", "inc."),
                 confer = c("confer$", "cf.", "cfr.", "conf."),
                 dubia = c("dubia$","sp. dub."),
                 incertae = c("incertae sedis", "inc. sed."),
                 problematic = c("problematica"),
                 informal = c("informal"),
                 unavailable = c("^NA$"),
                 trace = c("ex.", "exuvia", "exuviae"),
                 not_specified = c("NO_.*_SPECIFIED"),
                 custom = custom # User input values
                 )
```

Do you have any other suggestions for values to include? Any stand out as problematic for screening for (e.g. do we need exert greater control over any with regex)? @bethany-j-allen @willgearty

bethany-j-allen:
This looks great! So would the idea here be that a genus-level filter would just remove the subspecies and species lists?
Quote marks are commonly used in the Database for e.g. wastebasket identifications so should maybe be added to the `uncertain` list.
The punctuation is an interesting one, I could see people having just `cf` or `sp` in their own taxon lists, but removing the punctuation starts to make it easier for these to crop up in the middle of valid taxon names, so I think it's safer to say that the punctuation is required.

LewisAJones:
Hi @willgearty @bethany-j-allen, I've worked on this function a little bit this afternoon. I've taken it in a direction which provides a pre-defined comprehensive list of terms to screen and allows users to provide their own custom terms via the `terms` argument (e.g. `terms = list(custom = "species1")`). Functionality also allows users to remove or modify terms from the pre-defined list (e.g. `terms = list(species = "")` or `terms = list(genus = "genus")`). Let me know what you think!

R/tax_certainty.R
willgearty: All values within a column must be the same type. We might want to note here that one of the values will be coerced to the type of the other value if their types are different.

R/tax_certainty.R
```diff
@@ -5,14 +5,16 @@
...
+#' @param taxdf \code{data.frame}. A dataframe with a named column containing
+#'   the taxonomic names to be checked.
 #' @param name \code{character}. The column name of the taxonomic names you
-#'   wish to check (e.g. "genus"). Taxonomic names separated by underscores or
-#'   spaces are allowed.
-#' @param rank \code{character}. Taxonomic rank to be considered when
-#'   screening for certainty in taxonomic assignment (e.g. "species", "genus",
-#'   "family", "order", "class"). See details.
+#'   wish to check (e.g. "identified_name"). Taxonomic names separated by
+#'   underscores or spaces are allowed.
```
willgearty: > Taxonomic names separated by underscores or spaces are allowed.

I think this can be removed. I don't think there's any reason to think this isn't the case (and the Details is much more verbose about what conditions are checked by default).

R/tax_certainty.R
```diff
@@ -72,57 +98,54 @@ tax_certainty <- function(taxdf = NULL, name = NULL, rank = NULL,
...
-  matches <- sapply(match_patterns, grepl, taxdf$certainty, ignore.case = TRUE)
-  # Calculate row sums (1 = match/uncertain, 0 = no match/certain)
-  if (!is.matrix(matches)) {
-    matches <- sum(matches)
-  } else {
-    matches <- rowSums(as.matrix.data.frame(matches))
-  }
+  matches <- lapply(screen, function(x) {
+    sapply(x, grepl, taxdf$certainty, ignore.case = TRUE)
+  })
```
willgearty, line 162: Would `modifyList()` be simpler here?

R/tax_certainty.R
```diff
@@ -30,21 +32,45 @@
...
+#'         uncertain = c("incerta", "?", "inc\\.",
+#'                       "\\\"\\\"", "\\\'\\\'"),
+#'         confer = c("confer$", "cf\\.", "cfr\\.", "conf\\."),
+#'         dubia = c("dubia$","sp\\. dub\\.", "nomen dubium"),
+#'         incertae = c("incertae sedis", "inc\\. sed\\."),
+#'         problematic = c("problematica"),
+#'         informal = c("informal"),
+#'         unavailable = c("^NA$"),
+#'         trace = c("ex\\.", "exuvia", "exuviae"),
+#'         not_specified = c("NO_.*_SPECIFIED"))
```
willgearty: Would it be better to have this be a bulleted list of natural language instead of raw code? Some people might not understand the regexes?

tests/testthat/test-tax_certainty.R
```diff
@@ -4,42 +4,43 @@ test_that("tax_certainty() works", {
...
-                                                      rank = "species",
                                                       append = TRUE)))
-  ## Genus level working?
+  ## Custom terms
   expect_true(is.vector(tax_certainty(taxdf = tetrapods,
                                       name = "genus",
-                                      rank = "genus",
+                                      terms = list(custom = "test"),
                                       append = FALSE)))
 })
```
willgearty: Given that the built-in data is static, would it make sense to have a test that checks the correct values (not just the correct type) are returned for a specific `tax_certainty()` call?

willgearty (commented):
This looks great @LewisAJones, thanks for finding time to put this together! Most of my comments are more for discussion than required changes, so just let me know what you think. I also really like the inclusion of a reference; it makes it harder for folks to complain about individual filters and also makes it seem a lot less arbitrary.

R/tax_certainty.R
LewisAJones: Yes... this is something I was on the fence about... I think a bulleted list would be better for reading, but the raw code allows users to see exactly the structure of the list we're using in our screening (e.g. names of list elements to modify). Perhaps a bulleted list in natural language and then what that translates to in code in brackets?

tests/testthat/test-tax_certainty.R
LewisAJones: We could do that!

R/tax_certainty.R
LewisAJones: Just spotted that `\code{character}` should probably also be removed

R/tax_certainty.R
LewisAJones: I did prefer `modifyList()` but it doesn't play nicely with `terms = NULL`. So, I've opted not to change this rather than adding another line of code or changing the default (`terms = list()`).

R/tax_certainty.R
willgearty: Oh, I just realized that since this argument must be a vector (rather than a list), the two values must be of the same type anyways (or will be coerced in the process of putting them in a vector together, e.g., `c(1, "2")`).

R/tax_certainty.R
willgearty: Looks like you you've gone with just the bulleted list with natural language for now, but I think putting the regexes in brackets would also be fine.

LewisAJones: I did experiment with this... but it ended up looking... too overwhelming

R/tax_certainty.R
LewisAJones: So probably fine without additional text here? We don't want argument descriptions to be tooooooo long

willgearty: Yeah, probably fine as-is

LewisAJones:
Hi @bethany-j-allen, do you think you'll be able to get around to reviewing this any time soon? Just wondering whether to stage this for the next version of the package, or not!

bethany-j-allen:
Sorry for the delay! I think this structure is really smart, it's very flexible but at a surface level is still very easy to use.

I just have a few suggestions for making extra clear exactly what is going on:
- When `append = FALSE`, I wonder whether it would be better to also include the `name` column (i.e. two-column table). This would certainly help with checking for people who want to investigate exactly what is being filtered out.
- I think an extra sentence on potential for false fits would be good - i.e. we specifically include the "." on "sp." to reduce the risk of false matches mid-name, but false matches remain possible, and it's up to the user to know and understand the scale of this risk (especially when modifying it themselves).
- From the way that the `terms` list is presented I did not immediately understand that the name on the left is given to the list of terms on the right (i.e. it's a nested list). Maybe e.g. a table structure in the documentation would make this clearer? Otherwise this could just be explained with an extra sentence.
- In part following on from the above, I did not immediately understand that `terms = list(species = "")` would void this specific named list. Again, perhaps just needs explaining a little more clearly. Specifically, we previously discussed how it needs to be easy and intuitive for users to remove the `species` and `subspecies` terms in order to conduct an appropriate genus-level sift. While these are provided (very neatly) within the examples, it is not clearly explained that this is what is being done, and I think a novice user could misunderstand that we are pointing them towards how to do this specific and very relevant thing.
- I noticed that the question mark is missing from the `uncertain` list in the documentation, it's worth double checking that the internal and external fully correspond.
- I see that when the matches are done, this uses `ignore.case = TRUE`, so perhaps it should also be added in the documentation that the case does not matter.

Otherwise this looks great! And incredibly useful. I also really like pointing towards the external reference for the list of uncertainty terms, definitely a good idea.

LewisAJones:
Thanks for the review @bethany-j-allen and some great comments! If you are happy with these changes, I think we are ready to get this merged (after over 100 comments and 18 months 😆)! 

> When append = FALSE, I wonder whether it would be better to also include the name column (i.e. two-column table). This would certainly help with checking for people who want to investigate exactly what is being filtered out.

As `append = TRUE` already binds the output to the input dataframe, I was wondering whether a named vector would be better here. However, I ultimately ended up going with your suggestion as I think the expected input/output we will be wanting for the refactor are dataframes.

> I think an extra sentence on potential for false fits would be good - i.e. we specifically include the "." on "sp." to reduce the risk of false matches mid-name, but false matches remain possible, and it's up to the user to know and understand the scale of this risk (especially when modifying it themselves).

Good point, implemented!

> From the way that the terms list is presented I did not immediately understand that the name on the left is given to the list of terms on the right (i.e. it's a nested list). Maybe e.g. a table structure in the documentation would make this clearer? Otherwise this could just be explained with an extra sentence.

I've stuck with referring to this as a named list for now but given examples of what this actually looks like several times in the documentation. My fear with using "nested list" here is that people would use:

```
list(species = list("sp", ...))
```

Instead of:

```
list(species = c("sp", ...))
```

> In part following on from the above, I did not immediately understand that terms = list(species = "") would void this specific named list. Again, perhaps just needs explaining a little more clearly. Specifically, we previously discussed how it needs to be easy and intuitive for users to remove the species and subspecies terms in order to conduct an appropriate genus-level sift. While these are provided (very neatly) within the examples, it is not clearly explained that this is what is being done, and I think a novice user could misunderstand that we are pointing them towards how to do this specific and very relevant thing.

I've now switched from using:

```
terms = list(species = "")
```

to:

```
terms = list(species = NULL)
```

for entirely voiding specific named lists. I think this is a little more intuitive. I have also added more documentation and given examples of how this should be used.

> I noticed that the question mark is missing from the uncertain list in the documentation, it's worth double checking that the internal and external fully correspond.

Added the question mark, double-checked, and resolved one bug related to this (same term used twice in screening).

> I see that when the matches are done, this uses ignore.case = TRUE, so perhaps it should also be added in the documentation that the case does not matter.

Added to the documentation to make this clearer.


Note, @etiennebacher, I've had to add gnu.org to the `check-links.yaml` here, but let me know if there is a reason why this shouldn't be here (looks like it should have been added before and we missed it?).

bethany-j-allen:
This looks great, except...
I added an initial example where certainty is not appended and the two-column data frame is produced. Here you can see that because we have `gen.` and `sp.` flagged as uncertain, this is also catching all the type locality instances, i.e. `n. gen.` and `n. sp.`. This is a PBDB-specific issue but one I think we should probably remedy. What's the best way to resolve this, add `n. gen.`, `n. ssp.` and `n. sp.` as exceptions?

LewisAJones:
Thanks for adding an additional example @bethany-j-allen!

Hmmm, I am not super thrilled about adding PBDB-specific issues and exceptions... but this is clearly an issue. 

Would adding a note to the documentation suffice here (we already highlight the potential for false positives so we could expand on this)? We would probably have to rewrite quite a bit of the current code to support exceptions if we went down this route, we would also have to allow the users to have control over this.

bethany-j-allen:
I don't love it either but I think the function should work out-of-the-box with the most common data that anyone will put into it... I'm struggling to think of other ways to make this work.

LewisAJones:
We would also need to be careful with exceptions in that they don't completely overwrite other flags 🤔 

I've implemented some changes that uses a more explicit (complicated) regex that explicitly ignores `n. gen.` but still flags `gen.`, etc. If you could take it for a spin @bethany-j-allen and check the docs and let me know if you're happy with that!

bethany-j-allen:
This looks good to me! Could consider changing some of the examples given in the documentation to options with less punctuation :) but otherwise seems ready to go. Thanks for pushing this over the finish line @LewisAJones!

bethany-j-allen (approved):
Tweaked the last example but otherwise good to go!
One *tiny* thing is that we don't check the placement of `?` or `cf.` when using the genus-level filter. But we state clearly that we match based on presence in the string, which is exactly what is being done. Maybe enhancing this to consider placement within the string could be a nice small-ish issue for the future?

LewisAJones (approved):

PR #134, opened 2025-04-02
Add text autofitting and title options to `axis_geo()`

### LINKED ISSUE #21: Automated label scaling for axis_geo()
In `deeptime::coord_geo()`, there is the option to use `ggfittext` to scale the labels such that they don't overlap with one another and fit within their boxes. I'm not aware of something similar for base R, but we could try to emulate it for `axis_geo()`. @KEichenseer looked into it during development but couldn't figure out a solution, so it might not even be possible in base R?

### LINKED ISSUE #133: Automatic replacement of x-axis labels and title
Currently, `axis_geo()` adds a geological timescale axis without replacing the existing labels or title on the x-axis. 

```
# Load libraries
library(palaeoverse)
# Example plot
plot(0:100, xlim = c(100, 0), ylim = c(100, 0), 
     xlab = "Time (Ma)",
     ylab = "Some variable")
# Add geo axis
axis_geo(side = 1, intervals = "period")
```

Image

This results in users having to prevent `plot()` from automatically plotting any initial labels or title, and manually experimenting with the position of the x-axis title (e.g. via `title()`) to get the right placement. 

I think it would be a good addition if `axis_geo` added the x-axis title in a suitable position, and removed the former labels/title. The former should be doable (?) but the latter a little more tricky.... the only solution I can think of is essentially creating a white layer on top of the previous title/labels.

willgearty:
This PR does the following:

- Adds an option to autofit interval labels to their respective boxes like `deeptime` does with `ggfittext`. I've added documentation and tests for this functionality.
  - Interval labels with a size smaller than 0.1 are skipped
- Filter plotted intervals to just those within the visible limits (this has few, if any, visible changes, but it removes unnecessary plotted elements that were invisible outside of inspecting the plot as an SVG)
- Adds an option to add a title (using graphics::title) that inherits all title settings but is pushed away from the plot based on the thickness of the axis_geo scale

Fixes #21 and Fixes #133.


willgearty:
TODO: add type checking and tests for the title argument

willgearty:
Comments addressed and NEWS.md updated!

PR #138, opened 2025-05-16
New function: tax_rel_abun

### LINKED ISSUE #86: Summarise abundance in occurrence datasets
Understanding the abundance distribution within a set of occurrences is a fundamental way of exploring a dataset, but we currently don't have any functions that do this.

It would be great to have a function that summarises the abundance distribution in an occurrence dataset and outputs:
- a table (of taxa and their (relative) abundances)
- a graph (of the shape of the ranked abundance distribution)
- a string (vector of abundances suitable for input into other analysis functions, e.g. iNEXT, vegan)

iNEXT has functions to create strings but it would be nice to have a more flexible function specifically designed to do this from a wider range of input formats.

**Additional context**
It would be great to expand this to model fitting on the distribution, but this is straying into analysis rather than data exploration.

mcnikolic:
# Function name
tax_rel_abun

## Description

Function to calculate absolute and relative abundance of taxa in an occurrene dataset and plot the rank
  abundance distribution

## Motivation and Context

Fixes #86


LewisAJones (changes requested):
Hi @mcnikolic, thanks a lot for submitting this PR and so soon after joining the team 😁! This is a really good first draft given how picky we can be about code and documentation style. I've had a first look over the function without paying too much attention to the documentation at this stage. I have a few comments which I think should be relatively easy to address but give us a shout if you need help with anything or clarification. Once we're happy with where the function is at, it would be good to also add some tests (we can discuss when we get to that stage!).

Thanks again @mcnikolic, great job! 😁

R/tax_rel_abun.R
```diff
@@ -0,0 +1,118 @@
+#' Summarise abundance of fossil occurrences
+#'
+#' A function to calculate absolute/relative abundance and rank abundance distribution of taxa in a fossil occurrence dataset
```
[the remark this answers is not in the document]
willgearty: More broadly, we tend to follow the tidyverse style guidelines. You can check if you've followed these be either checking the details of the "lint" check when you make a PR or install the lintr package and use the lintr addins in RStudio.

R/tax_rel_abun.R
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+
+  # Tabulate absolute abundance of occurrences for each taxon
+  occ_table <- table(occdf[[rank]])
+  occ_table <- data.frame(occ_table)
+  colnames(occ_table) <- c('Taxon', 'Abundance')
+
+  # Calculate relative abundance for raw and log data
+  rel_abun <- occ_table$Abundance/max(occ_table$Abundance)
```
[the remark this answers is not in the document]
willgearty: I had been wondering the same thing but had assumed I just didn't know because I don't use abundance data. But I think @LewisAJones is correct here: https://en.wikipedia.org/wiki/Abundance_(ecology)#:~:text=Relative%20species%20abundance%20is%20calculated,of%20species%20from%20all%20groups.

LewisAJones:
Hi @mcnikolic, do you think you'll have chance to implement these changes soon? Should we aim to get this function into the next version?

mcnikolic:
> Hi @mcnikolic, do you think you'll have chance to implement these changes soon? Should we aim to get this function into the next version?

Yes, I think so! Hopefully by the end of the coming week

R/tax_rel_abun.R
```diff
@@ -0,0 +1,118 @@
...
+  rownames(occ_table) <- c(1:nrow(occ_table))
+  occ_table$Abundance_rank <- c(1:1:nrow(occ_table))
+
+
+  # Plot rank abundance distribution
+  if (plot == TRUE) {
+    # Default plot args
+    args <- list(main = "Rank abundance",
+                 xlab = "Abundance rank",
+                 ylab = "Relative abundance",
```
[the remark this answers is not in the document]
willgearty: It might be easier to plot it as relative abundance, specify `ylog = FALSE`, and then allow the user to override that with `plot_args`.

willgearty (changes requested):
Thanks for addressing our previous comments @mcnikolic! It's looking much cleaner. I think the functionality is basically there, but I do have several comments about the code (including some bugs I found). Let me know if you have any questions.

bethany-j-allen (commented):
I think this is a great start @mcnikolic! I have added some suggestions on how the function and documentation could be developed a bit, happy to help more actively with coding/writing if needed.

R/tax_abund.R
```diff
@@ -0,0 +1,160 @@
...
+                 col = "black",
+                 brks = 50,
+                 log = FALSE,
+                 abs = FALSE)
+
+    # Update any provided
+    rpl <- match(names(plot_args), names(args))
+    if (length(rpl) != 0) {
+      args[rpl] <- plot_args
+    }
```
willgearty: I think we can just use `utils::modifyList()` here?

willgearty (changes requested):
Thanks @mcnikolic for addressing our previous comments! I think this new version is nearly there! Just a few comments from me. @bethany-j-allen and @LewisAJones might have additional thoughts.

pull_request_description.md
willgearty: I think this file should be removed?

willgearty:
1. I fixed the failing test on the `main` branch, so if you merge that (or rebase), the R-CMD-check and test-coverage checks should be fixed.
2. You also still need to add `tax_abund` to `_pkgdown.yml` in a fitting category.
3. Finally, we still need tests for this new function.

Let me know if you have any questions about any of the above.

mcnikolic:
Ok everything is working and passing every test and check on my machine but is not passing the checks here, even when I 'fix' them, for reasons I do not understand...

R/tax_abund.R
```diff
@@ -0,0 +1,162 @@
+#' Summarise abundance of fossil occurrences
+#'
+#' A function to calculate the absolute and relative abundance of fossil
+#' occurrences in a dataset and optionally plot the rank abundance distribution
```
LewisAJones, line 4:
```suggestion
#' taxa in a dataset and optionally plot their rank abundance distribution.
```

R/tax_abund.R
```diff
@@ -0,0 +1,162 @@
+#' Summarise abundance of fossil occurrences
```
LewisAJones, line 1:
```suggestion
#' Summarise abundance of fossil taxa
```
I think this more accurately represents what the function does and is more aligned with other function titles in palaeoverse (e.g. `tax_range_time`).

R/tax_abund.R
```diff
@@ -0,0 +1,162 @@
...
+#'
+#' A function to calculate the absolute and relative abundance of fossil
+#' occurrences in a dataset and optionally plot the rank abundance distribution
+#' of taxa
+#'
+#' @param occdf \code{data.frame}. A dataframe of fossil occurrences.
```
LewisAJones, line 7: To bring this in line with our other function documentation, I suggest something along the lines of "A dataframe of fossil occurrences containing at least one column: names of taxa (and an optional column containing abundances)"... this then allows you to build on this for the other argument descriptions as the user already has an idea of what is needed before getting to the details.

R/tax_abund.R
```diff
@@ -0,0 +1,162 @@
...
+#' @param name \code{character}. A column name with the taxonomic rank at which
```
LewisAJones, line 8: Again, to bring this in line with our other function documentation, I would change this argument definition to something along the lines of "The name of the column you wish to be treated as the input names for abundance calculation, e.g. "genus" (default)." I also think that the current use of "taxonomic rank at which you want to calculate abundance" may be misinterpreted.

R/tax_abund.R
```diff
@@ -0,0 +1,162 @@
...
+#'   you want to calculate abundance (e.g. "genus").
+#' @param abund_vals \code{character}. An optional column name of abundance
```
LewisAJones, line 10: As above, perhaps something more along the lines of:
 
"The name of the column you wish to be treated as the input abundance values for the taxa (`name`). This argument is optional, but if it is not provided each value within `name` is assumed to represent an abundance value of 1."

I am sure you can phrase it better than I have!

R/tax_abund.R
```diff
@@ -0,0 +1,162 @@
...
+#'   values for a given taxon or collection. If this column is provided it
+#'   will be used to calculate the abundance values. Otherwise, row wise
+#'   occurrence counts will be used.
+#' @param plot \code{logical}. Should a rank abundance distribution plot be
+#'   generated in addition to the dataframe of results?
```
LewisAJones, line 15:
```suggestion
#'   generated?
```

R/tax_abund.R
```diff
@@ -0,0 +1,162 @@
...
+#' @param plot_args \code{list}. A list of optional arguments relevant to
+#'   plotting. See Details for options.
+#'
+#' @return A \code{data.frame} containing the following columns:
+#'   `Taxon` = taxon ID (from `name` column)
+#'   `Abundance` = absolute abundance of each fossil taxon
+#'   `Rel_abundance` = relative abundance of each fossil taxon
+#'   `Log_rel_abundance` = natural logarithm of relative abundance
+#'                           of each fossil taxon
+#'   `Abundance_rank` = abundance rank of each taxon within dataset
```
LewisAJones, line 25: Not really a major issue, but I wonder if the rank should come before all the different types of abundance values

R/tax_abund.R
```diff
@@ -0,0 +1,162 @@
...
+#'
+#' @details The (absolute and relative) abundance of each taxon is calculated
```
LewisAJones, line 27: I think this text could be streamlined, and I would probably avoid mention of collections. Essentially, the input can be one or two columns, if the input is one column, we treat each instance (row) as an individual occurrence to sum for abundances, if we have a two column input we already have a sum of abundances for that occurrence?

R/tax_abund.R
```diff
@@ -0,0 +1,162 @@
...
+#'   either from a given column of abundance values per taxon or collection or by
+#'   tabulating all rowwise counts of unique taxon occurrences (`name` column)
+#'   from the input `occdf`. Absolute abundance is calculated by counting the
+#'   number of occurrences, while relative abundance is calculated relative to
+#'   the total community abundance in the dataset. A plot of the rank abundance
+#'   distribution is also returned if `plot = TRUE`. Customisable argument
+#'   options to pass to `plot_args` as a list (and their defaults)
+#'   for plotting include:
```
LewisAJones, line 35: Note, this may need to change if you decide to implement my suggestions below

R/tax_abund.R
```diff
@@ -0,0 +1,162 @@
...
+#'   - main = "Rank abundance"
+#'   - xlab = "Abundance rank"
+#'   - ylab = "Relative abundance"
+#'   - col = "black"
+#'   - brks = 50
+#'   - log = FALSE
+#'   - abs = FALSE
```
LewisAJones, line 42: Should absolute be the default (curious of opinions on this)? I guess there is an argument to be made here that by default we are transforming data. However, relative is probably more often visualised. 

If you go with absolute being the default, this argument should probably change to `rel` for relative?

R/tax_abund.R
```diff
@@ -0,0 +1,162 @@
...
+#'
+#'   The `brks` argument will set the x axis label breaks. Use `log = TRUE` to
+#'   plot the logged rank abundance, and use `abs = TRUE` to plot the
+#'   absolute abundance data.
+#'
+#'   Caution: taxonomic synonyms will cause abundances for the accepted taxon
```
LewisAJones, line 48: Should "synonyms" be changed to "homonyms" here?

R/tax_abund.R
```diff
@@ -0,0 +1,162 @@
...
+#'   to be distributed across the synonymised taxa, causing the accepted taxon
+#'   to be under-counted. Use [tax_check()] to check for and clean synonyms.
+#'
```
LewisAJones, line 51: Don't forget to add a developer and reviewer section (see other function documentation for examples).

LewisAJones, line 51: If you continue to use barplot (see my comment below about this), I think you also need to add: 

```suggestion
#' @importFrom graphics barplot
```

You might also need to add:

```suggestion
#' @importFrom utils modifyList
```

R/tax_abund.R
```diff
@@ -0,0 +1,162 @@
...
+#' @export
+tax_abund <- function(occdf,
+                      name = "genus",
+                      abund_vals = NULL,
+                      plot = FALSE,
+                      plot_args = NULL) {
+
+  #=== Handling errors ===
+  if (is.data.frame(occdf) == FALSE) {
```
LewisAJones, line 73: There is nothing wrong this, but we are currently using a mix of `!x` (not operators) and `x == FALSE` conditional statements. We could standardise to use just one or the other.

R/tax_abund.R
```diff
@@ -0,0 +1,162 @@
...
+
+  if (name %in% colnames(occdf) == FALSE) {
+    stop(paste0("`occdf` must contain `", name, "` column"))
+  }
+
+  if (!is.null(abund_vals) && abund_vals %in% colnames(occdf) == FALSE) {
+    stop(paste0("`abund_vals` must either be NULL or a column in `occdf`"))
+  }
+
+  if (length(which(is.na(occdf[[name]]))) > 0) {
```
LewisAJones, line 93: Could this be simplified to?:
```suggestion
  if (any(is.na(occdf[[name]])) > 0) {
```

R/tax_abund.R
```diff
@@ -0,0 +1,162 @@
...
+    stop(paste0(name, " column contains NAs"))
+  }
+
+  # Tabulate absolute abundance of (row wise) occurrences for each taxon
+  if (is.null(abund_vals)) {
```
LewisAJones, line 98: To streamline code, I guess we could also just add an abundance value column if one is not provided, with all values equal to 1?

R/tax_abund.R
```diff
@@ -0,0 +1,162 @@
...
+    occ_table <- table(occdf[[name]])
+    occ_table <- data.frame(occ_table)
+    colnames(occ_table) <- c("taxon", "abundance")
+    occ_table$abundance <- as.numeric(occ_table$abundance)
```
LewisAJones, line 102: Is this needed?

R/tax_abund.R
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  if (length(which(is.na(occdf[[name]]))) > 0) {
+    stop(paste0(name, " column contains NAs"))
+  }
+
+  # Tabulate absolute abundance of (row wise) occurrences for each taxon
+  if (is.null(abund_vals)) {
+    occ_table <- table(occdf[[name]])
+    occ_table <- data.frame(occ_table)
```
LewisAJones, line 100: This could be merged with the former line 
```
data.frame(table(...))
```

R/tax_abund.R
```diff
@@ -0,0 +1,162 @@
...
+    colnames(occ_table) <- c("taxon", "abundance")
+    occ_table$abundance <- as.numeric(occ_table$abundance)
+  } else {
+    # Calculate abundances from abundance column if provided
+    occdf[[abund_vals]][is.na(occdf[[abund_vals]])] <- 1
```
LewisAJones, line 105: I don't think we should do this... we're assuming any NA values represent an abundance of 1, which is probably true for PBDB, but not for other datasets. I would suggest having the function error if an abundance value column is provided but it contains NAs.

R/tax_abund.R
```diff
@@ -0,0 +1,162 @@
...
+    occ_table <- group_apply(occdf = occdf, group = name,
+                             fun = function(x) sum(x[[abund_vals]]))
+    colnames(occ_table) <- c("abundance", "taxon")
+    occ_table <- occ_table[,c(2,1)]
```
LewisAJones, line 109: Correcting for code style

```suggestion
    occ_table <- occ_table[ ,c(2, 1)]
```

I generally try to be explicit when rearranging columns too by using names rather than index. This way, if there are some changes elsewhere that cause column order changes, it's less likely to cause an issue.

R/tax_abund.R
```diff
@@ -0,0 +1,162 @@
...
+    occ_table$abundance <- as.numeric(occ_table$abundance)
```
LewisAJones, line 110: Is this needed?

R/tax_abund.R
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  # Calculate relative abundance for raw and log data
+  occ_table$rel_abundance <- occ_table$abundance /
```
LewisAJones, line 114: Should these be converted to percentages? I don't think it matters too much, but it's quite a common practice in my circle (coral reef ecology)

R/tax_abund.R
```diff
@@ -0,0 +1,162 @@
...
+                             sum(occ_table$abundance)
+  occ_table$log_rel_abundance <- log(occ_table$abundance) /
+                                 sum(log(occ_table$abundance))
```
LewisAJones, line 117: I *think* you would want to sum abundances before logging them?

That said, I think you can also just log transform `occ_table$rel_abundance`?

I also wonder if this is a column users would generally want available by default for relative abundances. Perhaps it is more for visualisation purposes?

R/tax_abund.R
```diff
@@ -0,0 +1,162 @@
...
+
+  # Order by rank abundance
+  occ_table <- occ_table[order(-occ_table$rel_abundance), ]
+  rownames(occ_table) <- seq_len(nrow(occ_table))
```
LewisAJones, line 121: I believe you could simplify to:
```suggestion
  rownames(occ_table) <- NULL
```

R/tax_abund.R
```diff
@@ -0,0 +1,162 @@
...
+                 xlab = "Abundance rank",
+                 ylab = "Relative abundance",
+                 col = "black",
+                 brks = 50,
+                 log = FALSE,
+                 abs = FALSE)
+
+    # Update any provided
+    if (!is.null(plot_args)) {
+      args <- utils::modifyList(args, plot_args)
```
LewisAJones, line 139: This should be changed, and imported (see comment above)

```suggestion
      args <- modifyList(args, plot_args)
```

R/tax_abund.R
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    if (args$log == TRUE) {
```
LewisAJones, line 142: This should probably be moved to before `modifyList` otherwise we do not allow for user input.

R/tax_abund.R
```diff
@@ -0,0 +1,162 @@
...
+      args$y <- occ_table$log_rel_abundance
+      args$ylab <- "Log relative abundance"
+    }
+
+    if (args$abs == TRUE) {
```
LewisAJones, line 147: This should probably be moved to before `modifyList` otherwise we do not allow for user input.

R/tax_abund.R
LewisAJones, line 142: I wonder if log should just be something available for visualisation purposes, and users can choose whether to log absolute or relative abundances through combination of `log` and `abs`?

R/tax_abund.R
```diff
@@ -0,0 +1,162 @@
...
+      args$y <- occ_table$abundance
+      args$ylab <- "Absolute abundance"
+    }
+
+    barplot(height = args$y, names.arg = occ_table$abundance_rank,
```
LewisAJones, line 152: I'll leave this with you, but because you are already generating data for the rank order and abundances, you could actually just use the `plot` function. The advantage of this is that you wouldn't needed to import the barplot function, you could allow users to use different plot types (e.g. "h", "o", etc), and you could probably get rid of the next few lines of code from the axis breaks (plot will give you something pretty decent by default).

LewisAJones (changes requested):
Hi @mcnikolic, sorry for the delay on this! Thanks for updating the function based on previous comments. I've gone through and given the function a pretty extensive review. Sorry for all the comments but hopefully they're useful suggestions!

Let me know if you need help with anything or if any of my comments are unclear.

I'll have a thorough review of the documentation and tests once all comments have been addressed as some of this will likely change.

LewisAJones:
Hi @mcnikolic, do you think you'll be able to get around to addressing any of these changes soon? Just wondering whether to stage this for the next version of the package, or not!

mcnikolic:
Hey Lewis, getting to it now!

R/tax_abund.R
mcnikolic: In the tetrapods dataset, there are instances of the same taxon in multiple rows but with different values in the abundance value column. I take it these are from different collection events. So even if an abundance value column is supplied, we still need to sum over all the abundance values for that taxon

R/tax_abund.R
mcnikolic: Yes, good catch

R/tax_abund.R
mcnikolic: Hmm, yeah, I don't feel strongly either way. I had written it to log by default as that is the way it is usually visualised, but this does deviate from the rest of Palaeoverse presenting data as-is/not transforming data or doing analyses. I'm happy to put absolute as the default and give the log option

R/tax_abund.R
mcnikolic: Agreed!

R/tax_abund.R
mcnikolic: Oh sorry, I meant to say relative not log here

PR #142, opened 2025-07-20
add filter_time function

jf15558:
# Function name
filter_time

## Description
A function to trim fossil occurrences to within a specified temporal duration and/or level of precision, using either stratigraphic interval names or absolute numeric values. There is an additional convenience feature, which I have had recent cause to use, to permit some occurrences which fall above a specified level of temporal precision to be retained for the sake of taxonomic completeness of the occurrence dataset.

Current commit contains commented code, documentation and examples for code review. testhat to be added following any changes from code review.

## Motivation and Context

Trimming and filtering by stratigraphic duration is a common preliminary phase of many palaeobiological work flows. While such a procedure can be readily implemented using logical subsetting in base R, there are multiple ways of achieving this, hence why a single standardised function will conceivably be of use.

- [  ] If it is a core feature, I have added thorough tests.


willgearty (changes requested):
Sorry for the delay on this review @jf15558, but this is a really great first draft! I will definitely be using this function a lot in my own research scripts!

I have a batch of comments, but it's definitely very close already. Don't forget you'll still need the following before it can be merged:

- generated docs (`devtools::document()`) (I wonder if we could automate this in the future...)
- add filter_time to an appropriate category in _pkgdown.yml
- add a `NEWS.md` item

R/filter_occs_time.R
```diff
@@ -0,0 +1,277 @@
+#' filter_occs_time
```
willgearty, line 1: Sorry, I wasn't saying you needed to change the name of the function (in fact, I think the old function name worked quite well and matches our existing naming scheme), just that this first line should literally be a title for the function (see suggestion)

```suggestion
#' Filter fossil occurrences by time
```

R/filter_occs_time.R
```diff
@@ -0,0 +1,277 @@
...
+#' to trim to. Only one interval name should be supplied, which should be
+#' present in column `interval_name` in `bins`.
+#' @param min_age `numeric`. The minimum absolute occurrence age to trim to.
+#' Ignored if `youngest_bin` is supplied.
+#' @param max_age `numeric`. The maximum absolute occurrence age to trim to.
+#' Ignored if `oldest_bin` is supplied.
+#' @param uncertainty A numeric value denoting the maximum tolerable temporal
+#' uncertainty for retaining occurrences. Alternatively, if this argument is set
+#' to "bin", then only occurrences which span no more than `nbin` intervals in
+#' `bins` will be retained.
```
willgearty, line 37: I definitely did not expect you to actually implement the `nbin` option, so maybe now this should be "bins" like before 🙈 (I had suggested "bin" because it was originally only one bin)? It probably doesn't matter (and you could also say "bins" but also catch "bin" as well), so I'll leave the decision up to you (and the other reviewer).

R/filter_occs_time.R
```diff
@@ -0,0 +1,277 @@
...
+#' the filtering parameters used, it is possible that all or no occurrences
+#' will be discarded, resulting in an empty `data.frame` or an unchanged data
+#' set respectively.
+#'
+#' @details
+#' When trimming by stratigraphic interval names, the boundary ages of
+#' `youngest_bin` and `oldest_bin` are taken from their entries in `bins`.
+#' Occurrences with maximum ages younger than or equal to the upper boundary
+#' age of `youngest_bin` or minimum ages older than or equal to the lower
+#' boundary age of `oldest_bin` will be discarded.
```
willgearty, line 63: Unless I'm mistaken, I think this better matches what the code is doing here?
https://github.com/jf15558/palaeoverse/blob/b8df3dd76700c9f0532ae5c84c4c45dfe22145b3/R/filter_occs_time.R#L215-L216

```suggestion
#' Occurrences with maximum ages younger than the upper boundary
#' age of `youngest_bin` or minimum ages older than the lower
#' boundary age of `oldest_bin` will be discarded.
```

willgearty (changes requested):
Thanks for humoring my suggestions, @jf15558! Just a couple more comments. We'll also probably want a second review before merging, but I don't think I'll have any other comments after this.

willgearty (approved):
Thanks again for addressing all of my little requests @jf15558! Looks like there might still be some edge cases that we could catch in tests ([see here](https://app.codecov.io/github/palaeoverse/palaeoverse/pull/142/blob/R/filter_time.R)), but it's not super necessary. Otherwise looks good to me!

Maybe @LewisAJones or @bethany-j-allen could contribute the second review?

R/filter_time.R
```diff
@@ -0,0 +1,340 @@
+#' Filter fossil occurrences by time
+#'
+#' @description
```
LewisAJones, line 3: Currently, we don't actually include this roxygen tag in other function documentation (it's not strictly necessarily). However, perhaps we should start doing so... 

Any thoughts @willgearty @etiennebacher?

R/filter_time.R
```diff
@@ -0,0 +1,340 @@
...
+#' @param overlap `logical`. Occurrences which partially overlap with the set
+#' filtering duration are discarded by default, so that the returned data falls
+#' strictly within these bounds. Set to `TRUE` to retain overlaps instead.
+#' @param uncertainty A numeric value denoting the maximum tolerable temporal
+#' uncertainty for retaining occurrences. Alternatively, if this argument is
+#' set to "bins", then only occurrences which span no more than `n_bins`
+#' intervals in `bins` will be retained.
+#' @param n_bins `numeric` The maximum tolerable bin-wise uncertainty above
+#' which occurrences will be discarded if `uncertainty = "bins"`.
+#' @param retain.taxa `logical`. Complete loss of some taxa may occur if all
```
LewisAJones, line 41: Hmmm do we want to encourage this? I think it should be expected that we lose some data when filtering. With this argument, wouldn't we end up essentially keeping occurrences we are more uncertain about, while removing many occurrences which are more constrained? What does this mean for downstream analyses? 

Also, is there a strong reason to use "retain.taxa" over "retain_taxa"?

R/filter_time.R
```diff
@@ -0,0 +1,340 @@
+#' Filter fossil occurrences by time
+#'
+#' @description
+#' A function to trim fossil occurrences to within specified stratigraphic
+#' bounds and/or degree of precision. Filtering can be performed on
+#' occurrences with point or uncertainty-bounded ages using stratigraphic
+#' interval names or absolute numeric ages.
+#'
+#' @param occdf `data.frame`. The fossil occurrences you wish to filter. This
```
LewisAJones, line 9: Obviously, we're going to be doing a big overhaul of the package docs including parameter definitions. However, I'd suggest reviewing other functions in the package and bringing this more in line with the wording elsewhere, e.g.
```suggestion
#' @param occdf `data.frame`. A dataframe of the fossil occurrences you wish to filter. This
```

LewisAJones, line 9: This doesn't really matter, but just aiming for consistency with other function docs
```suggestion
#' @param occdf \code{data.frame}. The fossil occurrences you wish to filter. This
```

R/filter_time.R
```diff
@@ -0,0 +1,340 @@
...
+#' must contain minimum and maximum occurrence age columns, or a column of
+#' point ages. All must be numeric and cannot contain missing values.
+#' @param bins This may be left as `NULL` if filtering by numeric values only.
```
LewisAJones, line 12:
```suggestion
#' @param bins \code{data.frame}. This may be left as `NULL` if filtering by numeric values only.
```

I think it would also be best to flip this parameter definition. For example: 

A dataframe of time bins used to look up numeric ages for filtering fossil occurrences, such as that returned by [time_bins()]. This must contain at least three named columns: `max_ma`, `min_ma`, and `interval_name`. No column can contain missing values and columns `max_ma` and `min_ma` must be numeric. If `NULL` (default), filtering is by numeric values only.

R/filter_time.R
```diff
@@ -0,0 +1,340 @@
...
+#' Otherwise, a `data.frame` of time bins used to look up numeric ages for
+#' filtering fossil occurrences, such as that returned by [time_bins()]. This
+#' must contain at least three columns as named by `max_ma`, `min_ma` and
+#' `interval_name`. No column can contain missing values and columns `max_ma`
+#' and `min_ma` must additionally be numeric.
+#' @param min_ma `character`. The name of the column you wish to be treated as
+#' minimum ages in `occdf` and `bins`. `min_ma` in `occdf` can instead be
```
LewisAJones, line 19: In the `bins` parameter documentation you state that the column must be named as "max_ma" and "min_ma", but here you suggest that it can be specified?

R/filter_time.R
```diff
@@ -0,0 +1,340 @@
...
+#' treated as point ages.
+#' @param max_ma `character`. The name of the column you wish to be treated as
+#' maximum ages in `occdf` and `bins`. `max_ma` in `occdf` can instead be
+#' treated as point ages (i.e. the same column name as for `min_ma`).
+#' @param interval_name `character` The name of the column in `bins` that
+#' contains the interval names.
+#' @param min_age The minimum desired occurrence age. This can be either a
+#' numeric value or an interval name, which then must be present in column
+#' `interval_name` in `bins`.
+#' @param max_age The maximum desired occurrence age. This can be either a
```
LewisAJones, line 29: As the function can take both ages and interval names (e.g. stages, periods, etc)... Is there a more generic argument name we can have here?

R/filter_time.R
```diff
@@ -0,0 +1,340 @@
...
+#' filtering fossil occurrences, such as that returned by [time_bins()]. This
+#' must contain at least three columns as named by `max_ma`, `min_ma` and
+#' `interval_name`. No column can contain missing values and columns `max_ma`
+#' and `min_ma` must additionally be numeric.
+#' @param min_ma `character`. The name of the column you wish to be treated as
+#' minimum ages in `occdf` and `bins`. `min_ma` in `occdf` can instead be
+#' treated as point ages.
+#' @param max_ma `character`. The name of the column you wish to be treated as
+#' maximum ages in `occdf` and `bins`. `max_ma` in `occdf` can instead be
+#' treated as point ages (i.e. the same column name as for `min_ma`).
```
LewisAJones, line 23: We used to explicitly state we allowed point age estimates in `bin_time` but removed it to avoid confusion. I wonder if we should just mention how users could use the function with point age estimates in the details and provide an example?

R/filter_time.R
```diff
@@ -0,0 +1,340 @@
...
+#' @param interval_name `character` The name of the column in `bins` that
```
LewisAJones, line 24:
```suggestion
#' @param interval_name `character`. The name of the column in `bins` that
```

R/filter_time.R
```diff
@@ -0,0 +1,340 @@
...
+#' contains the interval names.
+#' @param min_age The minimum desired occurrence age. This can be either a
+#' numeric value or an interval name, which then must be present in column
+#' `interval_name` in `bins`.
+#' @param max_age The maximum desired occurrence age. This can be either a
+#' numeric value or an interval name, which then must be present in column
+#' `interval_name` in `bins`.
+#' @param overlap `logical`. Occurrences which partially overlap with the set
```
LewisAJones, line 32: Should this argument name be more explicit? e.g. keep_overlap/drop_overlap?

R/filter_time.R
```diff
@@ -0,0 +1,340 @@
...
+#' filtering duration are discarded by default, so that the returned data falls
+#' strictly within these bounds. Set to `TRUE` to retain overlaps instead.
+#' @param uncertainty A numeric value denoting the maximum tolerable temporal
+#' uncertainty for retaining occurrences. Alternatively, if this argument is
```
LewisAJones, line 36: Are there instances where individuals may want to combine both `uncertainty` and `n_bins`? In general, I think it is better when arguments don't interact too much... `n_bins` could just be applied when it isn't `NULL`? 

You could also have age_uncertainty and bin_uncertainty if you wanted to be really specific?

R/filter_time.R
```diff
@@ -0,0 +1,340 @@
...
+#' set to "bins", then only occurrences which span no more than `n_bins`
+#' intervals in `bins` will be retained.
+#' @param n_bins `numeric` The maximum tolerable bin-wise uncertainty above
+#' which occurrences will be discarded if `uncertainty = "bins"`.
+#' @param retain.taxa `logical`. Complete loss of some taxa may occur if all
+#' their occurrences exceed the temporal uncertainty criterion. If this is
+#' undesirable, occurrences for these taxa will be retained as exceptions to
+#' the this filtering criterion.
+#' @param taxa `character`. The name of the column you wish to be treated as
```
LewisAJones, line 45: If we removed `retain.taxa` this could also go, helping to streamline the function a little

R/filter_time.R
```diff
@@ -0,0 +1,340 @@
...
+#' taxon names in `occdf` if `retain.taxa = TRUE`.
+#' @param verbose `logical`. Sets if the impact of time filtering is reported
```
LewisAJones, line 47: I think some users would like this report, and Bethany seems to think so too, but personally I don't like the console being flooded with messages. If you want to keep this, I'd suggest having this argument set to FALSE by default. I'd also make the messaging more generic (e.g. X occurrences dropped). This might be something we want to eventually propagate across other functionality?... I can imagine a suite of `filter_*` functions

R/filter_time.R
```diff
@@ -0,0 +1,340 @@
...
+#' will be an additional column called `retained`, consisting of `0` for
+#' occurrences which met all filtering criteria and `1` for occurrences which
+#' were retained for taxonomic completeness. Depending on the filtering
+#' parameters used, all or no occurrences may be discarded.
+#'
+#' @details
+#' When filtering by stratigraphic bounds, minimum or maximum ages can be used
+#' individually, or at the same time. By default occurrences which do not fall
+#' fully within the bounds set by `min_age` and/or `max_age` are discarded.
+#' This behaviour can be toggled by setting `strict = FALSE` to retain
```
LewisAJones, line 62: This argument doesn't exist?

R/filter_time.R
```diff
@@ -0,0 +1,340 @@
...
+#' occurrences which partially overlap with the bounds set. If using
+#' stratigraphic interval names, `min_age` is internally set to its interval
+#' upper boundary age and `max_age` to its interval lower boundary age using
+#' the interval durations in `bins`. Otherwise, the numeric minimum and maximum
+#' values are used directly.
+#'
+#' When filtering by stratigraphic precision, if `uncertainty = "bins"`, the
```
LewisAJones, line 69: This is quite a high-level interaction between functions, and `bin_time` is likely to go through some changes in the coming months. This depends a little on how we want users to use our functions, but I could imagine this being a good opportunity for showing how things can be piped together instead? However, I do like the simplicity for users of including this here. I think this is a discussion point for us on this function.

R/filter_time.R
```diff
@@ -0,0 +1,340 @@
...
+#' not reintroduced in this way, nor are occurrences with missing entries in
+#' the column specified by `taxa` (i.e., those with `NA` values).
+#'
+#' * This procedure may be convenient for downstream analyses where the taxa
+#' present in a occurrence data set must match those in a different one (e.g.,
+#' a phylogeny), but a user should carefully consider whether it is necessary
+#' or reasonable to use poorer quality data analytically.
+#'
+#' @section Developer(s): Joseph T. Flannery-Sutherland
+#' @section Reviewer(s): William Gearty, Bethany Allen, Lewis Jones
```
LewisAJones, line 117:
```suggestion
#' @section Reviewer(s): William Gearty, Bethany Allen, Lewis A. Jones
```

R/filter_time.R
```diff
@@ -0,0 +1,340 @@
...
+#'
+#' @examples
+#' # get internal tetrapod data and GTS2020 stages
```
LewisAJones, line 120: Standardising to other function examples docs (implement for all)
```suggestion
#' # Get internal tetrapod data and GTS2020 stages
```

R/filter_time.R
```diff
@@ -0,0 +1,340 @@
...
+#' occdf <- tetrapods[1:100, ]
+#' stages <- time_bins(rank = "stage")
+#'
+#' # trim to Kungurian occurrences
```
LewisAJones, line 124: I think we should avoid using "trim" and maybe stick to filter, remove, or exclude

R/filter_time.R
```diff
@@ -0,0 +1,340 @@
...
+#'                         min_age = "Wuchiapingian", max_age = "Asselian",
+#'                         uncertainty = 20, retain.taxa = TRUE, taxa = "genus")
+#'
+#' @export
+
+filter_time <- function(occdf, bins = NULL,
+                        min_ma = "min_ma", max_ma = "max_ma", interval_name = "interval_name",
+                        min_age = NULL, max_age = NULL, overlap = FALSE,
+                        uncertainty = NULL, n_bins = 1,
+                        retain.taxa = FALSE, taxa = NULL, verbose = T) {
```
LewisAJones, line 158:
```suggestion
                        retain.taxa = FALSE, taxa = NULL, verbose = FALSE) {
```

R/filter_time.R
```diff
@@ -0,0 +1,340 @@
...
+
+  # occdf <- data.frame(genus   = c("A", "A", "B", "B", "B", "C", "D", "E", "E", "F"),
```
LewisAJones, line 160: A reminder to remove this for function testing!

R/filter_time.R
```diff
@@ -0,0 +1,340 @@
...
+  # overlap = F
+  # uncertainty = "bins"
+  # n_bins = 2
+  # retain.taxa = FALSE
+  # taxa = NULL
+  # verbose = T
+
+  # check occurrence data and copy for filtering
+  if (is.data.frame(occdf) == FALSE) {
+    stop("`occdf` should be a data.frame.")
```
LewisAJones, line 186: This needs changing elsewhere in palaeoverse but I guess "should" implies some flexibility (which is not correct)
```suggestion
    stop("`occdf` must be a data.frame.")
```

R/filter_time.R
```diff
@@ -0,0 +1,340 @@
...
+  }
+
+  # check arguments
+  if (any(c(!is.atomic(min_ma), length(min_ma) != 1, !min_ma %in% colnames(occdf),
```
LewisAJones, line 190: Should this be `is.character`? 
```suggestion
  if (any(c(!is.character(min_ma), length(min_ma) != 1, !min_ma %in% colnames(occdf),
```

I would also say you're checking multiple conditions here and the error message doesn't necessarily help for some cases. For example, imagine a user makes a typo in the column name.

R/filter_time.R
```diff
@@ -0,0 +1,340 @@
...
+            !is.atomic(max_ma), length(max_ma) != 1, !max_ma %in% colnames(occdf)))) {
+    stop("Please specify `min_ma` and `max_ma` as column names in `occdf`.")
+  }
+  if (!is.null(max_ma)) {
+    if (any(c(!is.atomic(min_ma), length(min_ma) != 1, !min_ma %in% colnames(occdf)))) {
+      stop("If not NULL, please specify `max_ma` as a column name in `occdf`.")
+    }
+  }
+  if (any(is.na(occdf[, min_ma])) || any(is.na(occdf[, max_ma]))) {
```
LewisAJones, line 199: Could also be split out to help the user locate issues in their dataset?

R/filter_time.R
```diff
@@ -0,0 +1,340 @@
...
+    stop(paste0("NA values detected in ", min_ma, " or ", max_ma, " in `occdf`"))
+  }
+  if (!is.numeric(occdf[, min_ma]) || !is.numeric(occdf[, max_ma])) {
```
LewisAJones, line 202: Could also be split out to help the user locate issues in their dataset?

R/filter_time.R
```diff
@@ -0,0 +1,340 @@
...
+      stop(paste0("Either: ", interval_name, ", ", max_ma, " or ", min_ma, " column(s) do not exist in `bins`."))
+    }
+    if (any(is.na(bins))) {
+      stop(paste0("NA values detected in `bins`."))
+    }
+    if (!is.numeric(bins[, min_ma]) || !is.numeric(bins[, max_ma])) {
+      stop("Columns `min_ma` and `max_ma` in `bins` should contain numeric values.")
+    }
+    if (!any(c(bins[, max_ma], bins[, min_ma]) %in% c(occdf[, max_ma], occdf[, min_ma]))) {
+      warning("No shared ages between `bins` and `occdf`. Check the time scale and range of your data.")
```
LewisAJones, line 229: Do they need to be shared?

I imagine we would want to check whether they are in range of each other and overlap (i.e. test maximum and minimum ages) but not whether exact numeric ages match?

R/filter_time.R
```diff
@@ -0,0 +1,340 @@
...
+#' # trim to occurrences within the stages of the Permian, and filter to those
+#' # with age uncertainties less than or equal to 15 million years, but allow
+#' # some less precise occurrences to remain to prevent loss of any taxa.
+#' trimmed7 <- filter_time(occdf, bins = stages,
+#'                         min_age = "Wuchiapingian", max_age = "Asselian",
+#'                         uncertainty = 20, retain.taxa = TRUE, taxa = "genus")
+#'
+#' @export
+
+filter_time <- function(occdf, bins = NULL,
```
LewisAJones, line 154: Is there value in having this split into a couple of functions? Something along the lines of:

- filter_by_age
- filter_by_interval

This could clean up and streamline a lot of the documentation?

R/filter_time.R
```diff
@@ -0,0 +1,340 @@
...
+    }
+    min_age <- bins[which(bins[, interval_name] == min_age), min_ma]
+  }
+  if (is.character(max_age)) {
+    if (!max_age %in% bins[, interval_name]) {
+      stop("`max_age` is not an interval name in `bins`.")
+    }
+    max_age <- bins[which(bins[, interval_name] == max_age), max_ma]
+  }
+  if (is.numeric(min_age)) {
```
LewisAJones, line 247: I guess we should have already checked whether min_age / max_age are character or numeric in the error handling, so we could probably just use if and else statements for L234--258? Equally, this is also fine!

R/filter_time.R
```diff
@@ -0,0 +1,340 @@
...
+    occdf <- occdf[which(occdf[,max_ma] > min_age),]
+    if (!overlap) {
+      occdf <- occdf[which(occdf[,min_ma] >= min_age),]
+    }
+  }
+  if (is.numeric(max_age)) {
+    occdf <- occdf[which(occdf[,min_ma] < max_age),]
```
LewisAJones, line 254: Small edit for code style/consistency (check throughout)
```suggestion
    occdf <- occdf[which(occdf[ ,min_ma] < max_age),]
```

R/filter_time.R
```diff
@@ -0,0 +1,340 @@
...
+    if (!overlap) {
+      occdf <- occdf[which(occdf[,max_ma] <= max_age),]
+    }
+  }
+  if (all(!is.null(min_age), !is.null(max_age))) {
+    if (min_age >= max_age) {
+      stop("The minimum filtering age is not younger than than the maximum filtering age.")
```
LewisAJones, line 261:
```suggestion
      stop("The minimum filtering age is greater than the maximum filtering age.")
```

LewisAJones (changes requested):
Hi @jf15558, sorry for the delay on reviewing this. I've now given this a once over, and I really like the intended purpose. 

As you'll see, I have a few comments, but my main point for discussion is whether this function should actually be split out into a couple of functions? Currently, you can filter using numeric ages or interval names. I like having these options but I wonder if it makes more sense to have a filter_by_age and filter_by_interval (or similar) function. This would streamline a lot of the documentation and the internal code making both easier to maintain and support the user in getting started quickly.

Minor points:
- I came across a few instances of typos, or outdated documentation, so it's probably worth doing a proofread before the next review stage.
- There are some cases where error handling and messaging could be clearer for the user (e.g. multiple error handling tests but the error message wouldn't always help the user in debugging).
- There are a few instances where code style does not strictly follow our standards ([tidyverse style guide](https://style.tidyverse.org)). We still need to improve some of this in the broader package, but better to catch this now!

I'll give this another, more thorough review once we've decided on the ultimate direction of this function. I don't think too fine edits are helpful at this stage!

PR #146, opened 2025-08-20
Switch to RETRY

### LINKED ISSUE #137: Tests failing due to unresponsive GPlates and Zenodo servers
Occasionally, our checks fail because one or more `palaeorotate()` tests fail. This appears to happen due to the GPlates and/or Zenodo servers being unresponsive. We used to `RETRY` for our GPlates API calls (https://github.com/palaeoverse/palaeoverse/pull/113), but we removed that because the API can now return 400 errors when the age of rotation is outside of the scope of a given model. However, I'd recommend putting `RETRY` back in for these API calls and adding it in for Zenodo downloads. If we're worried about catching the 400 Bad Requests errors, we can always specify those in the `RETRY` call (https://httr.r-lib.org/reference/RETRY.html) using the "terminate_on" argument.

willgearty:
This replaces GPlates and Zenodo downloads (currently `httr::GET` and `download.file`) with `httr::RETRY` in `palaeorotate()`. This should hopefully reduce the number of times the tests fail due to download errors.

Fixes #137.


LewisAJones (approved):
Not much to comment on here @willgearty, thanks for those improvements... hopefully this does indeed reduce the number of download errors. **Just a quick reminder to remember to update the NEWS.md.** Thanks again!

PR #158, opened 2026-04-30
Fix stale links in docs

etiennebacher:
# Function name

## Description

Related to #148. I've run `urlchecker` and `lychee` locally but they both have false positives, which is why I haven't added a CI workflow for this yet:

* `urlchecker` keeps reporting https://img.shields.io/twitter/url.svg?style=social&label=Follow%20%40ThePalaeoverse&url=https%3A%2F%2Ftwitter.com%2FThePalaeoverse as 403 (Forbidden) but this link is valid
* `lychee` reports https://besjournals.onlinelibrary.wiley.com/doi/10.1111/2041-210X.14099 as 403 (Forbidden) but it also works. It might be due to Cloudflare check?

I guess #148 shouldn't be closed until we have proper CI to check dead links.

Also, I just updated the URL to the "Structure and standards" so that it's no longer stale, but we probably want this to be stored on the Palaeoverse website and not on `palaeoverse` website, so this change might need to be reverted later.	

## Checklist before requesting a review
- [x] I have read palaeoverse's [standards and structure](https://palaeoverse.org/articles/structure-and-standards.html) 
- [x] I have performed a self-review of my code.
- [ ] I have added function documentation.
- [ ] I have provided sufficient examples of implementation.
- [ ] If it is a core feature, I have added thorough tests.

PR #160, opened 2026-05-11
Add `lychee` workflow to detect dead links

### LINKED ISSUE #148: Stale links in CONTRIBUTING.md
I noticed that there are some stale links in the contributing docs, including at least the following:

https://github.com/palaeoverse/palaeoverse/blob/91c99ef9ca02d08615ea774fccfb399e643b573c/CONTRIBUTING.md?plain=1#L3

https://github.com/palaeoverse/palaeoverse/blob/91c99ef9ca02d08615ea774fccfb399e643b573c/CONTRIBUTING.md?plain=1#L49

etiennebacher:
`lychee .` always fails because of some edge cases (see comments in the CI workflow). It's not ideal but at least we now detect failures in all other links.

Companion PR to #158, would fix #148


LewisAJones (changes requested):
I think we should also add a schedule trigger (e.g. every Monday at 08:00)? Otherwise, this looks good to me @etiennebacher, thanks!

It's worth @willgearty also having a quick look at this before we merge (much more familiar with CI workflows than I).

willgearty (approved):
Looks good to me. I worry that we'll just need to keep excluding more and more links in the future as more websites block automated web traffic, but that's a problem for another day, I suppose.

I agree a cron job would be good to check the links regularly (we probably should have our R-cmd-check workflow run regularly as well).

etiennebacher:
Thanks, I've added the cron job. I have some ideas of extra CI to run that I want to talk about in the meeting today so let's talk about cron jobs for other workflows as well.

LewisAJones (approved):

Issue #165, opened 2026-05-14
Meta-issue on infrastructure


etiennebacher:
I chose to make a meta-issue in the end, it's just easier to see progress than with small issues scattered around:

- automatic refactor:
  - [x] format with Air (#170)
  - [x] lint with Jarl (#169)
- CI:
  - [x] combine Air and Jarl in a single workflow and post a comment if they fail explaining how to solve it (#175)
  - [x] cover R versions until 4.1 (included) (#181)
  - [x] R CMD check without 'Suggests' dependencies (#182)
  - [x] add spelling check (#180)
  - [x] update versions of steps used, e.g. `actions/checkout` (#195)
  - [x] add cron jobs for R CMD check (#211)

willgearty:
How do you feel about subissues @etiennebacher?

etiennebacher:
I don't really see the value added compared to just referring to this issue to see what's done and what's left to do. 

I just created one by accident 😄 can you delete it? I don't see this option so maybe I don't have the right for this

etiennebacher:
@willgearty do you remember why we wanted cron jobs for R CMD check? In #182 you said:

> The normal R-CMD-checks are the ones we should be running as a cron job regularly in case CRAN packages are removed.

but CRAN will notify authors whose packages depend on packages that are removed from CRAN. Now that we have the workflow from #182 we shouldn't be affected if a package in `Suggests` is removed from CRAN, so I don't see in which case a cron job on R CMD check will give us new information.

willgearty:
CRAN will only check the released version of the package. We need to also check that the development version of the package isn't breaking due to upstream changes.

etiennebacher:
🎉

PR #169, opened 2026-05-14
Apply Jarl automatic fixes

etiennebacher:
Almost all changes in this PR were automatically generated by [Jarl](https://jarl.etiennebacher.com/) with `jarl check . --fix`.

A quick summary of those changes:

- readability: 
  - replace cases like `x == FALSE` by `!x`, and `x == TRUE` by `x`
  - `sort(x)` is more readable than `x[order(x)]`
  - add leading 0 to decimal numbers
- performance:
  - `anyNA(...)` is way faster than `any(is.na(...))`
  - `anyDuplicated(...) > 0` is way faster than `any(duplicated(...))`
- correctness
  - `1:nrow(...)` likely produces wrong results when the `...` has 0 rows. It's more robust to use `seq_len(nrow(...))` instead

Part of #165 


R/bin_time.R
```diff
@@ -168,8 +168,7 @@ bin_time <- function(occdf, min_ma = "min_ma", max_ma = "max_ma",
     #=== Reporting Info ===
 
     # Make an empty list that's the length of the occurrence dataframe.
-    bin_list <- list()
-    bin_list <- sapply(seq_len(nrow(occdf)), function(x) NULL)
+    bin_list <- vector("list", length = nrow(occdf))
```
etiennebacher: This was done by hand. There are two problems here:

- the first definition of `bin_list` is never used
- `vector("list", length = nrow(occdf))` is way faster than the `sapply()` approach

R/release.R
```diff
@@ -1,5 +0,0 @@
-release_questions <- function() {
-  c(
-    "Did you re-build the vignettes using `vignettes/precompile.R`?"
-  )
-}
```
etiennebacher, line 5: I might have missed the context where this is used, but it's not exported and not used anywhere in the package so I suppose it's just a leftover from an old implementation?

R/look_up.R
```diff
@@ -411,15 +411,15 @@ look_up <- function(occdf, early_interval = "early_interval",
...
+    if (!return_unassigned && length(unassigned) >= 1) {
       warning(
     c("The following intervals could not be matched with intervals from int_key
       or GTS: ", paste(unassigned, collapse = ", "))
       )
     }
 
   # return output
-  if (return_unassigned && length(unassigned >= 1)) {
+  if (return_unassigned && length(unassigned) >= 1) {
```
etiennebacher, line 514: This was likely a bug.

R/release.R
willgearty: I believe this is used by `usethis::use_release_issue()`

etiennebacher: Ah interesting, I'll edit that

willgearty (approved):
Looks good to me aside for the `release.R` deletion.

R/release.R
etiennebacher: Done

etiennebacher:
Hi @willgearty, the last two commits add a few more fixes I should included in this PR from the beginning (sorry, I forgot about this):

1. it fixes `testthat`-specific lints, for instance to use `expect_type()` because it gives better messages in case of test failures:

``` r
testthat::expect_true(is.list(1))
#> Error:
#> ! Expected `is.list(1)` to be TRUE.
#> Differences:
#> `actual`:   FALSE
#> `expected`: TRUE
testthat::expect_type(1, "list")
#> Error:
#> ! Expected 1 to have type "list".
#> Actual type: "double"

testthat::expect_true(is.character(1))
#> Error:
#> ! Expected `is.character(1)` to be TRUE.
#> Differences:
#> `actual`:   FALSE
#> `expected`: TRUE
testthat::expect_type(1, "character")
#> Error:
#> ! Expected 1 to have type "character".
#> Actual type: "double"
```

2. it adds the configuration file `jarl.toml` which contains a few settings to ensure that everyone running Jarl on this project uses the same config.

vignettes/structure-and-standards.Rmd
```diff
@@ -96,7 +96,11 @@ External data is the preferred approach for contributors to include data into `p
 #generate temp directory
 files <- tempdir()
 #download files
-download.file(url = "www.goo.com", destfile = paste0(files, "/mgoo.csv"))
+download.file(
+  url = "www.goo.com",
+  destfile = paste0(files, "/mgoo.csv"),
+  mode = "wb"
```
willgearty, line 102: Note that `mode = "wb"` can have unintended consequences on Windows for text files: https://stat.ethz.ch/R-manual/R-devel/library/utils/html/download.file.html

tests/testthat/test-look_up.R
```diff
@@ -13,7 +13,7 @@ test_that("look_up() works", {
 
   # check correct format and output
   expect_true(is.data.frame(look_up(occdf[1:10, ])))
-  expect_true(is.character((look_up(occdf[1:10, ]))$early_stage))
+  expect_type((look_up(occdf[1:10, ]))$early_stage, "character")
   expect_true(is.numeric((look_up(occdf[1:10, ]))$interval_min_ma))
```
willgearty, line 16: Why not for the lines above and below?

tests/testthat/test-time_bins.R
```diff
@@ -13,11 +13,11 @@ test_that("time_bins() works with macrostrat", {
...
-                                assign = c(232, 167, 33), plot = TRUE)))
-  expect_true(is.list(time_bins(
+  expect_type(time_bins(interval = c("Fortunian", "Meghalayan"),
+                                assign = c(232, 167, 33), plot = TRUE), "list")
+  expect_type(time_bins(
     interval = c("Fortunian", "Holocene"), scale = "GTS2012",
-    assign = c(232, 167, 33), plot = TRUE)))
+    assign = c(232, 167, 33), plot = TRUE), "list")
   expect_true(is.vector(time_bins(interval = c("Fortunian", "Meghalayan"),
                                   assign = c(232, 167, 33))$Assignation))
```
willgearty: Why not also for the lines above and below?

tests/testthat/test-look_up.R
etiennebacher: Because they don't have a clear `expect_type()` replacement:

``` r
library(testthat)

expect_true(is.data.frame(iris))
expect_type(iris, "data.frame")
#> Error:
#> ! Expected `iris` to have type "data.frame".
#> Actual type: "list"

expect_true(is.numeric(1))
expect_type(1, "numeric")
#> Error:
#> ! Expected 1 to have type "numeric".
#> Actual type: "double"

expect_true(is.vector(1:2))
expect_type(1:2, "vector")
#> Error:
#> ! Expected `1:2` to have type "vector".
#> Actual type: "integer"
```

I don't mind reverting those changes if you'd rather keep `expect_true(...)` everywhere for consistency.

vignettes/structure-and-standards.Rmd
etiennebacher: Interesting, I had followed `lintr` when implementing this in Jarl: https://lintr.r-lib.org/reference/download_file_linter.html

And `lintr` implemented this following this issue in `goodpractice`: https://github.com/ropensci-review-tools/goodpractice/issues/63

tests/testthat/test-look_up.R
willgearty: For data.frames, it looks like you can use 
`expect_s3_class(object, "data.frame")`

And `is.numeric` can probably be replaced with `expect_type(object, "double")` in most places, right? I don't think there are any places where we specifically cast to integers.

etiennebacher: > For data.frames, it looks like you can use expect_s3_class(object, "data.frame")

That should work, yes.

> And is.numeric can probably be replaced with expect_type(object, "double") in most places, right? 

That would probably work here then. However,  this will not be detected by Jarl since in general `expect_true(is.numeric(...))` isn't equivalent to `expect_type(..., "double")`. So I can make the change here, but future occurrences of `expect_true(is.numeric(...))` won't be automatically caught.

willgearty: Fine either way.

vignettes/structure-and-standards.Rmd
willgearty: Ah, yes, I had it backwards in my mind. `"w"` is the bad one on Windows.

ref: https://github.com/palaeoverse/2025-CPEG-workshop/commit/5d3f6fe5d75d17c446c5b23d6b8fda55a40ba9f7

tests/testthat/test-look_up.R
etiennebacher: I have reverted the `expect_type` changes, it adds confusion for little benefits.

willgearty: Sounds good 👍

vignettes/structure-and-standards.Rmd
LewisAJones: Always good to double check, this caused us a headache for some time!

LewisAJones (approved):
Looks good to me @etiennebacher!

PR #175, opened 2026-05-18
Set up Github workflow to post a comment when Air and/or Jarl fail

etiennebacher:
This adds two connected workflows to detect CI failure for Air and Jarl. It also posts a comment indicating which one failed and how to fix it locally. This could help new contributors who might be confused on the steps to follow when those checks fail. Note that if both pass on the first try, then we don't publish a comment to say that they pass since that would add noise IMO.  

Those two workflows are:

- `format-lint.yaml`: runs Air and Jarl and stores their results (just a success/failure indicator) in Github artifacts
- `format-lint-pr-comment.yaml`: takes those results and creates an appropriate Github comment 

While it is technically possible to have both in a single workflow using the `pull_request_target` event, it is discouraged to do so for security purposes (see link at the top of `format-lint.yaml`).

Finally, the comment will be published on PRs only once this branch is merged, so we cannot preview what it will look like (but you can refer to the templates to get an idea).

Most the workflows were generated by Claude but I have reviewed them and I've written the templates. 

Part of #165, but it requires #169 to be merged first.


willgearty:
Should we be using commits for the GitHub actions in line with #195?

willgearty (approved):

.github/workflows/templates/air-only.md
```diff
@@ -0,0 +1,6 @@
+<!-- format-lint-check -->
+:x: The formatting check failed. Please install [air](https://posit-dev.github.io/air/cli.html) and run the following command in the terminal (not the R console) to reformat the code:
```
LewisAJones: Should we be a little more verbose and set the expectation here? I am just thinking about potential contributors who have never heard or air/jarl/lint. 

Maybe something along the lines of "This Pull Request failed our automated formatting checks. Please resolve these prior to requesting review. To do so..."

LewisAJones (approved):
This is great for so many reasons, but especially because it should speed up the code review process. I just have one tiny comment!

Should we also do a test PR after this is merged to make sure this is behaving as expected?

.github/workflows/templates/air-only.md
etiennebacher: Done in [5d67445](https://github.com/palaeoverse/palaeoverse/pull/175/commits/5d67445c7790b5b27acaab68517be97ebc1643be), what do you think?

etiennebacher:
> Should we also do a test PR after this is merged to make sure this is behaving as expected?

Yes I'll make a test PR to check all states (no failures from the start, linting issues only, formatting issues only, issues in both).

.github/workflows/templates/air-only.md
LewisAJones: Looks great, thanks @etiennebacher!

PR #180, opened 2026-05-19
Add CI workflow to check spelling

etiennebacher:
This uses the package `spelling`, which flags misspelled words that are not already whitelisted in `inst/WORDLIST`. False positives can be stored in this list with `spelling::update_wordlist()`. This workflow is adapted from https://github.com/easystats/workflows/blob/main/.github/workflows/check-spelling.yaml.

I also fixed a couple of typos reported by `spelling`.

Part of #165 


willgearty (approved):
Looks good, thanks!

inst/WORDLIST
```diff
@@ -0,0 +1,143 @@
...
+occ
+palaeobiological
+palaeobiologists
+palaeobiology
+palaeocoordinates
+palaeolatitudinal
+palaeorotated
+palaeorotation
+palaeoverse's
+paleobiology
```
LewisAJones, line 123: en-GB is preferred over en-US (i.e. palaeobiology vs. paleobiology) in palaeoverse, but there are cases where we reference publications or names (e.g. Paleobiology Database).... How do we want to handle this? Do we automatically pass these over, or always flag them (I'm thinking about new contributors and docs)

inst/WORDLIST
```diff
@@ -0,0 +1,143 @@
...
+paleogeographic
```
LewisAJones, line 124: Same as above.

LewisAJones (approved):
This looks great thanks @etiennebacher! Just one small point for us to think about

inst/WORDLIST
etiennebacher: I don't think there's a way to distinguish those cases, the spellchecker would need extra context or the ability to ignore publication names for instance but this one cannot do that (and I don't know any spellcheckers that can). 

So both en-GB and en-US versions must be present in the wordlist and we'll need to be extra careful  when reviewing docs in future PRs.

LewisAJones: OK, then this sounds like the sensible way forward to me!

PR #181, opened 2026-05-19
Bump R dependency to >= 4.1, expand R version coverage in CI 

etiennebacher:
As discussed, this bumps the R version constraint and ensures that all subsequent versions are tested in CI.

At the time of writing:

- release = 4.6
- oldrel-1 = 4.5
- oldrel-2 = 4.4
- oldrel-3 = 4.3
- oldrel-4 = 4.2
- oldrel-5 = 4.1 (this one is much slower to run because it doesn't have binaries for many packages that take a while to compile, e.g. `s2`)

This PR adds two small fixes for failing tests on R < 4.3:

1. `array2DF()` was added in R 4.3 so I copied the current implementation and we use this if R < 4.3
2. `by()` started accepting formulas in R 4.3 only so I replaced the input to be  a dataframe instead

Part of #165 


.github/workflows/R-CMD-check.yaml
```diff
@@ -18,11 +18,15 @@ jobs:
       fail-fast: false
       matrix:
         config:
+          - {os: ubuntu-latest,  r: 'devel', http-user-agent: 'release'}
```
etiennebacher, line 50: This is just reordered so that we have devel -> release -> oldrel in the list

.github/workflows/R-CMD-check.yaml
```diff
@@ -39,6 +47,16 @@ jobs:
...
 
+      # Some packages (e.g. geosphere) have compilation errors on R 4.1 and 4.2 because by default they
+      # use C++14 while they require C++17. We manually set this required C++ version for those R 
+      # versions only.
+      - name: Set C++17 compiler for older R versions
+        if: ${{ matrix.config.r == 'oldrel-4' || matrix.config.r == 'oldrel-5' }}
+        run: |
+          mkdir -p ~/.R
+          echo "CXX = g++ -std=gnu++17" >> ~/.R/Makevars
+        shell: bash
```
etiennebacher, line 83: This fix was generated by Claude (after many failed attempts by me 😅), I checked this locally as well

R/utils-backports.R
```diff
@@ -0,0 +1,79 @@
+# Copied from the R 4.6.0 implementation of array2DF() on 2026-05-19
```
etiennebacher, line 1: `array2DF()` was added in R 4.3.0. I think this is probably the easiest way to replace it in R < 4.3, but there's always a risk that this becomes outdated if we forget to add potential future bug fixes in the original implementation.

.github/workflows/R-CMD-check.yaml
```diff
@@ -8,6 +8,10 @@ on:
 
 name: R-CMD-check
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref }}
+  cancel-in-progress: true
```
etiennebacher, line 38: This avoids queuing too many workflows when we push several commits in a short time span. The newest commit always cancels the running workflows so that it can run its workflows. The only downside is that sometimes you receive many "job cancelled" notifications from Github.

etiennebacher:
Still in draft because it seems that tests on 4.1 are stuck because of tests requiring access to gplates.org:

```
  Error in curl::curl_fetch_memory(url, handle = handle): Failure when receiving data from the peer [gws.gplates.org]:
  Recv failure: Connection reset by peer
```

Edit: this is unrelated to R 4.1, I see similar failures here: https://github.com/palaeoverse/palaeoverse/pull/182

tests/testthat/test-group_apply.R
```diff
@@ -152,17 +158,16 @@ test_that("error handling for argument 'group'", {
     error = TRUE
   )
 
-  # This is a corner case that gives an unhelpful error message in an interactive
+  # This is a corner case that used to give an unhelpful error message in an interactive
```
etiennebacher: Turns out that the fix made in `group_apply()` to ensure it works with older versions of R also fixes this issue (i.e. it gives a better error message now).

.github/workflows/R-CMD-check.yaml
```diff
@@ -64,6 +72,16 @@ jobs:
           http-user-agent: ${{ matrix.config.http-user-agent }}
           use-public-rspm: true
 
+      # Some packages (e.g. geosphere) have compilation errors on R 4.1 and 4.2 because by default they
+      # use C++14 while they require C++17. We manually set this required C++ version for those R 
+      # versions only.
+      - name: Set C++17 compiler for older R versions
+        if: ${{ matrix.config.r == 'oldrel-4' || matrix.config.r == 'oldrel-5' }}
```
willgearty: Hmm...I guess we'll need to update this with each R version? Is there a way to check what the actual R version is instead?

willgearty: I'm also a little surprised pak doesn't deal with this itself

R/group_apply.R
```diff
@@ -115,12 +115,23 @@ group_apply <- function(occdf, group, fun, ...) {
...
+  # `by()` is a wrapper of `tapply()`, but it ends up being MUCH faster than `tapply()`
+  # because of some data wrangling it does.
+  #
+  # Note: this used to use a formula in the INDICES param, but `by.data.frame()` started
+  # accepting formulas in R 4.3 only. From R 4.3 news:
+  #    "The tapply() function now accepts a data frame as its X argument, and allows INDEX
+  #    to be a formula in that case. by.data.frame() similarly allows INDICES to be a
+  #    formula."
+  #
+  # -> since we support R >= 4.1, we cannot use a formula,
```
willgearty: Since we were just making the formula for the user (and never accepting one from the user), I don't think this comment is necessary (it's valid, just not necessary). It's just cleaning up the old convoluted code.

R/utils-backports.R
willgearty, line 1: Good catch, I hadn't realized this was only added in 4.3.0

tests/testthat/test-group_apply.R
```diff
@@ -152,17 +158,16 @@ test_that("error handling for argument 'group'", {
...
   #    ! cannot xtfrm data frames
   #
-  # The behavior is correct (it should error) but the message should be improved.
-  # This only occurs when "group" contains a character value that has the same
+  # This only occurred when "group" contains a character value that has the same
   # name as an existing object in the environment and when "group" has more than
   # one value.
   #
-  # It seems that group_apply() tries to evaluate the object but it shouldn't.
+  # It seems that group_apply() tried to evaluate the object.
```
willgearty: Do we still need this comment if we're happy with the behavior now? It's not clear to me that the old behavior is important to mention.

willgearty (changes requested):

R/group_apply.R
etiennebacher: Yeah, originally I had put this to ensure we don't try to reuse the formula in the future, but after this PR this would be caught by CI anyway

tests/testthat/test-group_apply.R
etiennebacher: I can shorten it and link to this PR, but I think it's good to keep some info about why we test this edge case in particular

.github/workflows/R-CMD-check.yaml
etiennebacher: Me too, but I remember checking this locally and `pak` didn't help.

> Is there a way to check what the actual R version is instead?

We could use the version numbers instead of `oldrel` (https://github.com/r-lib/actions/tree/v2/setup-r#inputs) but then we'd need some manual tweaks every time a new R version is available. I think we'll have some manual tweaks to do anyway.

willgearty: Could the `if` condition somehow depend on the installed R version (like `if (getRversion() <= "4.2.0")`)? Or even make this step always run a small bit of R code that checks the R version first?

etiennebacher: Good idea, I made the change in [4117fef](https://github.com/palaeoverse/palaeoverse/pull/181/commits/4117feff60daa8c872df2d833507a0bb727eb784)

.github/workflows/R-CMD-check.yaml
```diff
@@ -64,6 +72,18 @@ jobs:
           http-user-agent: ${{ matrix.config.http-user-agent }}
           use-public-rspm: true
 
+      # Some packages (e.g. geosphere) have compilation errors on R 4.1 and 4.2 because by default they
+      # use C++14 while they require C++17. We manually set this required C++ version for those R 
+      # versions only. We use getRversion() instead of oldrel-X logic so that this block doesn't keep
+      # running when new R version become available, even if we forget to remove it.
```
willgearty: I don't see a need to document the evolution of this over the course of this PR
```suggestion
      # use C++14 while they require C++17.
```

willgearty (approved):

willgearty:
Thanks for addressing my requests!

R/utils-backports.R
LewisAJones: Nice! Worth adding a note about this, and indicating when this could/should be removed?

LewisAJones (approved):

R/utils-backports.R
etiennebacher: Done in 2fe911d39d2d26d88cb418fd69ac6c77862cf4e6

PR #182, opened 2026-05-19
Add CI workflow for R CMD check without 'Suggests' dependencies

etiennebacher:
This is very similar to R-CMD-check but it doesn't install any dependencies in  'Suggests'. The objective is to ensure that our package will not be threatened to be removed from CRAN if one of those 'Suggests' package is removed.

I also added a condition for an example that requires `paleotree`.

Part of #165 


willgearty:
Add a cron job to check weekly?

etiennebacher:
I don't think this is necessary because we don't install those Suggests packages anyway, so I don't see what could be uncovered by running this workflow weekly. Do you have something specific in mind?

willgearty:
Sorry, I read this too quickly. The normal R-CMD-checks are the ones we should be running as a cron job regularly in case CRAN packages are removed.

etiennebacher:
Right, I'll add this to the tracking issue, thanks

willgearty (approved):
Looks good, thanks!

R/phylo_check.R
```diff
@@ -36,7 +36,7 @@
 #' @section Reviewer(s):
 #' William Gearty & Pedro Godoy
 #'
-#' @examples
+#' @examplesIf requireNamespace("paleotree", quietly = TRUE)
```
LewisAJones, line 39: I think this is great for now, but depending on what comes out of the refactor, we might want to consider whether we just make the data (i.e. a tree) available in the package as we do with `tetrapods` and `reefs`. It would be good for examples/tests to be able to run without Suggests.

LewisAJones (approved):

PR #185, opened 2026-05-20
Update docs and tests to use `axis_geo(..., title = ...)`

### LINKED ISSUE #136: Update everything to use `axis_geo(..., title = ...)`
#134 added the ability to add a title as part of `axis_geo()`. We should update all of our docs, functions, tests, and vignettes to use this more user-friendly method to add a title to plots.

etiennebacher:
This replaces the combination of `axis_geo(...)` + `title(...)` by `axis_geo(..., title = ...)`. I followed #134, meaning that the `line` argument that is sometimes present in `title()` isn't moved to `axis_geo()`. This leads to slight differences in plots, see the updated snapshots. For `axis-geo-with-autofit` and `axis-geo-with-time-bins-scale`, I've slightly expanded the bottom margin otherwise the title was truncated. Something similar was already done in  `axis-geo-with-autofit2`.

I've checked all cases, there is just one case in the docs of `tax_range_strat()` that shouldn't be modified because `axis_geo()` is a component of a larger plot and therefore shouldn't take the `title` arg:

```
#' tax_range_strat(occdf, name = "taxon")
#' ...
#' axis_geo(side = 4, intervals = eras_custom, tick_labels = FALSE)
#' title(xlab = "Taxon", line = 10.5)    
``` 

Fixes #136


willgearty (changes requested):
Everything looks good so far, thanks. Can you also please update any time we use `axis_geo()` in any function documentation?
For example:
https://github.com/palaeoverse/palaeoverse/blob/ffaa745366e07296d2d2d4082cb95910e1220f57/R/tax_range_strat.R#L108-L109

etiennebacher:
This is the example I mentioned in the OP. In this case, `title()` updates the plot generated by `tax_range_strat()` and not the one generated by `axis_geo()`, so adding the `title` arg in `axis_geo()` and dropping `title()` would remove the title from the overall plot.

willgearty (approved):
Oh, good catch! Huh, I had assumed there were more usages of `axis_geo()` across the docs, but it looks like there aren't, so this is good to go.

etiennebacher:
I've also tried to re-knit the vignettes with `vignettes/precompile.R` since some code has changed there but it generates massive diffs unrelated to this PR (and I see some new errors too), I think they are outdated?

Edit: see https://github.com/palaeoverse/palaeoverse/pull/186

willgearty:
Yeah, you can skip the full recompile for now. A lot of the outputs are OS-dependent for some reason. In the future, maybe that should be some sort of post-commit automation instead of local rendering.

LewisAJones (approved):
Looks great!

PR #186, opened 2026-05-20
New errors when recompiling tetrapod biodiversity vignette

etiennebacher:
I ran `vignettes/precompile.R` after commenting out the line to knit `_phanerozoic-reefs.Rmd` (I was not interested in this vignette yet).

There are some new errors and some output changed so this vignette needs to be updated. I open this and not a new issue so that we can see the diff.


vignettes/tetrapod-biodiversity.Rmd
```diff
@@ -422,12 +441,12 @@ head(coll_taxa)
...
-## 3      3        106558
-## 4      8        108766
-## 5      3        109702
-## 6      1         11149
+## 1 104796             1
+## 2 106270             1
+## 3 106558             3
+## 4 108766             8
+## 5 109702             3
+## 6  11149             1
```
etiennebacher, line 449: This is the issue: the two columns got swapped somehow.

LewisAJones: Ah, this is probably related to changes in #139. I believe these changes switched the grouping column to be the first column(s) rather than the last column(s). In this vignette, we're manually specifying the column names too which doesn't add much resilience to changes.

This highlights the need for a better workflow to make sure changes are not breaking our vignettes. This would be good to resolve before the next release. Thanks for spotting this @etiennebacher!

LewisAJones (changes requested):

etiennebacher:
This is superseded by #198. I will explore how we can check vignettes in CI but I will open a new PR for this.

PR #173, opened 2026-05-18
Convert `expect_error(...)` to `expect_snapshot(..., error = TRUE)`

etiennebacher:
This PR affects a lot of files but the only changes are that we move from `expect_error(...)` to `expect_snapshot(..., error = TRUE)` (and therefore this PR also contains all the snapshots that were created). This helps ensuring that the current error messages are correctly displayed on the user side, and it will be especially useful  to see changes in formatting when we start using `rlang` and `cli`. 

Part of #161 

<details>
<summary>Tip to efficiently make this kind of refactor</summary>
<br>

I've created the package `flir` to easily make refactors using code patterns. For instance, this PR was generated with the following steps:

* `flir::setup_flir(".")`
* `flir::add_new_rule("expect_snapshot_error", ".")`
* set up the rule:
```yaml
id: expect_snapshot_error
language: r
severity: warning
rule:
  pattern: expect_error($A)
fix: expect_snapshot(~~A~~, error = TRUE)
message: foo
```
* `flir::fix_dir("tests", linters = "expect_snapshot_error")`
* in the terminal: `air format .`

and then I removed the `flir` folder.

</details>

## Checklist before requesting a review
- [x] I have read palaeoverse's [standards and structure](https://palaeoverse.palaeoverse.org/articles/structure-and-standards.html)
- [x] I have performed a self-review of my code.
- [ ] I have added function documentation.
- [ ] I have provided sufficient examples of implementation.
- [ ] If it is a core feature, I have added thorough tests.

PR #190, opened 2026-05-21
Fix `look_up()` with wrong input

etiennebacher:
I was too quick to merge #169  after integrating changes from #173  in it. One of the changes in #169 (changing `int_key != FALSE` to `int_key`) changed the error message for one case, which was fine before because this was wrapped in `expect_error()`, but after #173 we were checking the error message in detail and therefore the snapshot test failed.

This PR fixes it (it doesn't need to appear in NEWS since the failure was caused a couple of commits ago).

---

Edit: to clarify, changing `int_key != FALSE` to `int_key` would have been fine if `int_key` could only be a logical value, but here it can be something else.

### DIFF
--- a/R/look_up.R
+++ b/R/look_up.R
@@ -149,7 +149,7 @@ look_up <- function(
 
   # int_key checks
   if (!is.data.frame(int_key)) {
-    if (int_key) {
+    if (!isFALSE(int_key)) {
       stop("`int_key` should be a dataframe.")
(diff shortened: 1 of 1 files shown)


LewisAJones (approved):
As just discussed, looks good, but this will be an important thing to refactor down the line!

willgearty (approved):

Issue #184, opened 2026-05-20
Consider  the package `vcr` to record and mock HTTP requests


etiennebacher:
Tests can sometimes run for a very long time in CI because of failing requests, e.g. in #182:

```
  Error in curl::curl_fetch_memory(url, handle = handle): Timeout was reached [gws.gplates.org]:
  Operation too slow. Less than 1 bytes/sec transferred the last 600 seconds
```

Using `vcr` should improve the robustness of our test suite by storing the results of the request the first time it is made and then using this cached result on subsequent calls.

willgearty:
See my rredlist package for an example: https://github.com/ropensci/rredlist

etiennebacher:
I thought we could use this in `time_bins()` as well but this function passes the link to `read.csv()` and doesn't use a function that is supported by `vcr`, leading to this warning:

```
Warning (test-time_bins.R:4:3): time_bins() works with macrostrat
x "time-bins-north-am-mammals" cassette ejected without recording any interactions.
i Did you use {curl}, `download.file()`, or other unsupported tool?
i If you are using crul/httr/httr2, are you sure you made an HTTP request?
```

There are no other functions that need `vcr`, but it would still be nice to skip tests when the macrostrat API is down. We use `skip_if_offline(host = "macrostrat.org")` (and in `rmacrostrat`) as well but that's not enough because the site might be up while `macrostrat.org/api` is down (like last week). The issue is that `curl::nslookup()` doesn't work here:

``` r
curl::nslookup("macrostrat.org/api")
#> Error in `curl::nslookup()`:
#> ! Unable to resolve host: macrostrat.org/api
curl::nslookup("https://macrostrat.org/api")
#> Error in `curl::nslookup()`:
#> ! This looks like a URL, not a hostname
```

etiennebacher:
We could use `httr` for this:

``` r
library(httr)
http_status(GET("https://macrostrat.org/api"))
#> $category
#> [1] "Success"
#> 
#> $reason
#> [1] "OK"
#> 
#> $message
#> [1] "Success: (200) OK"
```

I'll check if this slows down the test suite a lot.

etiennebacher:
Seems like this issue isn't solved for `palaeorotate()`. Despite #193, there are many failing tests like this:

```
  Last 13 lines of output:
     11.     ├─base::tryCatch(request_perform(req, hu$handle$handle), error = function(e) e)
[...]

Issue #161, opened 2026-05-11
Possible improvements for the test suite


etiennebacher:
To be discussed in the meeting.

- [x] `expect_error()` could be replaced by `expect_snapshot(error = TRUE)` (#173)

  Having snapshot tests for errors allows to:
  - check that the error message is cleanly displayed 
  - be aware when some changes in the code lead to unexpected changes in the error messages

- [x] Split `test_that()` blocks into smaller blocks 
  
  This is already the case in some files, e.g. `test-tax_range_strat.R`, but some still have 100-line blocks that make it harder to quickly reproduce errors (e.g. `test-look_up.R`, `test-tax_unique.R`).
  - [x] axis_geo (#264)
  - [x] axis_geo_phylo (#264)
  - [x] bin_lat (#176)
  - [x] bin_space (#176)
  - [x] bin_time (#218)
  - [x] group_apply (#225)
  - ~lat_bins~ (deprecated)
  - [x] lat_bins_area (#217)
  - [x] lat_bins_degrees (#219) 
  - [x] look_up (#223)
  - [x] palaeorotate (#263)
  - [x] phylo_check (#244)
  - [x] tax_certainty (#239)
  - [x] tax_check (#241)
  - [x] tax_expand_lat (#242)
  - [x] tax_expand_time (#247)
  - [x] tax_range_space (#248)
  - [x] tax_range_strat (#249)
  - [x] tax_range_time (#250)
  - [x] tax_unique (#258)
  - [x] time_bins (#259)

- [ ] Reduce verbosity

  Nitpick but it makes it slightly harder to see if something is going wrong in tests:

  <img width="778" height="530" alt="Image" src="https://github.com/user-attachments/assets/a75bc765-e790-42ec-9d4d-e90c13381b36" />

- [x] Run tests in parallel (#178)

- [x] Use `vcr` for more robust tests requiring HTTP requests (#193)
  - #184

etiennebacher:
🎉

PR #193, opened 2026-05-27
Start using `vcr` to mock HTTP requests in tests

etiennebacher:
This PR focuses on `test-palaeororate.R`. It moves all `palaeorotate()` calls in `vcr::use_cassette()` to avoid making HTTP requests on every CI run since this can be brittle. Once you are happy with those changes, I will adapt the rest of the test files in separate PRs. 

We could add some extra CI to perform actual HTTP requests once every week for instance, just to ensure that we capture API changes in the future. This is recommended in the docs: https://docs.ropensci.org/vcr/articles/vcr.html#what-happens-if-the-api-changes. Let me know if you think we should do this and if you'd rather have this in this PR or in a separate one.

Part of #184, part of #161


tests/testthat/test-palaeorotate.R
```diff
@@ -129,37 +145,42 @@ test_that("palaeorotate() point method works", {
...
     3
   )
 
   # Check that multiple models are being returned
-  occdf <- palaeorotate(
+  grid_multi <- palaeorotate(
     occdf = occdf,
     method = "grid",
     model = c("PALEOMAP", "GOLONKA")
   )
```
etiennebacher: I didn't figure out how to put `palaeorotate(..., method = "grid")` in `use_cassette()` because it stores .RDS files in the `vcr` folder, e.g. `tests/fixtures/palaeorotate-grid-files` (since the `vcr` folder is configured to be `tests/fixtures`). When I run `devtools::test()`, `vcr` still temporarily creates a `tests/testthat/_vcr` folder to store fixtures and then those .RDS files aren't found. I don't know if I'm missing something obvious here.

willgearty: This is probably fine, Zenodo isn't really an API, so I'm not sure if using vcr for it makes sense.

tests/testthat/test-palaeorotate.R
```diff
@@ -54,21 +54,28 @@ test_that("palaeorotate() error handling works", {
 
 test_that("palaeorotate() point method works", {
   skip_if_offline(host = "gws.gplates.org")
+  skip_on_cran()
```
willgearty: If these are now wrapped with vcr, shouldn't they always work on CRAN?

tests/testthat/test-palaeorotate.R
willgearty: But the two cases below worked?

willgearty (commented):
Overall this looks good. Just a couple questions.

tests/testthat/test-palaeorotate.R
LewisAJones: I guess we don't need to test that the actual .RDS files are available. The current test isn't currently doing that anyway? It is inferring it based on available column names?

LewisAJones (approved):
Looks good to me! Just one comment in response to your point

tests/testthat/test-palaeorotate.R
willgearty: Ah, I understand now, the initial test downloads the files in the correct place, so then these ones work with vcr

tests/testthat/test-palaeorotate.R
etiennebacher: Makes sense, I just read the `vcr` docs  too fast

etiennebacher:
I don't know if I have something left to do here, the conversation above seems resolved, but maybe I missed something.

I left one `skip_on_cran()` since a couple of `palaeorotate()` calls are not put in cassettes as mentioned above.

willgearty:
I was going to review it, but you made it a draft, so I held off.

willgearty (approved):

etiennebacher:
(Sorry, didn't realize it was still in draft mode)

etiennebacher:
Thanks for the review!

Looking at `testthat` output in the failing coverage workflow,  I confirm this is the same problem that as in the  `devtools` issue mentioned in https://github.com/palaeoverse/palaeoverse/pull/217#issuecomment-4766611594

PR #195, opened 2026-05-27
Use exact commits and Dependabot in CI workflows

etiennebacher:
This PR does two things:

1. uses exact commits for external workflows (e.g. `actions/checkout`)
1. adds a weekly `dependabot` check to update external workflows

### Using exact commits

An action with a specific version (e.g. `r-lib/setup-r@v2`) can still have a change in behaviour, even if we didn't touch anything. This is because if we specify only the major version number then  we implicitly accept any change in minor version. This is how `r-lib/actions` introduces patches without people needing to bump the version in their CI workflows. 

This can be a security risk if any of those external workflows is compromised and bumps the minor version. An extra security layer is to specify the exact commit of the workflow we want to use, e.g.:

      - uses: r-lib/actions/setup-r@a51a8012b0aab7c32ef9d19bf54da93f3254335e # v2

### Using `dependabot`

This adds a weekly check to ensure that we use the latest versions of the various external workflows we use. I added a `cooldown` parameter so that this bot doesn't suggest versions that are less than two weeks old. Using a cooldown period gives slightly more security because it gives some time to the owners of external workflows to detect and fix compromised versions.

This bot will open a PR if a newer commit is available in some external workflows. This still requires a manual check to see what actually changed in those external workflows.

---

Part of https://github.com/palaeoverse/palaeoverse/issues/165

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: '/'
+    schedule:
+      interval: weekly
+    cooldown:
+      default-days: 7
+    groups:
+      ci:
+        patterns:
+          - '*'
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -5,7 +5,7 @@ updates:
     schedule:
       interval: weekly
     cooldown:
-      default-days: 7
+      default-days: 14
     groups:
       ci:
         patterns:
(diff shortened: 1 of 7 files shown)


willgearty (approved):
Cool, I didn't know about that dependabot feature! Looks good, thanks!

LewisAJones (approved):
The security risk side of this had not even crossed my mind, thanks @etiennebacher, this looks great.

PR #202, opened 2026-06-01
Create CITATION.cff

### LINKED ISSUE #191: Add CITATION.cff or .zenodo.json file
To improve how Zenodo generates our citation information, we should add one of the following (not both):

1. [CITATION.cff file](https://citation-file-format.github.io/)
2. [.zenodo.json file](https://help.zenodo.org/docs/github/describe-software/zenodo-json/) (example [here](https://github.com/palaeoverse/rphylopic/blob/main/.zenodo.json))

The former is more broadly used for citation information (and hooks into GitHub to automatically display a .bixtex entry). The latter is more specific to Zenodo.

Should be done alongside #152.

LewisAJones:
This is a first draft of our package `citation.CFF`.

I've attempted to make it comprehensive, but please do give this a look.

Going forward, we need to decide how we appropriately credit contributions to `palaeoverse`. I would like to advise people to cite both the software and the article describing the package. However, we need to make sure that those not on the original article are also given credit for contributions. For now, I've added you (@etiennebacher) to the author list of the software. However, I am keen to get other thoughts on how best to approach this.

A potential option is from version `2.0.0`, we distinguish clearly the article authors and software contributors.

Closes #191.


willgearty:
Apparently there is an undocumented `contributors` field: https://github.com/citation-file-format/citation-file-format/pull/439. Maybe that would be more appropriate for some people (or allow us to expand the list)?

Edit: looks like this isn't actually included in a release yet, so depending on who's using our CITATION.cff (Zenodo, GitHub) it may cause errors? Looks like it should be included in the next release of the schema. Furthermore, there are plans to add specific roles in the distant future, as discussed here: https://github.com/citation-file-format/citation-file-format/issues/112.

CITATION.cff
```diff
@@ -0,0 +1,134 @@
...
+    family-names: "Gearty"
+    email: "willgearty@gmail.com"
+    orcid: "https://orcid.org/0000-0003-0076-3262"
+  - given-names: "Bethany J."
+    family-names: "Allen"
+    email: "bethany.janet.allen@gfz.de"
+    orcid: "https://orcid.org/0000-0003-0282-6407"
+  - given-names: "Etienne"
+    family-names: "Bacher"
+    email: "e.bacher@ucl.ac.uk""
```
willgearty:
```suggestion
    email: "e.bacher@ucl.ac.uk"
```

etiennebacher:
```suggestion
    email: "etienne.bacher@protonmail.com"
```

LewisAJones:
I think the way I want to do this is keep the DESCRIPTION and CITATION.cff file equivalent. Essentially, once an author, always an author. 

However, in the long-term, we need to prepare some documentation on when a contributor becomes an author.

I think it's also OK to have a "Current Core Development Team" listed in the README.md (#203) which may be separate to the DESCRIPTION and CITATION.cff.

CITATION.cff
```diff
@@ -0,0 +1,134 @@
...
+  end: 2215
+  year: 2023
+  doi: "10.1111/2041-210X.14099"
+identifiers:
+  - type: doi
+    value: 10.5281/zenodo.7728638
+    description: Zenodo
+  - type: doi
+    value: 10.32614/CRAN.package.palaeoverse
+    description: CRAN
```
etiennebacher: Shouldn't we specify only one of them to cite the software? We say at the top of this file "If you use this software, please cite both the software and the associated paper.". We clearly see the paper citation in this file, but it might be confusing to see two ways to cite the software.

CITATION.cff
etiennebacher, line 1: A lot of information here is duplicated from the `DESCRIPTION`. I don't think this can be avoided but it's easy to update one and forget about the other. In other projects I have some CI workflows to check that if a file is changed, then another file must also be updated. For instance, we could have a workflow to add a label `forgot-to-update-description?` if a PR changes `CITATION.cff` but not `DESCRIPTION`. This wouldn't be a blocker because in some cases there is no need to update `DESCRIPTION`, it would just act as a small reminder. Would it be useful here?

CITATION.cff
LewisAJones: That's fair. Historically we've used Zenodo but that's because CRAN didn't provide DOIs in the past. I don't really have a strong preference on either. However, within the large Palaeoverse sphere, it might be good to continue to use Zenodo as our standard as it allows us to archive other materials (e.g. workshop materials).

CITATION.cff
LewisAJones: This sounds good to me! Could you put something together?

etiennebacher: Sure, do you want me to push to this branch directly or open a new PR?

LewisAJones: This branch is fine!

etiennebacher: Done and it looks like it works fine

LewisAJones: Thanks Etienne! Do you think we should expand this out to all of the following?

- DESCRIPTION
- CITATION.cff
- CITATION
- README.md

LewisAJones:
I've updated this PR to fully reflect `DESCRIPTION` (#203), listing package 'authors' as authors within the `CITATION.cff`. When they release the new version of `CITATION.cff`, which should allow a broader range of role specification, we will port the 'contributors' from DESCRIPTION to the CITATION.cff.

.github/workflows/check-description-citation-changes.yaml
```diff
@@ -0,0 +1,120 @@
...
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  check-changelog:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check if changelog update is needed
```
etiennebacher:
```suggestion
      - name: Check if DESCRIPTION or CITATION.cff update is needed
```

I forgot to remove this when I adapted the workflow from one of my projects.

CITATION.cff
etiennebacher: I think the first two items are already addressed, the current workflow:

-  adds the label `forgot-to-update-DESCRIPTION?` if a PR updates `CITATION.cff` but not `DESCRIPTION`
-  adds the label `forgot-to-update-CITATION?` if a PR updates `DESCRIPTION` but not `CITATION.cff`

There will already be false positives in some cases because there are several reasons to update `DESCRIPTION` that are unrelated to the citation. I think detecting changes (or the absence of changes) in other  files would just add noise, so I wouldn't expand this workflow.

CITATION.cff
```diff
@@ -0,0 +1,103 @@
...
+  start: 2205
+  end: 2215
+  year: 2023
+  doi: "10.1111/2041-210X.14099"
+identifiers:
+  - type: doi
+    value: 10.5281/zenodo.7728638
+    description: Zenodo
+repository-code: "https://github.com/palaeoverse/palaeoverse"
+url: "https://palaeoverse.org""
```
etiennebacher:
```suggestion
url: "https://palaeoverse.org"
```

etiennebacher (approved):
Thanks!

willgearty (approved):

Issue #132, opened 2024-10-29
tax_range_time row names do not match taxon_id


LewisAJones:
Currently, when using `tax_range_time`, row names are allocated based on alphabetical ordering of taxon. When using any `by` value other than "name" this is meaningless and should probably be tidied up before returning the dataframe. Row names should probably be the same as taxon_id.

```
# Load library
library(palaeoverse)
# Remove NAs
tetrapods <- subset(tetrapods, !is.na(order) & order != "NO_ORDER_SPECIFIED")
# Temporal range
tax_range_time(occdf = tetrapods, name = "order", plot = TRUE)

                 taxon taxon_id max_ma min_ma range_myr n_occ
30      Trilophosauria        1 251.20 247.20      4.00     1
7      Eosauropterygia        2 251.30 247.20      4.10     8
9          Hupehsuchia        3 251.30 247.20      4.10     4
15         Nothosauria        4 251.30 247.20      4.10     1
25          Testudines        5 251.30 247.20      4.10     3
```

willgearty:
Should we preserve the row names if they were set by the user before providing it to `tax_range_time()`? Looks like we can check if the row names have been set using the following base R code:
`!is.na(.row_names_info(.data, 0L)[[1L]]`

Maybe we only reset the row names if that is `TRUE`?

willgearty:
Oh you're right, never mind.

LewisAJones:
As an introduction to GitHub and software contribution, @CecilyNicholl and @etiennebacher will work on these changes together over the next week or two

etiennebacher:
I just noticed that this is no longer the case on R 4.6.

* R 4.5.3:

```r
> # Load library
> library(palaeoverse)
> # Remove NAs
> tetrapods <- subset(tetrapods, !is.na(order) & order != "NO_ORDER_SPECIFIED")
> # Temporal range
> head(tax_range_time(occdf = tetrapods, name = "order"))
             taxon taxon_id max_ma min_ma range_myr n_occ
30  Trilophosauria        1  251.2  247.2       4.0     1
7  Eosauropterygia        2  251.3  247.2       4.1     8
9      Hupehsuchia        3  251.3  247.2       4.1     4
15     Nothosauria        4  251.3  247.2       4.1     1
25      Testudines        5  251.3  247.2       4.1     3
26  Thalattosauria        6  251.3  247.2       4.1     3
```

* R 4.6.0:

```r
> # Load library
> library(palaeoverse)
[...]

PR #207, opened 2026-06-03
Test that `tax_range_time()` resets row names

CecilyNicholl:
Add a test for #132
- [ ] I have read palaeoverse's [standards and structure](https://palaeoverse.palaeoverse.org/articles/structure-and-standards.html)
- [ ] I have performed a self-review of my code.

### DIFF
--- a/tests/testthat/test-tax_range_time.R
+++ b/tests/testthat/test-tax_range_time.R
@@ -47,5 +47,5 @@ test_that("tax_range_time() resets row names", {
   tetrapods <- subset(tetrapods, !is.na(order) & order != "NO_ORDER_SPECIFIED")
   # Temporal range
   result <- tax_range_time(occdf = tetrapods, name = "order")
-  expect_equal(row.names(result), as.character(1:30))
+  expect_equal(row.names(result), as.character(seq_len(nrow(result))))
 })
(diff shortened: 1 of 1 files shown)


etiennebacher (approved):

tests/testthat/test-tax_range_time.R
```diff
@@ -42,3 +42,10 @@ test_that("tax_range_time() works", {
   occdf$max_ma[1] <- "test"
   expect_snapshot(tax_range_time(occdf = occdf), error = TRUE)
 })
+test_that("tax_range_time() resets row names", {
+  # Remove NAs
+  tetrapods <- subset(tetrapods, !is.na(order) & order != "NO_ORDER_SPECIFIED")
+  # Temporal range
+  result <- tax_range_time(occdf = tetrapods, name = "order")
+  expect_equal(row.names(result), as.character(1:30))
```
willgearty:
```suggestion
  expect_equal(row.names(result), as.character(seq_len(nrow(result))))
```
This might be more robust in case the data ever changes.

LewisAJones (approved):
Wonderful!

tests/testthat/test-tax_range_time.R
LewisAJones: I'm not sure I agree... isn't the point with tests to catch such changes?

willgearty: But that is not the the point of **this** test. This test should, in my opinion, still succeed as long as the column names are sequential, regardless of the output size.

LewisAJones:
This is a test (sorry!)

@all-contributors please add @CecilyNicholl for tests

tests/testthat/test-tax_range_time.R
LewisAJones: If that's all we're intending to test, then I agree 👍🏻

willgearty (approved):

PR #209, opened 2026-06-04
Add RStudio files to `.Rbuildignore` and `.gitignore`

etiennebacher:
I've just opened the project with the latest RStudio and it automatically made these changes.


willgearty:
This is actually a bug with the previous version of RStudio: https://github.com/rstudio/rstudio/issues/17407. No longer added in the latest version of RStudio: https://docs.posit.co/ide/news/

etiennebacher:
Ah okay so the fact that it is automatically added is the bug, but those changes would still be needed if some people activate Posit AI, right?

willgearty:
I think setting a global personal ignore file might be more appropriate for those people: https://stackoverflow.com/a/7335487. 

That being said, I suppose it doesn't hurt to include them in our .gitignore.

etiennebacher:
> I think setting a global personal ignore file might be more appropriate for those people: [stackoverflow.com/a/7335487](https://stackoverflow.com/a/7335487).

I agree in principle, it's just that for new contributors this is another thing to learn and git is already hard enough :smile:

LewisAJones:
Do we need to consider this in respect to our policy on AI (i.e. acknowledging AI usage or not)?

etiennebacher:
I don't think so. If someone wants to hide they are using AI, they can just not commit the AI-related files. This PR simply avoids polluting the git status.

LewisAJones:
True, well, I'm happy with these being left in at this stage

LewisAJones (approved):

willgearty:
Are there other standard AI directories/files we should add if the goal is to make sure they don't get accidentally committed by new contributors?

etiennebacher:
I don't know other files like this and I didn't see any in various tidyverse or r-lib repos where they use Claude. I guess we can exclude more files or folders as we encounter them.

willgearty (approved):

PR #210, opened 2026-06-04
Run CI workflows only if changes are relevant to them

### LINKED ISSUE #188: Skip R-CMD-check when package files are not changed
Changes to files that aren't bundled in the R package shouldn't trigger the GHA workflow. We should be able to edit the GHA workflow to only run when there are changes to the actual code (or other relevant files).

Although now that I'm thinking about it, this might be a fairly small list of exclusion files.

etiennebacher:
This will skip workflows that are irrelevant to the change being made (e.g. there's no point running R CMD check if a PR only changes the CI workflow `lint.yaml`).

Fixes #188


willgearty (approved):
Looks good, thanks!

.github/workflows/R-CMD-check.yaml
```diff
@@ -3,8 +3,28 @@
 on:
   push:
     branches: [main, master]
+    paths:
+      - .github/workflows/R-CMD-check.yaml
```
LewisAJones, line 7: Should this list (and some others) also include `data`? e.g. if somebody changed the underlying data, this could result in examples/tests failing?

LewisAJones (approved):
Looks good! Just one comment for consideration.

.github/workflows/R-CMD-check.yaml
etiennebacher: Yes I suppose that would affect tests and examples. I've made the change, thanks!

Issue #163, opened 2026-05-11
Use `rlang` to check inputs


etiennebacher:
Using `rlang` would allow us to use a range of predefined check functions with nice error messages (see examples below). Error messages will be handled by `cli` (cf #157).

`rlang` doesn't have dependencies and is already imported by `lifecycle` anyway so this wouldn't be an additional dependency.

**Examples** 

Those are simplified cases but adapted from existing code in `palaeoverse`.

1. Basic checks: `rlang` tells you what you provided when it's wrong

``` r
f <- function(x) {
  if (!is.data.frame(x)) {
    stop("`x` should be a dataframe.")
  }
}

f_rlang <- function(x) {
  rlang::check_data_frame(x)
}

obj <- list(1)
f(obj)
#> Error in `f()`:
#> ! `x` should be a dataframe.
f_rlang(obj)
#> Error in `f_rlang()`:
#> ! `x` must be a data frame, not a list.
```

2. Ensure a number is integerish

``` r
# input should be an integer 

is.numeric(1.5) # passes but shouldn't, it's not an integer
#> [1] TRUE
is.integer(1.5) # false, as expected
#> [1] FALSE
is.integer(1) # false, meaning users must pass 1L
#> [1] FALSE

rlang::check_number_whole(1.5, call = NULL)
#> Error:
#> ! `1.5` must be a whole number, not the number 1.5.
rlang::check_number_whole(1, call = NULL)
```

LewisAJones:
This sounds like a good improvement to me!

LewisAJones:
Coming back around to this as some PRs are in draft, I think I'm pretty much in support of this transition. 

The only reservation I might have that I'd like to discuss is whether there is value in having our own internal function that checks arguments? As we are aiming to standardise function arguments and likely have an internal dictionary this could be quite a nice feature and might do a lot of the heavy lifting for future contributors to the package. We have something along these lines in `rmacrostrat` ([see here](https://github.com/palaeoverse/rmacrostrat/blob/efeeed81baaca33b221b70079cb4abba73b2a89b/R/utils.R#L29)), and it works really well. **Perhaps this doesn't need to be one or the other though, and rlang can be used in the internal function.**

@willgearty do you have any thoughts before we commit to using rlang?

willgearty:
I use rlang for deeptime, and it works great. Fully in support of this.

etiennebacher:
- [x] axis_geo (#338)
[...]

PR #217, opened 2026-06-08
Rework tests for `lat_bins_area()`

etiennebacher:
This PR revamps tests for `lat_bins_area()`. It:

* uses snapshot tests to check output

Some of the new snapshots show "bad" error messages, those will be improved once `rlang` is used in the package (#163). 

Part of #161 


tests/testthat/_snaps/lat_bins_area.md
```diff
@@ -1,11 +1,61 @@
...
+    Code
+      lat_bins_area(n = 6)
+    Output
+        bin       min       mid       max         area area_prop
+      1   1  41.81031  65.90516  90.00000 8.501075e+13 0.1666667
+      2   2  19.47122  30.64077  41.81031 8.501075e+13 0.1666667
+      3   3   0.00000   9.73561  19.47122 8.501075e+13 0.1666667
+      4   4 -19.47122  -9.73561   0.00000 8.501075e+13 0.1666667
+      5   5 -41.81031 -30.64077 -19.47122 8.501075e+13 0.1666667
+      6   6 -90.00000 -65.90516 -41.81031 8.501075e+13 0.1666667
```
etiennebacher, line 31: Using a snapshot is a gold standard here: we can check the number of rows, columns, column names, and ensure that we know if values change for some reason in the future. Worth noting that snapshot tests are slightly slower to run, but I don't think it's a big problem here.

We could have an `expect_equal()` for this but it would just be very annoying to type all the decimals.

tests/testthat/_snaps/lat_bins_area.md
```diff
@@ -1,11 +1,61 @@
...
       Error in `lat_bins_area()`:
       ! `n` should be a numeric.
 
+---
+
+    Code
+      lat_bins_area(n = -1)
+    Condition
+      Error in `data.frame()`:
+      ! arguments imply differing number of rows: 3, 1, 0
```
etiennebacher, line 60: This (and a few other error messages) will be improved by `rlang` in the refactor.

tests/testthat/test-lat_bins_area.R
```diff
@@ -1,20 +1,44 @@
...
+test_that("different radius affects the results", {
+  orig <- lat_bins_area(n = 6)
+  new <- lat_bins_area(n = 6, r = 6000)
+  expect_all_false(orig$area == new$area)
+})
+
+# TODO: what should happen here? Currently it returns area = 0 and area_prop = NaN
+# test_that("lat_bins_aread handles case where min == max", {
+#   lat_bins_area(min = 90, max = 90)
+# })
```
etiennebacher: I guess this should be an error (i.e. the the `max >= min` condition should actually be `max > min`)? What do you think?

``` r
palaeoverse::lat_bins_area(3, min = 90, max = 90)
#>   bin min mid max area area_prop
#> 1   1  90  90  90    0       NaN
#> 2   2  90  90  90    0       NaN
#> 3   3  90  90  90    0       NaN
```

tests/testthat/test-lat_bins_area.R
```diff
@@ -1,20 +1,44 @@
...
+
   expect_snapshot(lat_bins_area(plot = "TRUE"), error = TRUE)
+  expect_snapshot(lat_bins_area(plot = logical(0)), error = TRUE)
+
   expect_snapshot(lat_bins_area(r = "Earth"), error = TRUE)
-  expect_snapshot(lat_bins_area(n = 3.5), error = TRUE)
+  expect_snapshot(lat_bins_area(r = numeric(0)), error = TRUE)
+
+  # TODO: I suppose this should error? Currently it doesn't
+  # expect_snapshot(lat_bins_area(r = -1), error = TRUE)
```
etiennebacher: Can you confirm this should error?

``` r
palaeoverse::lat_bins_area(3, r = -1)
#>   bin       min           mid       max    area area_prop
#> 1   1  19.47122  5.473561e+01  90.00000 4188790 0.3333333
#> 2   2 -19.47122 -3.552714e-15  19.47122 4188790 0.3333333
#> 3   3 -90.00000 -5.473561e+01 -19.47122 4188790 0.3333333
```

tests/testthat/test-lat_bins_area.R
etiennebacher: Note that this errors in `lat_bins_degrees()`:

``` r
palaeoverse::lat_bins_degrees(min = 30, max = 30)
#> Error in `palaeoverse::lat_bins_degrees()`:
#> ! `min` should be less than `max`
```

LewisAJones: Yes, this should be an error, the output isn't exactly useful!

tests/testthat/test-lat_bins_area.R
LewisAJones: Yeah, this is nonsensical (`r` = the radius of the Earth in km) but I am not sure if and/or when we should error. I think we should still keep `r` as a 'free' input value. Erroring when less than or equal to 0 probably makes a lo of sense, and perhaps adding a warning if less than/greater than standard values used for the radius?

LewisAJones (approved):
Good job on picking up some edge cases/bugs. The reworked tests look good to me! We will want to revisit some of this during the refactoring, but for now, this looks great.

tests/testthat/test-lat_bins_area.R
willgearty: I think for now we should error when equal to or less than 0. Implementing better warnings sounds like a good small thing for someone to work on in the future.

willgearty (approved):
Looks good, thanks! (coverage tests failing here as well?)

etiennebacher:
I have checked the test coverage output and also followed the steps detailed in the CI logs (download the Github artifact and use the snapshots contained in it) and can't find any problem. Lewis found another issue when mixing test coverage and snapshots (https://github.com/r-lib/devtools/issues/2698) so I think this is just a false positive.

Thanks both for your review!

PR #218, opened 2026-06-08
Rework tests for `bin_time()`

etiennebacher:
This PR revamps tests for `bin_time()`. It:

* uses snapshot tests to check output

Part of #161 

Coverage workflow failing because of https://github.com/palaeoverse/palaeoverse/issues/235.


tests/testthat/test-bin_time.R
```diff
@@ -1,61 +1,356 @@
...
+  expect_type(bin_mid$bin_assignment, "integer")
+  expect_type(bin_mid$bin_midpoint, "double")
+
+  # Test the actual values
+  expect_snapshot(bin_mid[, c(
+    "id",
+    "n_bins",
+    "bin_assignment",
+    "bin_midpoint"
+  )])
```
etiennebacher: This structure tests that:

- column names are correct
- types of new columns are correct
- values of the new columns are correct (I haven't checked the actual values, the snapshot just stores the current output)

The input is only the first 5 rows of `tetrapods` because I don't think we want to test all values of `tetrapods`, but let me know if you'd rather approach this differently. 

I follow the same structure for the other values of `method`.

tests/testthat/test-bin_time.R
```diff
@@ -1,61 +1,356 @@
...
+      expect_type(x$point_estimates, "double")
+    })
+  )
+
+  expect_snapshot(bin_point[[1]][, c(
+    "id",
+    "n_bins",
+    "bin_assignment",
+    "point_estimates"
+  )])
```
etiennebacher: This returns a long list of dataframes so I only snapshot the values for the first one, let me know if you're fine with that.

tests/testthat/test-bin_time.R
```diff
@@ -1,61 +1,356 @@
...
+      expect_named(
+        x,
+        c(names(occdf), "id", "n_bins", "bin_assignment", "point_estimates")
+      )
+      expect_type(x$id, "integer")
+      expect_type(x$n_bins, "integer")
+      expect_type(x$bin_assignment, "integer")
+      expect_type(x$point_estimates, "double")
+    })
+  )
```
etiennebacher: Looks a bit weird, but I didn't find a better way to test the shape and column names and types in all dataframes of the list.

tests/testthat/test-bin_time.R
```diff
@@ -89,94 +386,55 @@ test_that("bin_time() works", {
...
+#     bin = 1:54,
+#     max_ma = seq(10, 540, 10),
+#     min_ma = seq(0, 530, 10)
+#   )
+
+#   expect_snapshot(
+#     bin_time(occdf = occdf, bins = bins, method = "random", fun = dnorm),
+#     error = TRUE
+#   )
+# })
```
etiennebacher, line 497: IMO we should error if `fun` is provided but isn't relevant to the `method` specified, what do you think?

tests/testthat/test-bin_time.R
LewisAJones: Sounds good, I think checking all values might be overkill. However, there is a risk here of not catching all potential behaviour in relation to the age range of the occurrence. For example, whether the occurrence's age range falls completely within the time bin, overlaps two bins, multiple bins, etc. It's also not guaranteed that we capture this with the `tetrapods` dataset. We may want to then think about all possible cases and use a dummy dataset?

tests/testthat/test-bin_time.R
LewisAJones: I think this is OK but maybe add a short comment for a would-be contributor?

tests/testthat/test-bin_time.R
LewisAJones: I'm fine with that at this stage. We will likely revisit all of this as we refactor the function.

tests/testthat/test-bin_time.R
```diff
@@ -89,94 +386,55 @@ test_that("bin_time() works", {
...
+
+# TODO: shouldn't this error?
```
LewisAJones, line 499: We have a default value set for this which is why we don't have it set to error.

Is your suggestion to remove this default value, and then error if it is supplied and 'point' or 'random' are not the method input?

tests/testthat/test-bin_time.R
LewisAJones: I can see the value in that. A lot of these arguments have default values, which is why that isn't the case. Removing these would make researchers potentially think more carefully about what the input should be.

LewisAJones (changes requested):
Looks good, just a few comments, mostly in response to your queries!

willgearty (approved):
No comments from me beyond what Lewis has already said.

tests/testthat/test-bin_time.R
etiennebacher: I'd probably try to follow ["Extract strategies into objects"](https://design.tidyverse.org/strategy-objects.html) from the Tidyverse design. Basically we would have five small functions, each corresponding to a strategy:

```r
bin_time <- function(
  occdf,
  ...,
  bins,
  min_ma = "min_ma",
  max_ma = "max_ma",
  method = "mid"
) { }

bin_meth_mid <- function() { }

bin_meth_majority <- function() { }

bin_meth_all <- function() { }

bin_meth_random <- function(..., reps) { }

bin_meth_point <- function(..., reps, fun) { }
```

and they would be used like this:

```r
bin_time(occdf = occdf, bins = bins, method = bin_meth_all())
bin_time(occdf = occdf, bins = bins, method = bin_meth_random(reps = 50))
bin_time(occdf = occdf, bins = bins, method = bin_meth_point(fun = dnorm, mean = 0.5, sd = 0.25))
```

This would have several advantages:

1. simplify the definition of `bin_time()` because we remove the arguments `reps` and `fun`, and the `...` would stay but they would have to be empty (they'd only be here to force other arguments to be named)

1. it makes the intent of the function clearer. It's easy to read this without realizing that `fun` doesn't do anything because it only applies when `method = "point"`:
   ```r
   bin_time(occdf = occdf, bins = bins, method = "random", fun = dnorm)
   ```
   This mistake would be impossible with this:
   ```r
   bin_time(occdf = occdf, bins = bins, method = bin_meth_random(fun = dnorm))
   ```
   because `bin_meth_random()` doesn't have a `fun` argument so it would just error.

1. You can easily  see what arguments a method accepts thanks to autocomplete. If you type `bin_meth_point()` then `<TAB>` will suggest `reps` and `fun`.

1. Adding another method that would have extra arguments will be easier because it wouldn't require changing `bin_time()`, it would just require a new `bin_meth_*()`.

---

I realize now that this is a big answer :smile: I think at the minimum I would error if some arguments are incompatible or don't make sense together, e.g. `method = "random"` and `fun = dnorm`.

tests/testthat/test-bin_time.R
etiennebacher: [af64d2d](https://github.com/palaeoverse/palaeoverse/pull/218/commits/af64d2de4492a6cfab56d692588f65e66258d4b7)

tests/testthat/test-bin_time.R
willgearty: I like this strategy

etiennebacher: I have created a new issue for that (https://github.com/palaeoverse/palaeoverse/issues/237) so that this doesn't block the PR.

etiennebacher:
Following @LewisAJones's comments above about edge cases, I have made large changes in the test data used. Instead of using `tetrapods`, we now have two small dataframes (`test_bins` and `test_occdf`) that contain less values but cover a lot of possible edge cases and are easier to test. I'm re-requesting a review because there were many changes after the first review.

---

Claude was used to generate the basic dataframe to test but I've done some changes and written and reviewed all tests by hand.

willgearty (approved):

tests/testthat/test-bin_time.R
```diff
@@ -1,61 +1,455 @@
...
+
+test_that("bin_time() works with method 'mid'", {
+  # occ7's midpoint equals a bin midpoint, which triggers a warning
+  expect_warning(
+    bin_mid <- bin_time(occdf = test_occdf, bins = test_bins, method = "mid"),
+    "equivalent to a bin boundary"
+  )
+  # occ3's midpoint (10) falls on a bin boundary, so it is NA
+  expect_equal(
+    bin_mid[, c("name", "id", "n_bins", "bin_assignment", "bin_midpoint")],
```
LewisAJones: Is there a reason why we are excluding some columns here? Wouldn't we also want to ensure that "min_ma" and "max_ma" are not being modified in any way?

tests/testthat/test-bin_time.R
```diff
@@ -1,61 +1,455 @@
...
+    data.frame(
+      name = paste0("occ", 1:8),
+      id = 1:8,
+      n_bins = c(1, 1, 2, 2, 3, 5, 1, 1),
+      bin_assignment = c(1, 2, NA, 2, 2, 3, 1, 2),
+      bin_midpoint = c(5, 15, NA, 15, 15, 25, 5, 15)
+    )
+  )
+  # TODO: regression test for https://github.com/palaeoverse/palaeoverse/issues/236
```
LewisAJones, line 74: 😬

tests/testthat/test-bin_time.R
```diff
@@ -1,61 +1,455 @@
...
+  # )
+})
+
+test_that("bin_time() works with method 'majority'", {
+  bin_majority <- bin_time(
+    occdf = test_occdf,
+    bins = test_bins,
+    method = "majority"
+  )
+  # occ3 is a 50/50 tie, resolved to the first bin (bin 1)
```
LewisAJones, line 86: Not for the future, should this also result in a warning?

tests/testthat/test-bin_time.R
```diff
@@ -1,61 +1,455 @@
...
+  expect_equal(
+    bin_majority[, c(
+      "name",
+      "id",
+      "n_bins",
+      "bin_assignment",
+      "bin_midpoint",
+      "overlap_percentage"
+    )],
+    data.frame(
```
LewisAJones, line 89: As above, here and elsewhere, should we not just test all values?

tests/testthat/test-bin_time.R
LewisAJones: I like the sound of this, some details to discuss, but this will be part of the refactor (we will want to consider this for other functions too e.g. `tax_range_space`).

LewisAJones (changes requested):

tests/testthat/test-bin_time.R
etiennebacher: I think it would make sense since we already warn for other things, like midpoints that equal bin boundaries

tests/testthat/test-bin_time.R
etiennebacher: No good reason, I think I just wanted to shorten the output, I fixed this.

tests/testthat/test-bin_time.R
etiennebacher: https://github.com/palaeoverse/palaeoverse/issues/255

LewisAJones (approved):

PR #219, opened 2026-06-08
Rework tests for `lat_bins_degrees()`

etiennebacher:
This PR revamps tests for `lat_bins_degrees`(). It:

 - splits tests into more test_that() code blocks
 - adds more tests with wrong inputs
 - adds tests with some hardcoded results and some snapshots
 - adds visual tests when `plot = TRUE`

Part of #161


tests/testthat/test-lat_bins_degrees.R
```diff
@@ -1,16 +1,96 @@
...
+
+  # TODO: this looks suspicious, "min", "mid", and "max" should be column names?
+  # expect_equal(
+  #   lat_bins_degrees(40, min = 0, max = 10),
+  #   data.frame(
+  #     bin = c(1, 1, 1),
+  #     df = c(0, 20, 40),
+  #     row.names = c("min", "mid", "max")
+  #   )
+  # )
```
etiennebacher, line 48: The output structure looks weird if `min` and `max` are close but I don't know enough yet to determine if this is a bug:

``` r
library(palaeoverse)
lat_bins_degrees(40, min = 0, max = 10)
#>     bin df
#> min   1  0
#> mid   1 20
#> max   1 40
lat_bins_degrees(40, min = 0, max = 80)
#>   bin min mid max
#> 1   1  40  60  80
#> 2   2   0  20  40
```

Is this expected?

LewisAJones: Looks like a bug to me! I guess we didn't expect users to input a `size` larger than the range (`max` - `min`).

LewisAJones (approved):
Again, good job on picking up some edge cases/bugs 😄 

The reworked tests look good. We will want to revisit some of this again during the refactoring but this sets us up well for now.

tests/testthat/test-lat_bins_degrees.R
willgearty: Looks like this happens if you only get one bin (`size` smaller than the range but larger than half of the range also causes the bug).

willgearty (approved):
Looks good to me. Make sure to report the bug(s) as issues before merging.

The tests are running properly for me locally (and in the various R-cmd-check workflows), but it looks like they are failing exclusively for the codecov workflow? Thoughts?

tests/testthat/test-lat_bins_degrees.R
etiennebacher: https://github.com/palaeoverse/palaeoverse/issues/233

etiennebacher:
Thanks for the review! I have opened #233 for the bug.

The failure in the coverage workflow is annoying but I don't think there's much we can do, cf https://github.com/palaeoverse/palaeoverse/pull/217#issuecomment-4766611594.

PR #224, opened 2026-06-09
Cache data for vignettes so that they are always rendered

etiennebacher:
This PR basically removes the current infrastructure which consists in evaluating vignettes in `vignettes/source` from time to time and use their (static) output as the official vignettes used by R and `pkgdown`. This approach meant that vignettes were never rendered in CI, which could lead to bugs being undetected, e.g. #186.

I suppose the reason behind this approach was that `palaerotate()` calls are 1) time-consuming and 2) rely on `gplates.org` being available, which is not always the case, especially in CI. 

The new approach stores results of those `palaeorotate()` calls in `vignettes/_data/*.rds` files so that we can quickly read those files when rendering the vignettes. We still show the  `palaeorotate()` calls in the vignette, but in the background we run `readRDS()`. However, this is not perfect because if results from `gplates.org` change in the future, we won't automatically capture them. We could make a workflow run once a week or once a month to update those `.rds` files.

This PR contains a lot of changes but the large majority of those come from copying the content of  `vignettes/_source/tetrapod-biodiversity.Rmd` into `vignettes/tetrapod-biodiversity.Rmd` (and idem for the other vignette). In summary:

- chunks that contained rendered results are gone
- I removed `devtools::load_all()` and evaluate `library(palaeoverse)` instead (the `pkgdown` workflow installs the package in CI so it always uses the development version)
- chunks containing `palaeorotate()` now have `eval = FALSE` and are followed by a chunk to read the corresponding RDS file (this chunk is hidden)

I have compared the output of `pkgdown::build_articles()` to the currently deployed website and I didn't see any differences.
 

LewisAJones (commented):
I think this is a really good idea @etiennebacher, thanks for putting it together.

> I suppose the reason behind this approach was that palaerotate() calls are 1) time-consuming and 2) rely on gplates.org being available, which is not always the case, especially in CI.

Bang on!

My main concern (and one you've highlighted) is keeping these data files up to date. That said, these files should *in theory* never change as the returned values from the API should always be the same for the same input (age, coordinates, model). However, based on past experience, this has not always been the case. So, probably some workflow might be useful here? Let's discuss this a little. I'm keen to hear @willgearty's thoughts on this too.

My other concern with this approach is for individuals who would actually open and use the `.rmd` file as a tutorial. However, based on conversation elsewhere relating to this with `.qmd` files, it sounds like nobody really does this (apart from me :laughing), so this is probably fine.

Articles look good after being built to me (`pkgdown::build_articles()`).

I'd like to sit on this a little longer to think if there is any reason why we shouldn't do this (I'll revisit next week), but I suspect this is essentially good to go.

As a separate note, we will want to review these vignettes again at some point. I am sure there are improvements to be made and things to be updated (e.g. links to Google Group). Maybe a good task for someone on the team who wants to gain a bit more familiarity with using GitHub but also knows Palaeoverse.

willgearty (changes requested):
I like this strategy. I agree with Lewis that the palaeorotate results shouldn't change that frequently, but I think we want to make it easier to check for updates (I think once before each release seems reasonable). Could you add a script to do this? And then update the item in our [release checklist](https://github.com/palaeoverse/palaeoverse/blob/main/R/release.R)?

etiennebacher:
Thanks both!

> My main concern (and one you've highlighted) is keeping these data files up to date. That said, these files should in theory never change as the returned values from the API should always be the same for the same input (age, coordinates, model). However, based on past experience, this has not always been the case. So, probably some workflow might be useful here?

> I agree with Lewis that the palaeorotate results shouldn't change that frequently, but I think we want to make it easier to check for updates (I think once before each release seems reasonable). Could you add a script to do this? And then update the item in our [release checklist](https://github.com/palaeoverse/palaeoverse/blob/main/R/release.R)?

That's a good idea. I have added two scripts to run the original code from the vignettes until the `palaeorotate()` calls. This compares the output we get to the stored output and errors if there are any differences. These scripts also run in CI (I put a cron job to run them weekly but no problem to change that if needed) and I have modified the bullet point for the checklist in `R/release.R`. There are two points I would like to discuss:

- these scripts might hit some connectivity issues (especially in CI). We don't want the workflow to fail in that case because there's nothing we can do about it. At the same time, this might lead us to believe that the RDS files are always up-to-date while the reality is that the scripts always fail to connect to gplates.org. I don't really know how to reconcile the fact that we want to be aware of that without being spammed.
- these scripts re-use the same code as in the vignettes (except code that is irrelevant, like `head()` calls to show the data). This means we need to keep the code in vignettes and the code in scripts in sync. Should we have yet another workflow for that, similar to the one detecting changes in `DESCRIPTION` and `CITATION.cff`?

---

Claude Opus 4.8 helped to create the helper script and the workflow

willgearty:
I haven't fully thought through this idea, but what about this: instead of having a second copy of the script, what if we had a non-echo'd block in the vignettes that saved the RDS file if it doesn't exist. Then we have a script that deletes the RDS file, then runs the vignette, which should regenerate the RDS file. Git should then be able to identify if the RDS file has changed. Same process could be used for a GHA workflow.

etiennebacher:
I like this approach, it adds a bit of complexity in the vignettes but it removes the burden of duplicated code.

vignettes/tetrapod-biodiversity.Rmd
```diff
@@ -252,73 +164,57 @@ We'd recommend playing around with these different options and seeing how they i
...
 ```
 
-Now let's check that it's worked...
+```{r echo = FALSE, eval = !rds_point_paleomap_tetrapods_exists}
+saveRDS(maj_tetrapods, "_data/palaeorotate_point_paleomap_tetrapods.rds")
+```
+
+```{r echo = FALSE, eval = rds_point_paleomap_tetrapods_exists, warning = FALSE, message = FALSE}
+maj_tetrapods <- readRDS("_data/palaeorotate_point_paleomap_tetrapods.rds")
+```
```
etiennebacher, line 189: In the new approach suggested by Will, we first check the RDS existence:

- doesn't exist (because we manually deleted it) --> evaluate `palaeorotate()` and save the output to the RDS file, but skip the `readRDS()`
- already exists --> skip `palaeorotate()` and `saveRDS()`, only evaluate `readRDS()`

LewisAJones: I think this works well, but worth adding some comments to explain why we do this to potential contributors?

LewisAJones (commented):
Thanks for putting this together @etiennebacher!

I am not convinced that we need a workflow (`check-rds.yaml`) for periodically checking whether the RDS files are up to date. The results of these calls to the GPlates API should stay the same unless they realise something drastic is not quite right on their end (e.g. used wrong model). We will observe that these changes have taken place through `release.R` and I don't see much added benefit of knowing about these changes immediately (within one week). That said, I also don't think it does any harm.

Do we need to be mindful of the `.RDS` file size here, will this contribute to CRAN's hard limits?

etiennebacher:
> Do we need to be mindful of the .RDS file size here, will this contribute to CRAN's hard limits?

The limit is 5MB and `R CMD build` currently generates a 3.4MB tarball so this is not an issue.

> The results of these calls to the GPlates API should stay the same unless they realise something drastic is not quite right on their end (e.g. used wrong model). We will observe that these changes have taken place through release.R and I don't see much added benefit of knowing about these changes immediately (within one week)

On one hand, if there are drastic changes it's probably better to learn about them before being about to release a new version. On the other hand, those changes should be captured by R CMD check which runs weekly anyway. I'm fine with dropping the workflow to reduce the project complexity, unless @willgearty wants to keep it.

vignettes/tetrapod-biodiversity.Rmd
etiennebacher: Done in [475eaaa](https://github.com/palaeoverse/palaeoverse/pull/224/commits/475eaaa8f71648efc2991563103c17dd49d728c4), let me know if you want to modify this comment

willgearty:
Glad to hear my half-baked idea worked out! I agree with Lewis that just recompiling the rds files for a new release of the package should be sufficient.

willgearty (approved):

LewisAJones:
> The limit is 5MB and `R CMD build` currently generates a 3.4MB tarball so this is not an issue.

Sounds good!👍🏻 This might be something we want to keep an eye on though, especially if we add any additional example datasets in the future.

LewisAJones (approved):

PR #225, opened 2026-06-10
Rework tests for `group_apply()`

etiennebacher:
This PR revamps tests for `bin_time()`. It:

Part of #161 

Coverage workflow failing because of https://github.com/palaeoverse/palaeoverse/issues/235.


tests/testthat/test-group_apply.R
```diff
@@ -1,57 +1,91 @@
...
   expect_equal(
-    nrow(
-      group_apply(occdf = occdf, group = c("cc"), fun = nrow)
+    group_apply(occdf = occdf, group = "cc", fun = nrow),
+    data.frame(
+      nrow = c(1, 5, 6, 35, 3),
+      cc = c("CA", "RU", "UK", "US", "ZA")
     ),
-    37
+    # TODO: this is necessary because the first column is a named vector but
```
LewisAJones, line 10: 👍🏻

tests/testthat/test-group_apply.R
```diff
@@ -1,57 +1,91 @@
...
-    1248
+    group_apply(occdf = occdf, group = c("collection_no", "cc"), fun = nrow),
+    data.frame(
+      nrow = c(1, 5, 4, 2, 1, 1, 1, 1, 1, 9, 6, 6, 3, 6, 1, 1, 1),
+      collection_no = c(
+        "13219", "22644", "22725", "22726", "12943", "13044", "13046", "13048", "13049", 
+        "13080", "13257", "13947", "22635", "22714", "13004", "13043", "13083"
+      ),
+      cc = c("CA", "RU", "UK", "UK", rep("US", 10), "ZA", "ZA", "ZA"),
+      row.names = 1:17
```
LewisAJones: Any reason why we're explicitly defining row names here but not elsewhere?

tests/testthat/test-group_apply.R
```diff
@@ -1,57 +1,91 @@
...
+    group_apply(occdf = 1, group = "cc", fun = nrow),
+    error = TRUE
+  )
+  expect_snapshot(
+    group_apply(occdf = data.frame(), group = "cc", fun = nrow),
+    error = TRUE
+  )
+})
+
+test_that("group_apply() accepts functions that return less or more rows that in the input", {
```
LewisAJones:
```suggestion
test_that("group_apply() accepts functions that return less or more rows than in the input", {
```

tests/testthat/test-group_apply.R
```diff
@@ -1,57 +1,91 @@
...
   )
 
-  expect_null(nrow(group_apply(
-    occdf = occdf,
-    group = c("collection_no"),
-    fun = tax_check,
-    verbose = FALSE
-  )))
+  # can return no rows at all
+  expect_null(
```
LewisAJones, line 76: I've just noticed/remembered that `tax_check` doesn't play nicely with `group_apply` and `tax_check` has it own `group` argument. Presumably because `tax_check` returns a list. This is probably something to be refactored? 

Should this test be changed?

```
# Load occurrence data
data("tetrapods")
# Check taxon names alphabetically (works)
tax_check(taxdf = tetrapods, name = "genus", dis = 0.1)
# Check by country (does not work)
group_apply(occdf = occdf, group = "cc", fun = tax_check, name = "genus", dis = 0.1)
# Check by country (works)
tax_check(taxdf = tetrapods, name = "genus", group = "cc", dis = 0.1)
```

tests/testthat/test-group_apply.R
```diff
@@ -99,53 +133,67 @@ test_that("group_apply() puts groups last", {
...
+    group_apply(occdf = occdf, group = 1, fun = nrow),
     error = TRUE
   )
   expect_snapshot(
-    group_apply(occdf = occdf, fun = tax_range_time),
+    group_apply(occdf = occdf, group = c("cc", "foobar"), fun = nrow),
+    error = TRUE
+  )
+
+  # using an external object in 'group' when 'group' has multiple values
```
LewisAJones: Test is fine, but we probably want to remove this behaviour in the future?

LewisAJones (changes requested):
Thanks @etiennebacher, looks good! Just a few small comments, perhaps some for discussion.

tests/testthat/test-group_apply.R
etiennebacher: I think the refactor will lead to giving a different error message here (currently it errors with "cannot xtfrm data frames"). I don't understand what you mean by "removing this behaviour", we don't want to make this work, right?

tests/testthat/test-group_apply.R
etiennebacher: No particular reason, I just use [constructive](https://cynkra.github.io/constructive/) to generate the code that would produce the expected output (if you've already used `dput()` before, it's similar but produces better code). It automatically adds row names, but I'll remove them.

tests/testthat/test-group_apply.R
etiennebacher: I think there are two different points here:

1. the objective of this test is to check whether `group_apply()` is allowed to return `NULL` (we might also force it to return a 0-row dataframe, but then what columns should it contain?)
2. how should `group_apply()` and `tax_check()` interact? I think this is a broader question, I can open an issue for that if you want.

tests/testthat/test-group_apply.R
willgearty: Given the test passes, it seems this currently errors properly. But maybe we should have better error reporting to the user if one or more of the group identifiers doesn't correspond to columns in occdf?

willgearty (approved):
No additional comments from me.

tests/testthat/test-group_apply.R
etiennebacher: Yes I agree

tests/testthat/test-group_apply.R
LewisAJones: I had not seen [constructive](https://cynkra.github.io/constructive/) before, thanks for sharing! 👀 

I'm ultimately fine with this either way, it was just for my own clarity/understanding more than anything!

tests/testthat/test-group_apply.R
LewisAJones: > 1. the objective of this test is to check whether `group_apply()` is allowed to return `NULL` (we might also force it to return a 0-row dataframe, but then what columns should it contain?)

Fair, I'm happy with this for now, especially as it is the current behaviour of the function! 

> 2. how should `group_apply()` and `tax_check()` interact? I think this is a broader question, I can open an issue for that if you want.

Probably is worth opening an issue. I think ideally we would want all/most functions to work similarly with `group_apply()`, or have an internal `group` argument?

etiennebacher: https://github.com/palaeoverse/palaeoverse/issues/262

tests/testthat/test-group_apply.R
LewisAJones: This needs to be confirmed as the behaviour is a little odd, but currently it seems that `group_apply` is considering defined objects (e.g. here `foo`) rather than strictly using the columns available in `occdf` based on the defined `group`. This is the behaviour that should probably change.

etiennebacher: @LewisAJones 986a5caaccec9f1aed490c28186bcf8cc0808c21

etiennebacher: This will actually be an easy fix:

https://github.com/palaeoverse/palaeoverse/blob/c4f8fe9fbfa45b02c2c3b867c431b65b549090f2/R/group_apply.R#L93-L95

This should be `if (!all(group %in% colnames(occdf)))` instead.

LewisAJones (approved):

PR #226, opened 2026-06-10
Make `lychee` workflow more robust to connection errors

etiennebacher:
I often see false positives like this in the `check-links` workflow:

```
Error: R] <https://palaeoverse.org/conduct.html> (at 49:68) | Connection failed. Check network connectivity and firewall settings
```

This PR adds some parameters to try to limit this kind of errors. Most of this was suggested by Claude. I made a few empty commits here and this workflow always passed so the fix seems to work.


willgearty (approved):

LewisAJones (approved):
Looks good! R CMD Checks aren't currently passing because the Macrostrat API is currently down but check-links is so this should be good to go.

PR #238, opened 2026-06-23
Fix coverage workflow

etiennebacher:
This adds a few more input checks to `lat_bins_*()` functions so that `covr` reports the same errors as `devtools::test()` in snapshots. 

Cf https://github.com/palaeoverse/palaeoverse/issues/235#issuecomment-4768858551 for the full explanation.

Fixes #235 

## Checklist before requesting a review
- [x] I have read palaeoverse's [standards and structure](https://palaeoverse.palaeoverse.org/articles/structure-and-standards.html)
- [x] I have performed a self-review of my code.
- [ ] I have added function documentation.
- [ ] I have provided sufficient examples of implementation.
- [ ] If it is a core feature, I have added thorough tests.

Issue #235, opened 2026-06-22
Coverage workflow keeps failing


etiennebacher:
I thought it happened in PRs where snapshots changed but it even fails on new PRs that don't touch tests: https://github.com/palaeoverse/palaeoverse/actions/runs/27953084530/job/82714546830?pr=234

```
══ Failed tests ════════════════════════════════════════════════════════════════
── Failure ('test-lat_bins_area.R:23:3'): lat_bins_area errors with wrong inputs ──
Snapshot of code has changed:
    old                              | new                                
[1] Code                             | Code                            [1]
[2]   lat_bins_area(n = numeric(0))  |   lat_bins_area(n = numeric(0)) [2]
[3] Condition                        | Condition                       [3]
[4]   Error in `if (n%%1 != 0) ...`: -   Error in `if (...) NULL`:     [4]
[5]   ! argument is of length zero   |   ! argument is of length zero  [5]
* Run `testthat::snapshot_accept("lat_bins_area", "testthat")` to accept the change.
* Run `testthat::snapshot_review("lat_bins_area", "testthat")` to review the change.
``` 

I can't figure it out for now but opening this to keep track of it.

etiennebacher:
Claude to the rescue (this is a summary of its explanation):

### Why do snapshot messages change?

When `covr` runs, it doesn't run tests right away but it [instruments the functions in the package](https://covr.r-lib.org/articles/how_it_works.html) first. This means that it inserts pieces of code to count the number of times a line of code is run during the tests. For instance, this piece of code:

```r
 if (n %% 1 != 0) {
    stop("not int")
 }
```
would become this:

```r
 if (if (TRUE) {
      covr:::count("...:2:7:...")
      n %% 1 != 0
  }) {
      if (TRUE) {
          covr:::count("...:3:5:...")
          stop("not int")
      }
  }
```

Therefore, when `n` has length 0 (e.g. `numeric(0)`) then the error:

```r
Error in `if (n%%1 != 0) ...`:   ! argument is of length zero
```

becomes:

```r
Error in `if (...) NULL`:    ! argument is of length zero
```
as we see in the snapshots, because the entire block 

```r
if (TRUE) {
      covr:::count("...:2:7:...")
      n %% 1 != 0
  }
```

has length 0.

### Why doesn't it fail locally?

[...]

PR #239, opened 2026-06-23
Rework tests for `tax_certainty()`

etiennebacher:
This PR revamps tests for `tax_certainty()`. It:

* checks more values

There are many `# TODO` but I don't think we need to discuss them, they all seem pretty clear to me.

### DIFF
--- a/R/tax_certainty.R
+++ b/R/tax_certainty.R
@@ -119,7 +119,7 @@ tax_certainty <- function(
     stop("`terms` must be of class list or NULL.")
   }
   # Check for valid append input
-  if (!is.logical(append)) {
+  if (is.na(append) || !is.logical(append)) {
     stop("`append` must be of class logical (TRUE/FALSE).")
   }
   # Create temporary taxdf column to not replace original values
--- a/tests/testthat/test-tax_certainty.R
+++ b/tests/testthat/test-tax_certainty.R
@@ -1,74 +1,156 @@
-test_that("tax_certainty() works", {
+test_that("tax_certainty() basic behavior", {
   data("tetrapods")
-  # Error handling
-  expect_error(tax_certainty(vector()))
-  expect_error(tax_certainty(NULL))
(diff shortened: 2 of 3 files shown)


tests/testthat/test-tax_certainty.R
```diff
@@ -1,74 +1,156 @@
...
+
+  # check mixing character and numeric
+  expect_equal(
+    tax_certainty(
+      taxdf = occdf,
+      name = "identified_name",
+      certainty = c("A", 0)
+    ),
+    cbind(occdf, data.frame(certainty = c("A", "0", "A", "A", "A")))
+  )
```
etiennebacher, line 101: We could be stricter and throw an error when mixing types in `certainty` to avoid footguns related to automatic type coercion (e.g. `c("A", 0)` -> `c("A", "0")`).  What do you think? 

Edit: I'm actually not sure how to do that since we'll only access `certainty` internally after it has been coerced...

LewisAJones (approved):
Good job on catching additional error cases, looks like there are many to discuss and implement! 

Note, we're hitting the same bug with `test_coverage_active_file()` here as previously discussed on other PRs (nothing we can do about this).

etiennebacher:
> Note, we're hitting the same bug with test_coverage_active_file() here as previously discussed on other PRs (nothing we can do about this).

I actually know how to fix this now, it requires changes similar to #238

R/tax_certainty.R
```diff
@@ -119,7 +119,7 @@ tax_certainty <- function(
     stop("`terms` must be of class list or NULL.")
   }
   # Check for valid append input
-  if (!is.logical(append)) {
+  if (is.na(append) || !is.logical(append)) {
```
etiennebacher, line 122: Needed to prevent `covr` and `devtools` from reporting different snapshots, cf #235 and #238

willgearty (approved):

PR #244, opened 2026-06-30
Rework tests for `phylo_check()`

etiennebacher:
This PR revamps tests for `phylo_check()`. It:

* checks more values


tests/testthat/test-phylo_check.R
```diff
@@ -16,30 +16,158 @@ test_that("phylo_check() works", {
...
+
+test_that("arg 'list' works", {
+  skip_if_not_installed("paleotree")
+  library(paleotree)
+  data(RaiaCopesRule)
+  tree <- ceratopsianTreeRaia
+
+  out <- phylo_check(tree, c("abc", "def"))
+  # TODO: do we expect the change from "abc" -> "Abc"?
+  expect_equal(out$taxon_name, c("Abc", "Def", tree$tip.label))
```
etiennebacher: For example:

``` r
library(palaeoverse)
library(paleotree)
#> Loading required package: ape
data(RaiaCopesRule)
tree <- ceratopsianTreeRaia

head(phylo_check(tree, c("abc", "def")))
#>                    taxon_name present_in_tree present_in_list
#> 38                        Abc           FALSE            TRUE
#> 39                        Def           FALSE            TRUE
#> 1        Centrosaurus_apertus            TRUE           FALSE
#> 2   Styracosaurus_albertensis            TRUE           FALSE
#> 3  Pachyrhinosaurus_perotorum            TRUE           FALSE
#> 4   Pachyrhinosaurus_lakustai            TRUE           FALSE
```

tests/testthat/test-phylo_check.R
```diff
@@ -16,30 +16,158 @@ test_that("phylo_check() works", {
...
+  # input checks
+  expect_snapshot(phylo_check(tree), error = TRUE)
+
+  # TODO: all those cases should error, docs say it should be a character vector
+  # expect_snapshot(phylo_check(tree, list = 1), error = TRUE)
+  # expect_snapshot(phylo_check(tree, list = list()), error = TRUE)
+  # expect_snapshot(phylo_check(tree, list = NA), error = TRUE)
+
+  # TODO: what should be done with NAs?
+  # expect_snapshot(phylo_check(tree, list = c("a", NA)), error = TRUE)
```
etiennebacher: Currently NA are integrated as separate rows in the output, should we maybe error here?

``` r
library(palaeoverse)
library(paleotree)
#> Loading required package: ape
data(RaiaCopesRule)
tree <- ceratopsianTreeRaia

head(phylo_check(tree, c(NA, "def")))
#>                    taxon_name present_in_tree present_in_list
#> 38                       <NA>           FALSE            TRUE
#> 39                        Def           FALSE            TRUE
#> 1        Centrosaurus_apertus            TRUE           FALSE
#> 2   Styracosaurus_albertensis            TRUE           FALSE
#> 3  Pachyrhinosaurus_perotorum            TRUE           FALSE
#> 4   Pachyrhinosaurus_lakustai            TRUE           FALSE
```

tests/testthat/test-phylo_check.R
```diff
@@ -16,30 +16,158 @@ test_that("phylo_check() works", {
...
+
+  # TODO: should it error since "sort" is useless when out = "tree" or "counts"
+  # expect_snapshot(
+  #   phylo_check(tree, list, out = "tree", sort = "az"),
+  #   error = TRUE
+  # )
+  # expect_snapshot(
+  #   phylo_check(tree, list, out = "counts", sort = "az"),
+  #   error = TRUE
+  # )
```
etiennebacher: It is easy to miss that `sort` would be unused when reading code so I think I'd throw an error in this case.

R/phylo_check.R
```diff
@@ -97,6 +97,14 @@ phylo_check <- function(
     )
   }
 
+  if (length(out) == 0 || is.na(out)) {
+    stop("`out` must be of length 1.")
+  }
+
+  if (length(sort) == 0 || is.na(sort)) {
+    stop("`sort` must be of length 1.")
+  }
```
etiennebacher, line 106: Needed to prevent `covr` and `devtools` from reporting different snapshots, cf #235 and #238

tests/testthat/test-phylo_check.R
```diff
@@ -16,30 +16,161 @@ test_that("phylo_check() works", {
...
+})
+
+test_that("arg 'list' works", {
+  skip_if_not_installed("paleotree")
+  library(paleotree)
+  data(RaiaCopesRule)
+  tree <- ceratopsianTreeRaia
+
+  out <- phylo_check(tree, c("abc", "def"))
+  # TODO: do we expect the change from "abc" -> "Abc"?
```
LewisAJones: Interesting! Taxon names should have their first letter capitalised but I don't think we should be transforming data as a side effect in this function.

tests/testthat/test-phylo_check.R
```diff
@@ -16,30 +16,161 @@ test_that("phylo_check() works", {
...
+  expect_equal(out$taxon_name, c("Abc", "Def", tree$tip.label))
+
+  # doesn't accept punctuation
+  expect_snapshot(phylo_check(tree, c("foo.bar")), error = TRUE)
+
+  # input checks
+  expect_snapshot(phylo_check(tree), error = TRUE)
+
+  # TODO: all those cases should error, docs say it should be a character vector
```
LewisAJones, line 47: 👍🏻

tests/testthat/test-phylo_check.R
LewisAJones: Good question. Currently, I think we shouldn't error here as the point of this function is to check the input values against tree. I don't think this would ever happen as it would be quite weird to have NAs in a tree, but users should be able to check whether any input value are present in the tree?

LewisAJones (approved):
A few comments, but looks good to me @etiennebacher. 

As this function was originally developed by @bethany-j-allen, it's worth getting her review and approval before merging.

tests/testthat/test-phylo_check.R
willgearty: I agree with Lewis here. I think including it in the output like this is fine.

willgearty (approved):

tests/testthat/test-phylo_check.R
willgearty: I've seen plenty of phylogenies with lower case taxon names, so we definitely shouldn't do this since the user might go off trying to find the wrong name.

PR #247, opened 2026-07-01
Rework tests for `tax_expand_time()`

etiennebacher:
This PR revamps tests for `tax_expand_time()`. It:

* checks more values


willgearty (approved):
Looks good, thanks!

R/tax_expand_time.R
```diff
@@ -88,12 +88,18 @@ tax_expand_time <- function(
   }
 
   if (
-    length(rank) > 1 ||
+    length(rank) != 1 ||
```
LewisAJones, line 91: What if length of `rank` is equal to 0? i.e. `NULL`

tests/testthat/test-tax_expand_time.R
```diff
@@ -1,89 +1,374 @@
-test_that("tax_expand_time() works", {
+test_that("basic behavior works", {
```
LewisAJones, line 1:
```suggestion
test_that("basic behaviour works", {
```

LewisAJones (changes requested):

R/tax_expand_time.R
etiennebacher: Then this would error with the same message as if a wrong value is given by the user. I think this is what we want?

LewisAJones: I believe you should be able to use `tax_expand_time` by just defining the `bins` argument (scale/rank should essentially be ignored if `bins` are supplied). Currently, you can if you leave the default values, but the following would fail (and shouldn't really):

```
# Define taxdf
taxdf <- data.frame(name = c("A", "B", "C"),
                    max_ma = c(150, 60, 30),
                    min_ma = c(110, 20, 0))
# Define custom bins
bins <- time_bins(scale = "GTS2012", rank = "stage")
# Use custom bins
ex2 <- tax_expand_time(taxdf, bins = bins, rank = NULL)
```

etiennebacher: Given that this error already exists on the main branch (and in the latest release), I added a test that is disabled for now but should be fixed later: 0d2e9407c062c8f62213ca91aa1813c3927d3d56

This change should be reported in NEWS and I'd rather keep these "rework tests" PRs invisible to users.

``` r
library(palaeoverse)
packageVersion("palaeoverse")
#> [1] '1.5.0'

taxdf <- data.frame(name = c("A", "B", "C"),
                  max_ma = c(150, 60, 30),
                  min_ma = c(110, 20, 0))

bins <- time_bins(scale = "GTS2012", rank = "stage")
tax_expand_time(taxdf, bins = bins, rank = NULL)
#> Error in `if (...) NULL`:
#> ! missing value where TRUE/FALSE needed
```

LewisAJones (approved):

PR #248, opened 2026-07-01
Rework tests for `tax_range_space()`

etiennebacher:
This PR revamps tests for `tax_range_space()`. It:

* checks more values


tests/testthat/test-tax_range_space.R
```diff
@@ -1,115 +1,281 @@
...
   )
+
+  small <- tax_range_space(occdf, method = "occ", spacing = 100)
+  large <- tax_range_space(occdf, method = "occ", spacing = 500)
+
+  expect_true(unique(large$spacing) > unique(small$spacing))
+  expect_length(unique(small$spacing), 1L)
+
+  # "spacing" is only relevant for the "occ" method: it is ignored otherwise
+  # TODO: should this error ?
```
willgearty, line 212: I think it would make sense to error (or at least throw a warning)

willgearty (approved):
Looks good, thanks!

R/tax_range_space.R
```diff
@@ -134,6 +134,16 @@ in `occdf`"
     )
   }
 
+  if (length(method) != 1) {
+    stop("`method` must be of length 1.")
+  }
+  if (length(coords) != 1) {
```
LewisAJones, line 140: `coords` should be logical, looks like we're not actually explicitly checking that?

tests/testthat/test-tax_range_space.R
LewisAJones: I guess if we error, we should remove the default value?

tests/testthat/test-tax_range_space.R
```diff
@@ -1,115 +1,281 @@
...
-      occdf = occdf,
-      name = "accepted_name",
-      method = 1
-    ),
+    tax_range_space(occdf, method = "occ", spacing = NA),
     error = TRUE
   )
-  occdf$accepted_name[1] <- NA
+  # TODO: should error
+  # expect_snapshot(
```
LewisAJones, line 232: Are you implementing this as a separate PR?

LewisAJones (changes requested):

R/tax_range_space.R
etiennebacher: There is a test for that (see "argument 'coords' works" in the snapshot results) but the error message will be overriden by `rlang` so I'd rather not make more changes here. Current changes in the function definition (and in all PRs related to tests) are to ensure `devtools::test()` and `covr` play nicely together (cf https://github.com/palaeoverse/palaeoverse/issues/235).

tests/testthat/test-tax_range_space.R
etiennebacher: Yes, when implementing `rlang` input checks.

LewisAJones (approved):

PR #249, opened 2026-07-01
Rework tests for `tax_range_strat()`

etiennebacher:
This PR revamps tests for `tax_range_strat()`. It:

* checks more values


tests/testthat/test-tax_range_strat.R
```diff
@@ -1,186 +1,258 @@
...
+  occdf <- data.frame(
+    genus = c(
+      "Anconastes", "Procolophon", "Procolophon", "Anconastes", "Araeoscelis", "Araeoscelis", 
+      "Edaphosaurus", "Procolophon"
+    ),
+    bed = c(1, 1, 2, 2, 2, 3, 3, 4),
+    certainty = c(1, 1, 0, 1, 0, 1, 1, 1)
+  )
+
+  # "FAD" (default) -> sort by min_bin
```
LewisAJones: Note for later consideration, "FAD" (first-appearance datum) represents the oldest age (e.g. `max_ma`). As set up here, we sort by `min_bin`, which is correct in this context as it represents the first (and oldest bed). However, I can imagine some mental juggling is needed here, so we should consider if there is a way to reduce potential confusion.

Same applies for "LAD".

LewisAJones (approved):
I'll miss the tasty vs. not tasty, but looks good, thanks @etiennebacher!

tests/testthat/test-tax_range_strat.R
etiennebacher: I'm not sure I grasp the nuance here :sweat_smile: I would need an example to understand this better, but I can already remove the "sort by min/max_bin" comments to reduce confusion

LewisAJones: Don't worry about this Etienne, I was just musing. It's not about your comment but the function itself.

willgearty (approved):

PR #263, opened 2026-07-14
Rework tests for `palaeorotate()`

etiennebacher:
Part of #161


willgearty (approved):
Looks good!

R/palaeorotate.R
```diff
@@ -223,6 +226,10 @@ palaeorotate <- function(
     stop("`uncertainty` should be of class logical (TRUE/FALSE).")
   }
 
+  if (length(model) == 0) {
+    stop("`model` should have length > 1.")
```
LewisAJones, line 230: `model` can also be of length 1?

```suggestion
    stop("`model` should have length >= 1.")
```

We could also just state "`model` length equals 0. No model defined."

tests/testthat/test-palaeorotate.R
```diff
@@ -216,4 +342,41 @@ test_that("palaeorotate() grid method works", {
...
+  })
+  expect_equal(nrow(paleomap), 3)
+
+  # input checks
+  expect_snapshot(palaeorotate(occdf = occdf, round = TRUE), error = TRUE)
+  expect_snapshot(palaeorotate(occdf = occdf, round = NA), error = TRUE)
+  expect_snapshot(palaeorotate(occdf = occdf, round = numeric(0)), error = TRUE)
+
+  # TODO: this should error
+  # expect_snapshot(palaeorotate(occdf = occdf, round = 1:2), error = TRUE)
```
LewisAJones, line 381: As you've already modified error handling in `palaeorotate.R` within this PR, worth just adding this now?

LewisAJones (changes requested):

R/palaeorotate.R
etiennebacher: I tweaked the message but in the end I don't think this really matters since it will be replaced by the `rlang` message once we do https://github.com/palaeoverse/palaeoverse/issues/163

tests/testthat/test-palaeorotate.R
etiennebacher: Done

LewisAJones (approved):

Issue #157, opened 2026-04-28
Switch to `cli` package for error messages and progress bars


willgearty:
The `cli` package provides pretty error messages and progress bars. It would be pretty easy to switch over with find/replace. The package is entirely based on base R, so it wouldn't add any extra upstream dependencies.

Ref: https://cli.r-lib.org/index.html

LewisAJones:
Apologies for not commenting before. I don't see any reason why we shouldn't make this transition. Many widely used packages have also already made this transition so it seems landscape appropriate.

etiennebacher:
Not closing this yet because there are more places where we could use `cli`, e.g progress bars (#252)

Issue #261, opened 2026-07-14
`tax_expand_lat()` should warn or error in case of duplicated names with different values


etiennebacher:
Ref: https://github.com/palaeoverse/palaeoverse/pull/242#discussion_r3474563364

``` r
library(palaeoverse)

bins <- data.frame(
  bin = 1:9,
  min = seq(from = 40, to = -40, by = -10),
  mid = seq(from = 45, to = -35, by = -10),
  max = seq(from = 50, to = -30, by = -10)
)

taxdf <- data.frame(
  name = c("A", "A", "C"),
  max_lat = c(60, 50, -10),
  min_lat = c(20, 40, -60)
)
taxdf
#>   name max_lat min_lat
#> 1    A      60      20
#> 2    A      50      40
#> 3    C     -10     -60

tax_expand_lat(taxdf = taxdf, bins = bins)
#>   name max_lat min_lat bin min mid max
#> 1    A      60      20   1  40  45  50
#> 2    A      50      40   1  40  45  50
#> 3    A      60      20   2  30  35  40
#> 4    A      60      20   3  20  25  30
#> 5    C     -10     -60   7 -20 -15 -10
#> 6    C     -10     -60   8 -30 -25 -20
#> 7    C     -10     -60   9 -40 -35 -30
```

From  @LewisAJones:

> Hmmm, in theory, you can have two different species with the same name ([homonyms](https://en.wikipedia.org/wiki/Homonym_(biology))). However, as currently written, the function wouldn't be able to handle this well, and I don't think there is an elegant way to do this. I think perhaps the best approach would be to error on duplicates names, and provide an informative message along the lines of
> 
>     Non-unique name detected. Do you have homonyms in your dataset? Try unique IDs...
> 
> Or something along those lines...

From @willgearty:

> Definitely something that should be checked for several functions and should be made as an internal (or even exported) function. (I envision a comprehensive check_taxonomy function that might have a boolean argument homonyms that allows or warns about homonyms)

PR #265, opened 2026-07-14
Use `cli` and `rlang` in `tax_expand_lat()`

etiennebacher:
This doesn't change existing behavior, it just affects error messages.

Here's what it looks like in the console:

image

Related to #157 and #163


tests/testthat/_snaps/tax_expand_lat.md
```diff
@@ -54,7 +55,28 @@
...
+
+---
+
+    Code
+      tax_expand_lat(taxdf = data.frame(name = "a", max_lat = 91:100, min_lat = 1),
+      bins = bins)
+    Condition
+      Error in `tax_expand_lat()`:
+      ! All values of column "max_lat" in `taxdf` must be between -90 and 90.
+      i Value(s) outside the range (first 5): 91, 92, 93, 94, 95.
```
etiennebacher, line 69: I don't know how much information you like to have in error messages, but IMO it helps to show some wrong values in the error message.

willgearty:
😢 They're gorgeous. And with stacked PRs even!

etiennebacher:
> And with stacked PRs even!

I haven't made up my mind on those yet. They help to see the relationship between PRs, but automatically rebasing 10+ PRs at once can take a looong time in CI (and AFAIK you can't decide which CI should run first).

willgearty (approved):
Looks good, thanks! I might recommend increasing the list of rows to 10 values (since it's row numbers, it shouldn't be too much longer of a message), but 5 is probably fine, too.

R/tax_expand_lat.R
```diff
@@ -50,45 +50,41 @@ tax_expand_lat <- function(
...
+    )
+    cli::cli_abort(
+      c(
+        "Maximum latitude must be larger than or equal to minimum latitude.",
+        "i" = "Row(s) where max latitude is smaller than min latitude{truncated}: {.val {to_report}}."
+      )
+    )
   }
 
   if (anyDuplicated(taxdf) > 0) {
```
LewisAJones, line 86: For a future PR. This checks whether entire rows are duplicated, but what we're probably wanting to know is if names are duplicated? This would also be bypassed if `taxdf` had columns not relevant for the function but were still present. I believe this is a bug.

R/tax_expand_lat.R
```diff
@@ -50,45 +50,41 @@ tax_expand_lat <- function(
...
+      ""
+    }
+    to_report <- cli::cli_vec(
+      head(rows_with_max_lat_smaller_than_min_lat, n = 5),
+      list(`vec-last` = ", ")
+    )
+    cli::cli_abort(
+      c(
+        "Maximum latitude must be larger than or equal to minimum latitude.",
+        "i" = "Row(s) where max latitude is smaller than min latitude{truncated}: {.val {to_report}}."
```
LewisAJones: I'm fine with this either way as long as we standardise, but a possible alternative could be quoting input arguments (we do this elsewhere in the package), e.g.:

```
 "i" = "Row(s) where `max_lat` is smaller than `min_lat`{truncated}: {.val {to_report}}."
```

R/tax_expand_lat.R
```diff
@@ -50,45 +50,41 @@ tax_expand_lat <- function(
...
-  if (any(c(taxdf[, c(min_lat, max_lat)] > 90))) {
-    stop("Maximum and minimum latitudes must be less than or equal to 90")
-  }
-
-  if (any(taxdf[, max_lat, drop = TRUE] < taxdf[, min_lat, drop = TRUE])) {
-    stop("Maximum latitude must be larger than or equal to minimum latitude")
+  rows_with_max_lat_smaller_than_min_lat <- which(
+    taxdf[, max_lat, drop = TRUE] < taxdf[, min_lat, drop = TRUE]
+  )
+  if (length(rows_with_max_lat_smaller_than_min_lat) > 0) {
```
LewisAJones, line 68: I think this is great. We could also use 6 (same default as `head`?), but 5 is a nice number.

Is this more useful than returning some quoted code that the user could run for checking which values are errors? Thinking in the case of a large dataset. They could also assign the rows to an object then?

R/tax_expand_lat.R
```diff
@@ -50,45 +50,41 @@ tax_expand_lat <- function(
...
-  }
-
-  if (!all(c("bin", "max", "min") %in% colnames(bins))) {
-    stop("Either 'bin', 'max' or 'min' is not a named column in `bins`")
-  }
-
-  if (!all(c(max_lat, min_lat) %in% colnames(taxdf))) {
-    stop("Either `max_lat` or `min_lat` is not a named column in `taxdf`")
-  }
+  rlang::check_data_frame(taxdf)
```
LewisAJones: Do we need to be mindful/careful of the fact that a lot of these functions from `rlang` are currently "experimental"?

R/utils-rlang.R
```diff
@@ -0,0 +1,36 @@
+check_column_presence <- function(data, column) {
```
LewisAJones, line 9: I think it is obvious for the current functions, especially given the naming convention. However, if we start to have more complicated functions here, we might want to add some docs?

R/utils-rlang.R
```diff
@@ -0,0 +1,36 @@
...
+  )
+  if (!column %in% names(data)) {
+    cli::cli_abort(
+      "Column {.val {column}} not found in {.arg {rlang::caller_arg(data)}}.",
+      call = rlang::caller_env()
+    )
+  }
+}
+
+check_range <- function(data, x, min, max) {
```
LewisAJones: Should `x` be `column` to try to maintain some naming consistency?

R/utils-rlang.R
```diff
@@ -0,0 +1,36 @@
...
+  vals <- data[[x]]
+  if (!is.numeric(vals)) {
```
LewisAJones, line 40: Should we explicitly check for class here, or should this be a separate generalised function (e.g. `check_class`)?

tests/testthat/_snaps/tax_expand_lat.md
```diff
@@ -4,39 +4,39 @@
       tax_expand_lat(taxdf = 5)
     Condition
       Error in `tax_expand_lat()`:
-      ! `taxdf` should be a dataframe
+      ! `taxdf` must be a data frame, not the number 5.
```
LewisAJones: I guess this is the default implementation in `rlang` but one thing we've done in some places in the package is quoting the class directly in such instances, e.g.

```
`taxdf` must be of class data.frame, not the number 5.
```

I'm not tied to this, but it perhaps add some level of clarification for users on exact class? However, the docs of `data.frame` refer to data frames so I'm probably just being difficult.

R/tax_expand_lat.R
etiennebacher: Done in [05152ec](https://github.com/palaeoverse/palaeoverse/pull/265/commits/05152ec0e5a68d5e73ca15b290b09d4ec80aee65)

R/utils-rlang.R
etiennebacher: Done in [f372150](https://github.com/palaeoverse/palaeoverse/pull/265/commits/f37215052149deae6a67a37ec52007da2190fb81)

R/utils-rlang.R
etiennebacher: Done in [f372150](https://github.com/palaeoverse/palaeoverse/pull/265/commits/f37215052149deae6a67a37ec52007da2190fb81)

tests/testthat/_snaps/tax_expand_lat.md
etiennebacher: This message is generated by `rlang::check_data_frame()` so we can't change it directly. However, `rlang` provides some "standalone check helpers" to easily build your own check functions.  

In https://github.com/palaeoverse/palaeoverse/pull/265/changes/ba356ad81effbd4c7767b892e8551a2c5cb35269, I have run `usethis::use_standalone("r-lib/rlang", file = "types-check")` to import these standalone checks, and I added a custom `check_data_frame()` with the error message you wanted. This may seem like a big change for such a small tweak in the error message, but we will likely have to import these standalone files anyway in other cases (e.g. in #293 because weirdly there is no `rlang::check_function()`).

Let me know if you're happy with that.

Edit: a nice side effect of using our own `check_data_frame()` is that this addresses the concern about the "experimental" status of `rlang::check_data_frame()`.

R/utils-rlang.R
etiennebacher: I think this is fine to leave here for now because we want the error message to report the name of the dataframe and of the column. We might want to refactor later once we have a better view of input checks in other functions.

R/tax_expand_lat.R
etiennebacher: See comment below about the error message of `check_data_frame()`.

R/tax_expand_lat.R
etiennebacher: I wanted to add an extra point in  the message, like `` See problematic rows with `taxdf[taxdf$max_lat < taxdf$min_lat , ]` `` (and replace `taxdf`, `max_lat` and `min_lat` by the user inputs). The problem is that if the user passes an anonymous dataframe in `taxdf` (as below) then this code suggestion would become unreadable:

```r
    tax_expand_lat(
      taxdf = data.frame(
        name = c("A", "B", "C"),
        max_lat = c(60, 20, -10),
        min_lat = c(72, -40, -60)
      ),
      bins = bins
    )
```

R/tax_expand_lat.R
etiennebacher: Yes I think this is https://github.com/palaeoverse/palaeoverse/issues/261

willgearty (approved):
I thought it was fine with just "data frame", but I guess this is fine, too (although, do general users know what a "class" is?)

R/tax_expand_lat.R
LewisAJones: That would be nice, but I agree in that edge case it would not be great. 

Would it potentially lead to more confusion if we treated them as placeholders (i.e. not replacing taxdf, max_lat, min_lat, etc with user inputs)?

R/utils-rlang.R
LewisAJones: That's fair and I'm fine with that for now. I know there are a lot of instances where we check explicitly for class though, so we likely will want something generalised at some point.

R/utils-rlang.R
```diff
@@ -0,0 +1,78 @@
...
+#' This duplicates the behaviour of `rlang::check_data_frame()` because we wanted
+#' the error message to contain "be of class dataframe".
+#'
+#' @param data dataframe to check
+#'
+#' @noRd
+check_data_frame <- function(data) {
+  if (rlang::is_missing(data) || !is.data.frame(data)) {
+    cli::cli_abort(
+      "{.arg {rlang::caller_arg(data)}} must be of class dataframe, not {obj_type_friendly(data)}.",
```
LewisAJones: I guess this still isn't quoting the class directly i.e.

```
> class(data.frame())
[1] "data.frame"
```

Is this good practice to try to use throughout the package? Especially as we start to use our own classes?

```suggestion
      "{.arg {rlang::caller_arg(data)}} must be of class 'data.frame', not {obj_type_friendly(data)}.",
```

Happy for this not to be implemented though if we don't see value in that.

LewisAJones:
> I thought it was fine with just "data frame", but I guess this is fine, too (although, do general users know what a "class" is?)

I think so? Aren't classes normally one of the first things covered in Introduction to R courses? We are also planning to have custom classes in the package, so maybe this is also an area to better passively communicate classes?

LewisAJones (approved):

R/tax_expand_lat.R
etiennebacher: I think this loses most of its interest if it cannot be directly copied and pasted in the console because it forces the user to tweak this line of code in several places, which in the end may take longer than just typing it themselves. I know this is also to show beginners how to do this, but at the same time I think filtering data is one of the first things you learn when you start using dataframes.

R/utils-rlang.R
etiennebacher: `cli` actually provides custom class formatting with `{.cls}` (that I forgot to use here), it looks like this:

image

I can add it if you're happy with that.

LewisAJones: I like it!

etiennebacher: Done in https://github.com/palaeoverse/palaeoverse/pull/265/commits/2a5da28731481699c86df9ddbfb709c299c2f27a

PR #272, opened 2026-07-23
Use `cli` and `rlang` in `bin_lat()`

etiennebacher:
No behaviour changed, just error messages.


willgearty (approved):

R/bin_lat.R
```diff
@@ -35,25 +35,23 @@
...
+  check_data_frame(occdf)
+  check_data_frame(bins)
+  rlang::check_bool(boundary)
+  check_column_presence(occdf, lat)
+  check_column_presence(bins, "min")
+  check_column_presence(bins, "max")
+  check_column_presence(bins, "bin")
+
+  lat_vals <- occdf[[lat]]
+  if (anyNA(lat_vals)) {
```
LewisAJones: Should this be pulled out into a generalised function? We will want to check whether NAs are present in many functions throughout the package. This should help standardise our treatment of these values too?

R/bin_lat.R
```diff
@@ -35,25 +35,23 @@
...
-    stop("`lat` contains NA values")
-  }
-  if (any(occdf[, lat, drop = TRUE] > 90 | occdf[, lat, drop = TRUE] < -90)) {
-    stop("Latitudes should be more than -90 and less than 90")
-  }
-  if (sum(c("max", "min", "bin") %in% colnames(bins)) != 3) {
-    stop("`bins` does not contain bin, max and min named columns")
+  check_data_frame(occdf)
+  check_data_frame(bins)
+  rlang::check_bool(boundary)
```
LewisAJones, line 40: We should establish clear rules for when we use `::` vs. `@importFrom`. I think the latter has historically been preferred if a function is used repeatedly throughout the package, while the former when it is infrequently used. I think the former is also useful when there are very similar name functions from different packages? 

Either way, I think this would be a nice clarification point in the contributor/developer guide?

LewisAJones (changes requested):

R/bin_lat.R
etiennebacher: I opened https://github.com/palaeoverse/palaeoverse/issues/302 to track this.

R/bin_lat.R
etiennebacher: I created `check_na()` in https://github.com/palaeoverse/palaeoverse/pull/272/changes/201ce824ab018b4c72561b443e544c6c1931ee77 but we might need to refactor that later on. We want to report the names of the column and of the dataframe in this case, but in some cases we might just want to report the name of the argument passed by the user, e.g. `tax_check(dis = NA)` should just mention `dis` in the error message.

LewisAJones (approved):
Thanks @etiennebacher, looks good... just double check with @willgearty that he is happy with these changes since his last approval before merging.

willgearty (approved):

PR #282, opened 2026-08-05
Replace `plot` parameters by `plot()` S3 methods

etiennebacher:
This PR is an experiment related to #352.

It removes the `plot` and `plot`-related args (like `plot_args`) from all functions. Instead, those functions now have:

- a custom class (e.g. `palaeo_lat_bins`)
- custom attributes containing the necessary info for plotting
- dedicated S3 plotting methods

This removes a side effect of those functions (plotting) and simplifies their definition.

If we choose to go this way, we'll probably want to keep those arguments but deprecate them (and internally call `plot()` when `plot = TRUE`).


etiennebacher:
Closing in favour of one PR per function

PR #267, opened 2026-07-14
Run `lychee` on rendered website, fix more dead links

etiennebacher:
I just realised that we run `lychee` on the current repo but it cannot check many links in `roxygen2` docs or in vignettes. This PR runs `lychee` on the `pkgdown` website after rendering it in the docs workflow but before its deployment, meaning that it will block the docs deployment if some links are broken.

This PR also brings several related fixes:

- remove leading space in DOI in `DESCRIPTION`
- update dead links to https://stratigraphy.org/
- regenerate several favicons using `pkgdown::build_favicons(overwrite = TRUE)`
- fix links to "About us" page in vignettes   

## Checklist before requesting a review
- [x] I have read palaeoverse's [standards and structure](https://palaeoverse.palaeoverse.org/articles/structure-and-standards.html)
- [x] I have performed a self-review of my code.
- [ ] I have added function documentation.
- [ ] I have provided sufficient examples of implementation.
- [ ] If it is a core feature, I have added thorough tests.

PR #283, opened 2026-08-06
Remap links before checking them in `pkgdown` workflow

etiennebacher:
#267 added the links check for the rendered `pkgdown` website. This would fail in some cases, for instance if a PR adds a new function and a new reference page for this function. The rendered website would create a link to `palaeoverse.palaeoverse.org/reference/my_new_page.html`, but this page wouldn't exist until the PR is merged and the website is updated, causing the links check to fail.

This PR solves this by forcing `lychee` to remap `palaeoverse.palaeoverse.org` to the `docs` folder in the github workspace used by each PR.

The first three commits in this PR show that the fix works:

1. introduce a fake function `foo()` that would make the `pkgdown` workflow [fail](https://github.com/palaeoverse/palaeoverse/actions/runs/31087904344/job/92571650385?pr=283#step:7:199) because `palaeoverse.palaeoverse.org/reference/foo.html` doesn't exist
2. introduce the remapping procedure in the `pkgdown` workflow, which [passes](https://github.com/palaeoverse/palaeoverse/actions/runs/31088347679/job/92573077354?pr=283#step:7:140)
3. remove the fake function

The `--remap` changes were added by Claude, I reviewed the changes and added the comments.


willgearty (approved):
Looks good, thanks for the walkthrough!

LewisAJones (approved):
Thanks for the thorough line comments, looks good!

Issue #240, opened 2026-06-24
Show function output in "Examples" section


etiennebacher:
For instance, the [reference page for `tax_certainty()`](https://palaeoverse.palaeoverse.org/reference/tax_certainty.html) shows these examples:

<img width="761" height="404" alt="Image" src="https://github.com/user-attachments/assets/858ce168-4285-417a-94b0-52570fae46b5" />

This is useful but I think it would be better to print the results instead of assigning them to objects, so that someone reading the docs doesn't need to open R and run this to see what the output looks like. Printing results would require taking less rows and columns in the input, otherwise the output would be unreadable.

willgearty:
Could also do e.g. `head(certainty)`

Taha-Waqar:
I'd like to work on this! Just to confirm scope - should I fix the example output across all functions in the package, or start with tax_certainty() and a few others? Happy to follow the head() pattern suggested above.

LewisAJones:
If I recall correctly, we previously used to use `head()` for examples but it was problematic for outputs with many columns, and resulted in the reference pages ending up being very long/ugly (you either have to make the output code chunk scrollable, wrap across lines, or return just a subset of data across columns). Happy for you to experiment again with this though!

willgearty:
Some other options that might work:
- base `str()`
- `broom::glance()`
- `pillar::glimpse()` (also exported by dplyr)

etiennebacher:
Hi @Taha-Waqar, thanks for your interest in this issue. We'll have a dedicated programme to bring more contributors to Palaeoverse in the next few months, so we are reserving some issues exclusively for participants to this programme. Feel free to explore other "good first issues" but note that we might still reserve a few more for this programme.

PR #291, opened 2026-08-25
Remove some unused objects

etiennebacher:
This removes some unused objects (i.e. objects that are defined but not used) detected by Jarl 0.6.0 with `jarl check . --select unused_object`. These changes are not included in #290 because this rule is disabled by default:

- there are (few) known false positives that need to be fixed in Jarl
- many of these unused objects are in the `roxygen` examples and I think these will be addressed later when we revamp the examples (e.g. #240)

Depends on #290.

### DIFF
--- a/vignettes/tetrapod-biodiversity.Rmd
+++ b/vignettes/tetrapod-biodiversity.Rmd
@@ -213,6 +213,9 @@ head(maj_tetrapods[, c("cell_ID", "cell_centroid_lng", "cell_centroid_lat")])
 Now that we have both temporal and spatial bins, we can use these data to see how the spatial coverage of tetrapod fossils in our dataset varies through time.
 
 ```{r message = FALSE}
+# Extract unique interval midpoints
+midpoints <- sort(unique(maj_tetrapods$bin_midpoint))
+
 # Calculate the number of unique cells in each time bin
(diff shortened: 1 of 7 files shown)


vignettes/tetrapod-biodiversity.Rmd
```diff
@@ -213,9 +213,6 @@ head(maj_tetrapods[, c("cell_ID", "cell_centroid_lng", "cell_centroid_lat")])
 Now that we have both temporal and spatial bins, we can use these data to see how the spatial coverage of tetrapod fossils in our dataset varies through time.
 
 ```{r message = FALSE}
-# Extract unique interval midpoints
-midpoints <- sort(unique(maj_tetrapods$bin_midpoint))
-
```
etiennebacher, line 218: Currently this is never shown or explained to the reader.

I chose to remove this but maybe we should add a couple of sentences and show the results in the vignette instead.

LewisAJones: It looks like this is safe to remove. Perhaps leave it at that for now, we will likely be reviewing and updating the vignettes with the incoming refactor changes anyway.

LewisAJones (approved):

willgearty (approved):

PR #293, opened 2026-08-26
Use `cli` and `rlang` in `group_apply()`

etiennebacher:
No behaviour changed, just error messages.


R/group_apply.R
```diff
@@ -85,34 +85,32 @@
...
+    cli::cli_abort("{.arg group} must specify at least one column.")
   }
-
-  if (!is.function(fun)) {
-    stop("Supplied `fun` is not a function")
+  unknown_cols <- setdiff(group, colnames(occdf))
+  if (length(unknown_cols) > 0) {
+    cli::cli_abort(
+      "Column{?s} {.val {unknown_cols}} not found in {.arg occdf}."
+    )
```
etiennebacher, line 98: This function is the only place where the user can pass multiple column names in a single argument, which is why I do this check manually instead of using a helper.

willgearty (approved):
Looks good, thanks!

R/group_apply.R
```diff
@@ -85,34 +85,32 @@
 #' ex4 <- group_apply(occdf = occdf, group = "lat_bin", fun = nrow)
 #' @export
 group_apply <- function(occdf, group, fun, ...) {
-  # Handle errors
-  if (!is.data.frame(occdf)) {
-    stop("`occdf` should be a dataframe")
-  }
+  rlang::check_data_frame(occdf)
```
LewisAJones:
```suggestion
  check_data_frame(occdf)
```
I think we're supposed to be using an internal version?

LewisAJones (changes requested):

R/group_apply.R
etiennebacher: Yes, good catch!

LewisAJones (approved):

Issue #216, opened 2026-06-08
Check that all arguments are named


etiennebacher:
Instead of adding `...` to most functions and checking that these dots are empty, we will add a new helper `ensure_args_are_named()`.

### Partial argument matching

We decided to forbid partial argument matching since this can be prone to bugs.

### Differences in error messages

With this approach:

``` r
devtools::load_all("/home/etienne/Desktop/Git/palaeoverse")
#> ℹ Loading palaeoverse

lat_bins_area(2, 3, 10)
#> Error in `lat_bins_area()`:
#> ! All arguments must be named.
#> ℹ Currently, there are 3 arguments that should be named.

lat_bins_area(2, 3, foo = 10)
#> Error in `lat_bins_area()`:
#> ! unused argument (foo = 10)
```

With `...`:

``` r
devtools::load_all("/home/etienne/Desktop/Git/palaeoverse")
#> ℹ Loading palaeoverse

lat_bins_area(2, 3, 10)
#> Error in `lat_bins_area()`:
#> ! `...` must be empty.
#> ✖ Problematic arguments:
#> • ..1 = 3
#> • ..2 = 10
#> ℹ Did you forget to name an argument?

lat_bins_area(2, 3, foo = 10)
#> Error in `lat_bins_area()`:
#> ! `...` must be empty.
#> ✖ Problematic arguments:
#> • ..1 = 3
#> • foo = 10
#> ℹ Did you forget to name an argument?
```

The error message in case of a misnamed argument is less clear if we use `...`.

### Differences in documentation

With this approach, we should specify in the documentation which arguments can be unnamed and which have to be named. The function definition doesn't change.

With `...`, we modify the function definition and we must add some documentation such as ``@param ... These dots should be empty, all arguments after `...` must be named. ``

---

<details>

<summary>Old description</summary>

Currently, all functions accept passing inputs by position only, e.g.:

 ```r
lat_bins_area(6, 15, 30, 6000)
```

This can lead to code that is very hard to read and prone to mistakes. It also means that we are constrained by the current argument order if we decide to change the argument order or to add a new argument in the future. For instance, if the function definition is like this:

```r
function(n = 12, min = -90, max = 90, r = 6371, plot = FALSE)
```
then we will not be able to add an argument between `min` and `max` because doing so would break code that relies only on arguments position.

[...]

PR #294, opened 2026-08-27
Use new helper `ensure_args_are_named()` in `lat_bins_area()`

etiennebacher:
This PR describes one way to address  #216. 

I preferred opening a draft PR rather than continuing the discussion in #216. I think this is easier to directly try different approaches and see what's best. In addition to the PR description, the new snapshot results show how this would behave in practice.

What do you think about the  approach described below? 

--- 

Instead of adding `...` to most functions and checking that these dots are empty, this PR adds a new helper `ensure_args_are_named()` to ensure that all arguments that must be named are actually named. It is possible to pass an argument `exceptions` to specify arguments that can be unnamed (if we had `...`, these arguments would be placed before `...`).

### Differences in behaviour

This approach allows partial matching of argument names:

``` r
devtools::load_all("/home/etienne/Desktop/Git/palaeoverse")
#> ℹ Loading palaeoverse

# "mi" -> "min", "ma" -> "max"
lat_bins_area(2, mi = 1, ma = 2)
#>   bin      min      mid      max         area area_prop
#> 1   1 1.499943 1.749971 2.000000 2.224785e+12       0.5
#> 2   2 1.000000 1.249971 1.499943 2.224785e+12       0.5
```

Using `...` wouldn't:

``` r
devtools::load_all("/home/etienne/Desktop/Git/palaeoverse")
#> ℹ Loading palaeoverse

lat_bins_area(2, mi = 1, ma = 2)
#> Error in `lat_bins_area()`:
#> ! `...` must be empty.
#> ✖ Problematic arguments:
#> • mi = 1
#> • ma = 2
```

Whether this is desirable is up for discussion. Personally I generally prefer having full argument names.

### Differences in error messages

With this approach:

``` r
devtools::load_all("/home/etienne/Desktop/Git/palaeoverse")
#> ℹ Loading palaeoverse

lat_bins_area(2, 3, 10)
#> Error in `lat_bins_area()`:
#> ! All arguments must be named (except for "n").
#> ℹ Currently, there are 2 arguments that should be named.

lat_bins_area(2, 3, foo = 10)
#> Error in `lat_bins_area()`:
#> ! unused argument (foo = 10)
```

With `...`:

``` r
devtools::load_all("/home/etienne/Desktop/Git/palaeoverse")
#> ℹ Loading palaeoverse

lat_bins_area(2, 3, 10)
#> Error in `lat_bins_area()`:
#> ! `...` must be empty.
#> ✖ Problematic arguments:
#> • ..1 = 3
#> • ..2 = 10
#> ℹ Did you forget to name an argument?

lat_bins_area(2, 3, foo = 10)
#> Error in `lat_bins_area()`:
#> ! `...` must be empty.
#> ✖ Problematic arguments:
#> • ..1 = 3
#> • foo = 10
#> ℹ Did you forget to name an argument?
```

The error message in case of a misnamed argument is less clear if we use `...`.

### Differences in documentation

With this approach, we should specify in the documentation which arguments can be unnamed and which have to be named. The function definition doesn't change.

With `...`, we modify the function definition and we must add some documentation such as ``@param ... These dots should be empty, all arguments after `...` must be named. ``


R/utils.R
```diff
@@ -0,0 +1,28 @@
+#' Check whether some arguments are unnamed but shouldn't. This is an alternative to
+#' putting `...` in the function definition and checking whether they are empty.
```
willgearty:
```suggestion
#' Check whether there are unnamed arguments that should be named.
```

willgearty (approved):
I like this strategy. I definitely agree that the error messages are better, and I think partial argument names are infinitely better than no argument names (and is probably an uncommon edge case).

LewisAJones:
Thanks for this @etiennebacher, I love to see worked examples, much easier to follow.

From these options, I prefer the helper function by quite a bit. However, I really do not like the partial matching of argument names and I think we should strictly enforce full argument names. 

While I see where @willgearty is coming from in that it is perhaps better than no argument name, I don't think this is going to help individuals read/review code in future. I could also imagine a scenario where we add a new argument which is similar in name to another which might break existing code?

R/lat_bins_area.R
```diff
@@ -40,6 +40,8 @@
 #' # Generate latitudinal bins and a plot
 #' bins <- lat_bins_area(n = 24, plot = TRUE)
 lat_bins_area <- function(n = 12, min = -90, max = 90, r = 6371, plot = FALSE) {
+  ensure_args_are_named(exceptions = "n")
```
LewisAJones: Should there be exceptions? When would we want this?

etiennebacher:
I also don't like partial matching of arg names. We could forbid it and have a message like this:

image

> I could also imagine a scenario where we add a new argument which is similar in name to another which might break existing code?

Yes, this could happen, I had forgotten about this.

R/lat_bins_area.R
etiennebacher: All examples use named arguments everywhere, but there are some occurrences in tests or vignettes where the first argument is unnamed, e.g. `group_apply(space_tetrapods, "bin_midpoint", function(df) mean(df$area))` or `phylo_check(tree, c("abc", "def"))`. Maybe we want to keep this possibility?

Hard for me to say how often people use unnamed arguments in these functions.

LewisAJones:
> I also don't like partial matching of arg names. We could forbid it and have a message like this:
> 
> image

I'm fine with this, but it perhaps should be "Invalid argument names supplied" instead of "Argument names must be fully written"... this then puts emphasis on there being a mistake/error and probably covers the more likely scenario of a spelling mistake?

etiennebacher:
A spelling mistake would actually be caught before this check:

image

This is why I wanted to mention the "must be fully written" part of the message.

R/lat_bins_area.R
LewisAJones: I think in `phylo_check` they should be named arguments. I don't see a need for exceptions here.

`group_apply` is an exception because you provide a function and any additional arguments that function may have, for example with a palaeoverse function:

```
group_apply(
  occdf = occdf, 
  group = "collection_no", 
  fun = tax_unique, 
  genus = "genus", 
  family = "family", 
  order = "order", 
  class = "class", 
  resolution = "genus"
)
```

I guess in the case of palaeoverse functions, `ensure_args_are_named` would still get used with `group_apply` but this would not be the case for non-palaeoverse functions. Would exceptions handle `...`?

I am not sure where else in the package, we would want exceptions.

LewisAJones:
> A spelling mistake would actually be caught before this check:
> 
> image
> This is why I wanted to mention the "must be fully written" part of the message.

~In which case, the code shouldn't ever run to the partial name checking?~

Nevermind (my own misunderstanding), we do need this check for arguments needing to be fully written.

etiennebacher:
@willgearty there are two main changes since your last review:

1. partial matching of arg names is forbidden
2. all arguments must be named (there are no more `exceptions`)

Are you fine with those?

R/lat_bins_area.R
etiennebacher: I think we want all arguments passed in `...` in `group_apply()` to be named, but I'm pretty sure we can customise `ensure_args_are_named()` later on if we want to allow unnamed args in `...` because a custom function is applied.

https://github.com/palaeoverse/palaeoverse/pull/294/changes/abe21a6b8e0520defcfbafb48def950d6ce8d73d removes `exceptions`.

LewisAJones: I agree with all arguments being named, even those supplied to `group_apply()`, I just wasn't sure how easy it would be to enforce for custom input functions 👍🏻

willgearty (approved):
Yep, looks good to me!

LewisAJones (approved):

Issue #171, opened 2026-05-14
Produce a prototype for one or several `"palaeoverse"` classes


etiennebacher:
There should be a function that runs multiple checks on the data and outputs a dataframe with a class `"palaeoverse"` (or similar). This class indicates to other `palaeoverse` functions that those checks were already run. 

But now that I'm writing this, it doesn't seem ideal: what if my data passes the checks but then I do some changes to it? The functions I apply on it will not necessarily remove the custom classes, so future calls to other `palaeoverse` functions will wrongly assume that the data hasn't changed since the checks.

willgearty:
Hmm...that's a good point. So maybe the only point of the function would be to rename the columns so the user doesn't need to provide custom arguments to downstream functions? But we will still always need to perform the checks, I guess?

LewisAJones:
How do other packages handle this? `sf` is probably a good example? 

The checks could take place both within the initial function and subsequent functions with the option to control whether they are performed via an argument? 

Alternatively, can we not do something along the lines of warning the user and stripping the class when it is modified? Or when running the initial function, we could message the user that they should re-run the function if they make any changes to the data?

willgearty:
sf has `st_sf()` which is similar to what our cleaning function would be:
https://github.com/r-spatial/sf/blob/72ad7d000b939c375c7ba3d6ec1e4982119b81c8/R/sf.R#L240-L321

Because they are only ever dealing with one special column, checking for corruption seems to be fairly simple, for example:
```r
st_geometry.sf = function(obj, ...) {
	ret =  obj[[attr(obj, "sf_column")]]
	if (!inherits(ret, "sfc")) # corrupt!
		stop('attr(obj, "sf_column") does not point to a geometry column.\nDid you rename it, without setting st_geometry(obj) <- "newname"?')
	ret
}
```

Otherwise, I don't see much duplicate checking in other functions

willgearty:
Ah, this is the magic sauce for {sf}:
https://github.com/r-spatial/sf/blob/72ad7d000b939c375c7ba3d6ec1e4982119b81c8/R/sf.R#L325-L428

This all defines what happens when an `sf` object is modified (for example, if you remove the `geometry` column, it just becomes a normal `data.frame` again without the `sf` class).

Same thing for defining what happens when tidyverse functions are used on an `sf` object here:
https://github.com/r-spatial/sf/blob/main/R/tidyverse.R (`dplyr_reconstruct.sf()` at the top does most of the work)
[...]

Issue #145, opened 2025-07-29
map_occdf


LewisAJones:
In palaeoverse, we have several common arguments in functions with default values (column names) such as `max_ma = "max_ma"`. Should we offer a function which allows users to standardise and map their occurrence datasets to these defaults.

willgearty:
So basically `dplyr::rename()` but specific to Palaeoverse? I could see this being pretty useful. It could be used at the very beginning of a pipeline to save the user a bunch of time/code later.

LewisAJones:
Hi @harrietdrage, as discussed at CPEG, this is probably a good first function to develop to gain a bit of experience. If you are still interested, we'll reserve this issue for you to resolve for when you have time. We can also have a proper chat/call before you get started and I'll make myself available to help 😄

LewisAJones:
A couple of additional features we could also consider for this function if we wanted to broaden scope:

- Convert from tibble to dataframe
- Generic error checking 

This function could then form the basis for a pipeline by ensuring your input dataframe is compatible with all palaeoverse functions?

LewisAJones:
This is definitely something to discuss if we get the RSMF....

PR #295, opened 2026-08-27
Implement `as_palaeo()`

etiennebacher:
Related to #145 and #171


R/as_palaeo.R
```diff
@@ -0,0 +1,112 @@
...
+#' bin_lat(occdf = dat, bins = bins)
+as_palaeo <- function(x, lat = "lat", lon = "lon") {
+  ensure_args_are_named(exceptions = "x")
+  rlang::check_data_frame(x)
+
+  if (missing(lon)) {
+    if ("lon" %in% names(x)) {
+      attr(x, "palaeo_lon") <- "lon"
+    }
+  } else {
```
etiennebacher, line 52: This means that if the data contains a column `"lon"` and the user calls `as_palaeo(my_data)` then we automatically register `"lon"` as the longitude attribute.

This was one of the decisions we made during the refactor meeting and I can see how this is a bit more convenient, but I think it would be better  to be explicit, especially in a function that affects how many other `palaeoverse` functions process the data.

R/as_palaeo.R
```diff
@@ -0,0 +1,118 @@
...
+#' names(dat)[names(dat) == "lat"] <- "my_lat"
+#'
+#' # We can store this information in as_palaeo()
+#' dat <- as_palaeo(dat, lat = "my_lat")
+#'
+#' # Other palaeoverse functions that have a "lat" argument will automatically use
+#' # this information (unless another value is explicitly passed in the call)
+#' bins <- lat_bins_degrees(size = 10)
+#' bin_lat(occdf = dat, bins = bins)
+as_palaeo <- function(x, lat = "lat", lon = "lon") {
```
willgearty: It might be worthwhile making the argument dictionary before we make this function so we can get a better handle on what the (required and/or optional) arguments should be for this function. For example, taxonomic data (e.g., used for `tax_expand_time`) probably won't have lat/lon values.

R/as_palaeo.R
willgearty: So would every argument be set to `NULL` and the user would need to specify them in order to register the column?

R/as_palaeo.R
etiennebacher: I think it's already useful to review this PR for the infrastructure and how it would work. We can determine the arguments to add or remove from this function later on.

R/as_palaeo.R
```diff
@@ -0,0 +1,118 @@
...
+  ensure_args_are_named(exceptions = "x")
+  rlang::check_data_frame(x)
```
LewisAJones: Should we be using our internal `check_data_frame` here?

R/as_palaeo.R
LewisAJones: I guess `x` could be `data` to align with the rest of the package?

R/as_palaeo.R
```diff
@@ -0,0 +1,118 @@
...
+
+  if (missing(lon)) {
+    if ("lon" %in% names(x)) {
+      attr(x, "palaeo_lon") <- "lon"
```
LewisAJones: Why not just "lon" here? This might become confusing when we are using both palaeocoordinates and regular coordinates?

Did you have concerns about conflicts again?

R/as_palaeo.R
```diff
@@ -0,0 +1,118 @@
...
+#'
+#' # Other palaeoverse functions that have a "lat" argument will automatically use
+#' # this information (unless another value is explicitly passed in the call)
+#' bins <- lat_bins_degrees(size = 10)
+#' bin_lat(occdf = dat, bins = bins)
+as_palaeo <- function(x, lat = "lat", lon = "lon") {
+  ensure_args_are_named(exceptions = "x")
+  rlang::check_data_frame(x)
+
+  if (missing(lon)) {
```
LewisAJones, line 48: We will probably want to use `*apply`, for loop, or something here eventually when we have many arguments?

R/as_palaeo.R
```diff
@@ -0,0 +1,118 @@
...
+  if (missing(lat)) {
+    if ("lat" %in% names(x)) {
+      attr(x, "palaeo_lat") <- "lat"
+    }
+  } else {
+    check_column_presence(x, lat)
+    attr(x, "palaeo_lat") <- lat
+  }
+
+  class(x) <- c("palaeo", class(x))
```
LewisAJones: "palaeoverse_data", "palaeoverse_data.frame", or similar?

R/as_palaeo.R
LewisAJones: I agree with both of you. I think the dictionary probably needs to come first, but it is useful having the infrastructure here to see how it would be implemented and work in practice.

R/as_palaeo.R
LewisAJones: I agree being explicit is valuable but at the same time this could be quite annoying for users when registering many columns, especially when they are quite consistent (e.g. taxonomic hierarchy):

```
as_palaeo(data, phylum = "phylum", class = "class", order = "order"...)
``` 

Would a middle ground be printing to the console the actual attributes after running the function?

R/as_palaeo.R
```diff
@@ -0,0 +1,118 @@
...
+    cli::cli_inform(c("i" = "Attributes:"))
+    # Indent the bullet points
+    d <- cli::cli_div(
+      theme = list(ul = list("margin-left" = 2, "padding-left" = 0))
+    )
+    cli::cli_ul(paste0(att, ": \"", att_list, "\""))
+    cli::cli_end(d)
+  }
+  cat("\n")
+  print.data.frame(x)
```
LewisAJones, line 90: Just a thought, most occurrence dataframes have many columns, so using print here doesn't easily reveal the attributes of the dataframe. Would it be more useful for the attributes to be printed after the dataframe?

R/as_palaeo.R
```diff
@@ -0,0 +1,118 @@
...
+#' @noRd
+resolve_info <- function(data, column) {
+  column_present_in_call <- column %in%
+    names(rlang::call_args(rlang::caller_call()))
+  value_from_attr <- attr(data, paste0("palaeo_", column))
+
+  if (isTRUE(column_present_in_call)) {
+    if (!is.null(value_from_attr)) {
+      cli::cli_inform(
+        "Overrode the data attribute {.val {paste0(\"palaeo_\", column)}}."
```
LewisAJones: Again, do we really need the "palaeo_" here? I think we should just use the attribute name to prevent confusion (i.e. "Latitude" here)?

Image

LewisAJones: Should we be more explicit? "User input argument detected. Overrode data attribute..."

LewisAJones (commented):
Just some initial comments @etiennebacher to promote further discussion around a few aspects. 

I'll want to review this again as the PR develops, including the documentation. However, I really like the direction this is going in and I think it'll be useful for individuals!

R/as_palaeo.R
etiennebacher: Yes, I think it is useful to be very explicit about which attributes come from `palaeoverse`, especially with generic names such as `lat` and `lon`. I've changed `palaeo_` to `palaeoverse_`.

For reference, some packages do the same, e.g. `sf::st_as_sf()` adds the attribute `sf_column`.

R/as_palaeo.R
etiennebacher: Changed to `palaeoverse_data`

R/as_palaeo.R
etiennebacher: A message containing `data attribute "Latitude"` while this attribute doesn't exist in the data would be more confusing IMO.

etiennebacher: I changed this to `` "Preferring user input `lat` over data attribute "palaeoverse_lat" ``, what do you think?

R/as_palaeo.R
willgearty: It should already have the data.frame class, so palaeoverse_data.frame seems overkill

R/as_palaeo.R
etiennebacher: We'll probably  want a helper but for now I'd rather discuss about the expected behaviour e.g. when the input is not explicitly passed before refactoring this part.

R/as_palaeo.R
willgearty: If a user is already using all of the default column names, they wouldn't need to use this function, right?

R/as_palaeo.R
willgearty: For the record, there wasn't really any context here and the title had been changed, so I assumed this was the plan for the "final" product. In the future, if you're looking for feedback on an early version of something or there's extra context for a PR, please make sure it's in the PR description (or post a comment).

R/as_palaeo.R
etiennebacher: > I agree being explicit is valuable but at the same time this could be quite annoying for users when registering many columns

True, but OTOH this has to be done only once and clearly sets the expected behaviour, hence limiting footguns. For instance, one could set `phylum = "phylum"` and if this column disappears from the data in the future then `as_palaeo()` will error, which will force the user to handle this situation. On the opposite, allowing implicit values (as in the current implementation) means that we wouldn't store the `phylum` attribute anymore but the user wouldn't notice that, potentially leading to errors that hard to debug further downstream.

PR #297, opened 2026-08-28
Remove default value `NULL` from mandatory arguments

### LINKED ISSUE #296: Remove default arguments of mandatory arguments
**The problem**
Function arguments that are mandatory should not have a default value. Currently, multiple functions use `NULL` as default value for a mandatory argument, and the function errors when the default value is kept.

**Proposed solution**
Remove the default `NULL` from the function signature. This should not break any functionality. If no value is passed for this argument in future versions, R will complain and throw a `argument "foo" is missing, with no default` error. This is a more robust solution than the unexpected failure with default arguments.

**Affected functions**
* `phylo_check(tree, list)`
* `tax_certainty(taxdf, name)`
* `tax_unique(occdf)`
* (please add other affected functions below)

NiklasHohmann:
## Removes default value NULL from required arguments

Solves #296 by removing NULL as default value for required arguments in `phylo_check`, `tax_certainty`, and `tax_unique`.

## Motivation and Context

See #296 

### DIFF
--- a/R/phylo_check.R
+++ b/R/phylo_check.R
@@ -73,19 +73,22 @@ phylo_check <- function(
   out = "full_table",
   sort = "presence"
 ) {
-
   #Errors for incorrect input
-  if(missing(tree)) stop("argument \"tree\" is missing, with no default")
+  if (missing(tree)) {
+    stop("argument \"tree\" is missing, with no default")
+  }
 
   if (!inherits(tree, "phylo")) {
     stop("Phylogeny must be a phylo object")
   }
 
+  if (missing(list)) {
+    stop("argument \"list\" is missing, with no default")
+  }
+
   if (!is.vector(list)) {
     stop("List of taxa must be a vector")
   }
-
-  if(missing(list)) stop("argument \"list\" is missing, with no default")
 
   if (any(grepl("[^[:alnum:][:space:]_]", list))) {
     stop(
--- a/R/phylo_check.R
+++ b/R/phylo_check.R
@@ -68,22 +68,15 @@
 #' par(oldpar)
 #' @export
 phylo_check <- function(
-  tree = NULL,
-  list = NULL,
+  tree,
+  list,
   out = "full_table",
   sort = "presence"
 ) {
   #Errors for incorrect input
-  if (is.null(tree)) {
-    stop("Phylogeny must be provided")
-  }
 
   if (!inherits(tree, "phylo")) {
     stop("Phylogeny must be a phylo object")
-  }
-
-  if (is.null(list)) {
-    stop("List of taxa to check against must be provided")
   }
 
   if (!is.vector(list)) {
--- a/R/phylo_check.R
+++ b/R/phylo_check.R
@@ -73,7 +73,9 @@ phylo_check <- function(
   out = "full_table",
   sort = "presence"
 ) {
+
   #Errors for incorrect input
+  if(missing(tree)) stop("argument \"tree\" is missing, with no default")
 
   if (!inherits(tree, "phylo")) {
     stop("Phylogeny must be a phylo object")
@@ -82,6 +84,8 @@ phylo_check <- function(
   if (!is.vector(list)) {
     stop("List of taxa must be a vector")
   }
+
+  if(missing(list)) stop("argument \"list\" is missing, with no default")
 
   if (any(grepl("[^[:alnum:][:space:]_]", list))) {
     stop(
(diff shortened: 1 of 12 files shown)


R/phylo_check.R
```diff
@@ -68,22 +68,22 @@
...
+  tree,
+  list,
   out = "full_table",
   sort = "presence"
 ) {
   #Errors for incorrect input
-  if (is.null(tree)) {
-    stop("Phylogeny must be provided")
+  if (missing(tree)) {
+    stop("argument \"tree\" is missing, with no default")
```
etiennebacher: This entire `if` block can be removed since R will already give this message by default if a mandatory argument isn't provided:

``` r
f <- function(x) {
  x + 1
}
f()
#> Error in `f()`:
#> ! argument "x" is missing, with no default
```

This can also be removed elsewhere.

tests/testthat/test-tax_certainty.R
```diff
@@ -1,3 +1,11 @@
+test_that("throws error for missing required arguments", {
+  expect_snapshot(tax_certainty(), error = TRUE)
+  data("tetrapods")
+  occdf <- tetrapods[1:5, ]
+  expect_snapshot(tax_certainty(taxdf = occdf), error = TRUE)
```
etiennebacher:
```suggestion
  expect_snapshot(tax_certainty(taxdf = tetrapods), error = TRUE)
```

Just to simplify a bit (we don't need to subset since this errors early anyway).

etiennebacher (changes requested):
Thanks @NiklasHohmann, this looks good to me. I just have a couple of comments below to simplify the code.

Can you also add a test for this change for `tax_unique()`?

R/phylo_check.R
NiklasHohmann: I had this initially, but ran into problems with lazy evaluation on some of the older R versions in CI (ubuntu-latest (oldrel-4). E.g., for `phylo_check`, the error is not thrown in the main function call, but in the first function call that uses the missing argument (in `phylo_check`, this is `inherits`), see checks on https://github.com/palaeoverse/palaeoverse/actions/runs/33149400087/job/98777655218
The current implementation is consistent across version and the error remains "shallow", meaning is thrown by the called function (as expected).

tests/testthat/test-tax_certainty.R
NiklasHohmann: Fair point

NiklasHohmann: fixed the associated snapshot test

NiklasHohmann:
Thanks for the review @etiennebacher
RE tests for `tax_unique`: The external behavior of the function has not changed, as the error text remained identical. The previous test accordingly covers the new implementation with `missing()`

R/phylo_check.R
etiennebacher: Yes, this is an issue that occurred with other error messages as well. The problematic error messages have been put after `skip_if(getRversion() < "4.3.0")` in the test suite so that they don't generate wrong snapshots on older versions. 

I think we should do the same here, it will be easy to remove  these `skip_if()` when the minimum R version is bumped in the future, but determining which `if` conditions become outdated in the code would be more annoying.

NiklasHohmann:
Test for `tax_unique()` is adjusted after removal of `missing()`.

etiennebacher (approved):
Thanks for addressing my comments!

willgearty (approved):
Looks good, thanks @NiklasHohmann!

R/tax_certainty.R
```diff
@@ -96,13 +96,14 @@
 #'                            append = FALSE)
 #' @export
 tax_certainty <- function(
-  taxdf = NULL,
-  name = NULL,
+  taxdf,
+  name,
```
LewisAJones, line 100: I think this is good for now, and is particular for this instance, but in most instances throughout the package, we use "genus" as the default input for name (i.e. `name = "genus"`). We might want to evaluate whether this should be standardised throughout the package.

LewisAJones (approved):
Thanks @NiklasHohmann, this looks good to me, just one comment I'll convert into an issue.

<s>As this impacts the user's interaction with the function, should we also update `NEWs.md`? </s> Etienne made the point to me that in reality, this doesn't really change anything to the user as it would have errored anyway.

R/tax_certainty.R
LewisAJones: Issue here: #301.

LewisAJones:
@NiklasHohmann Are you happy with this being merged now?

NiklasHohmann:
@LewisAJones sure! Are you merging?

PR #298, opened 2026-08-28
Change default value of `round` in `axis_geo()` to `NULL`

NiklasHohmann:
## Description

The `round` argument of `axis_geo` previously was set to a logical value. It is now changed to `NULL` and the description states clearly that the expected input is a numeric value.

## Motivation and Context

This keeps Booleans for strictly logical inputs, and NULL as (yet unspecified) default value.

- [x] I have updated function documentation.

### DIFF
--- a/R/axis_geo.R
+++ b/R/axis_geo.R
@@ -825,11 +825,12 @@ axis_geo <- function(
     if (phylo) {
       tick_at <- (tick_at - phylo_alpha) / phylo_beta
     }
-    if (is.numeric(round)) {
-      tick_labels <- round(tick_at, digits = round)
-    } else {
+    if (is.null(round)){
       tick_labels <- as.character(tick_at) # removes trailing zeros
+    } else {
+      tick_labels <- round(tick_at, digits = round)
     }
+
   }
   if (phylo) {
     if (is.null(tick_at)) {
--- a/R/axis_geo.R
+++ b/R/axis_geo.R
@@ -129,10 +129,10 @@
 #'   marks and numeric tick labels placed at the interval boundaries. If
 #'   \code{TRUE}, this overrides \code{tick_at} and \code{tick_labels}.
 #' @param round \code{integer}. Number of decimal places to which exact axis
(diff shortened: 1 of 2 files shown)


NiklasHohmann:
Function behavior remains unchanged

R/axis_geo.R
```diff
@@ -826,7 +826,7 @@ axis_geo <- function(
       tick_at <- (tick_at - phylo_alpha) / phylo_beta
     }
     if (is.numeric(round)) {
-      tick_labels <- round(tick_at, round)
+      tick_labels <- round(tick_at, digits = round)
     } else {
       tick_labels <- as.character(tick_at) # removes trailing zeros
     }
```
etiennebacher: Shouldn't this entire if-else be replaced by `tick_labels <- round(tick_at, digits = round)`?

etiennebacher (changes requested):

R/axis_geo.R
NiklasHohmann: `round()` throws an error if the digits argument is not numeric, so passing `NULL` directly to it does not work. A one-line solution would be
```tick_labels <- as.character(round(tick_at, digits = ifelse(is.numeric(round), round, Inf)))```
but I'd argue that's not adding clarity.

etiennebacher: Hmm, I see, thanks. Maybe we can just replace the `is.numeric(round)` by `!is.null(round)`. That way, this will error if the user passes `round = "foo"` instead of silently going to the `else` branch. (We could also check early in the function that `round` is a numeric vector of length 1)

NiklasHohmann: I changed the branching order. In the new setting, the code first checks for `NULL` to then move to take inputs at face value as tick labels, else it passes `round` to `round()` and lets this function handle potentially malformed inputs. This allows to e.g., round different axis labels to different precision.
This is more intentional input handling with no silent fallbacks.

etiennebacher (approved):

willgearty (approved):
Looks good to me, thanks @NiklasHohmann!

PR #303, opened 2026-09-02
Check that args are named in `bin_lat()`

etiennebacher:
Part of #216 

### DIFF
--- a/R/bin_lat.R
+++ b/R/bin_lat.R
@@ -1,6 +1,9 @@
 #' Assign fossil occurrences to latitudinal bins
 #'
+#' @description
 #' A function to assign fossil occurrences to user-specified latitudinal bins.
+#'
+#' All arguments must be named.
 #'
 #' @param occdf `dataframe`. A dataframe of the fossil occurrences you
 #' wish to bin. This dataframe should contain a column with the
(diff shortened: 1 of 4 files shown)


R/bin_lat.R
```diff
@@ -1,7 +1,10 @@
 #' Assign fossil occurrences to latitudinal bins
 #'
+#' @description
 #' A function to assign fossil occurrences to user-specified latitudinal bins.
 #'
+#' All arguments must be named.
+#'
```
etiennebacher: I don't know if this is the best place to document this (keeping in mind that this would be added to almost all reference pages).

willgearty: Yeah, I'm not a huge fan of this being listed on every function's doc page. Since it's a general package rule, it should go somewhere one level up (readme, vignette, etc).

willgearty (approved):

R/bin_lat.R
LewisAJones: I agree that this doesn't need to be in every function, and probably should just be in an "Introduction to palaeoverse" vignette... I think after hitting the error once or twice, users will click on too.

There are some additional high-level rules/information like this that the user might want to have access to. For example, the eventual parameter dictionary? Perhaps again for a vignette.

LewisAJones (approved):

R/bin_lat.R
etiennebacher: Ok, I reverted this change

PR #311, opened 2026-09-02
Check that args are named in `phylo_check()`

etiennebacher:
This PR and subsequent ones that implement `ensure_args_are_named()` are co-authored by Claude Opus 5.

Part of #216 

### DIFF
--- a/R/phylo_check.R
+++ b/R/phylo_check.R
@@ -73,6 +73,8 @@ phylo_check <- function(
   out = "full_table",
   sort = "presence"
 ) {
+  ensure_args_are_named()
+
   #Errors for incorrect input
   if (!inherits(tree, "phylo")) {
     stop("Phylogeny must be a phylo object")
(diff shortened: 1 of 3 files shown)


willgearty (approved):

R/phylo_check.R
```diff
@@ -73,6 +73,8 @@ phylo_check <- function(
   out = "full_table",
   sort = "presence"
 ) {
+  ensure_args_are_named(exceptions = c("tree", "list"))
```
LewisAJones: In general, I'm not sure how I feel about having multiple exceptions... 

What is our hard rule for multiple exceptions? This should be clear and predictable to users.

etiennebacher: I don't think we have defined a rule for that, this is the first time we have this case. I'm fine with removing `list` from the exceptions.

LewisAJones: I think for now I would prefer just the first argument being the exception unless we see a clear need for this, and we define a rule for such instances.

LewisAJones (approved):
Thanks!

PR #317, opened 2026-09-02
Check that args are named in `bin_time()`

etiennebacher:
Part of #216 

### DIFF
--- a/R/bin_time.R
+++ b/R/bin_time.R
@@ -123,6 +123,8 @@ bin_time <- function(
   fun = dunif,
   ...
 ) {
+  ensure_args_are_named()
+
   check_data_frame(occdf)
   check_data_frame(bins)
 
--- a/R/bin_time.R
+++ b/R/bin_time.R
@@ -123,7 +123,7 @@ bin_time <- function(
   fun = dunif,
   ...
 ) {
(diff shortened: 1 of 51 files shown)


R/bin_time.R
```diff
@@ -123,6 +123,8 @@ bin_time <- function(
   fun = dunif,
   ...
 ) {
+  ensure_args_are_named(exceptions = "occdf")
+
```
etiennebacher, line 127: We could argue about whether `bins` should be unnamed as well but this function is going to be heavily refactored anyway so I chose not to add it here.

tests/testthat/test-bin_time.R
```diff
@@ -364,7 +371,7 @@ test_that("wrong input for occdf", {
 
   # dataframe that doesn't have the expected columns
   expect_snapshot(
-    bin_time(mtcars, occdf = c(50, 20, 10)),
+    bin_time(bins = mtcars, occdf = c(50, 20, 10)),
```
etiennebacher, line 374: So that the previous error message that is tested isn't overwritten by "Argument is unnamed"

LewisAJones (approved):

willgearty (approved):

PR #318, opened 2026-09-02
Use `cli` and `rlang` in `tax_unique()`

etiennebacher:

### DIFF
--- a/R/tax_unique.R
+++ b/R/tax_unique.R
@@ -160,9 +160,7 @@ tax_unique <- function(
   higher_cols <- unname(unlist(higher_args))
   if (length(higher_args) == 0) {
     cli::cli_abort(
-      c(
-        "At least one higher taxonomic level must be supplied (e.g. {.code family = \"family\"})."
-      )
+      "At least one higher taxonomic level must be supplied (e.g. {.code family = \"family\"})."
     )
   }
 
--- a/R/tax_unique.R
+++ b/R/tax_unique.R
@@ -154,6 +154,9 @@ tax_unique <- function(
   if (!is.null(genus)) {
     check_column_presence(occdf, genus)
   }
+  if (!is.null(name)) {
+    check_column_presence(occdf, name)
+  }
 
   higher_args <- list(...)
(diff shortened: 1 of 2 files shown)


R/tax_unique.R
```diff
@@ -205,62 +180,57 @@ tax_unique <- function(
         occdf[[col_name]]
       )
     if (any(grepl("[[:punct:]]", occdf[[col_name]]))) {
-      stop(paste0("`", level_label, "` column should not contain punctuation"))
+      cli::cli_abort(
+        "Column {.val {level_label}} must not contain punctuation."
+      )
```
etiennebacher, line 185: This looks like it could be a helper `check_punctuation()` but I'm not convinced because sometimes we allow spaces and underscores and sometimes not. We only check for punctuation once outside of this function (in `phylo_check()` ) so I don't think it warrants a helper.

willgearty (approved):
Looks good, thanks!

R/tax_unique.R
LewisAJones: I also don't think this is needed right now but I imagine as the package expands this may be useful 👍🏻

R/tax_unique.R
```diff
@@ -205,62 +180,57 @@ tax_unique <- function(
...
+      is.null(name)
   ) {
-    stop(
-      "Species names must be supplied by specifying `binomial`, `genus` and
-    `species`, or `genus` and `name` columns to estimate richness at species
-    level"
+    cli::cli_abort(
+      c(
+        "Species names must be supplied to estimate richness at species level.",
+        "i" = "Specify {.arg binomial}, or {.arg genus} and {.arg species}, or {.arg genus} and {.arg name}."
```
LewisAJones:
```suggestion
        "i" = "Specify {.arg binomial}, {.arg genus} and {.arg species}, or {.arg genus} and {.arg name}."
```

I think this better represents the original error message?

LewisAJones (approved):
Looks good, just one small comment! 

Thanks!

PR #320, opened 2026-09-03
Allow some arguments to be unnamed

### LINKED ISSUE #319: Piping doesn't work when all args must be named
piping is now broken after implementing https://github.com/palaeoverse/palaeoverse/issues/216 (requiring all arguments to be named). I propose either a) (if possible) allow piped variables to not be named or b) allow required arguments to not be named.

etiennebacher:
This PR allows passing unnamed arguments, which is particularly useful when using the pipe. The only function that has this check for now is `lat_bins_area()` and I don't think we want any unnamed argument (no one would do `12 |> lat_bins_area(min = 90, ...)`, but I checked locally that this works correctly in `bin_lat()`.

Fixes https://github.com/palaeoverse/palaeoverse/issues/319

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -1,11 +1,14 @@
 #' Check whether there are unnamed arguments that should be named.
 #'
-#' This errors if some arguments are unnamed or if their names are partially
-#' matched.
+#' This errors if some arguments (apart from `exceptions`) are unnamed or
+#' if their names are partially matched.
+#'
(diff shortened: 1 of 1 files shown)


willgearty (approved):
Looks good!

LewisAJones (approved):
Looks good, thanks for the quick fix @etiennebacher!

PR #324, opened 2026-09-03
Use `cli` and `rlang` in `tax_expand_time()`

etiennebacher:

### DIFF
--- a/R/tax_expand_time.R
+++ b/R/tax_expand_time.R
@@ -63,67 +63,55 @@ tax_expand_time <- function(
   rank = "stage",
   ext_orig = TRUE
 ) {
-  # Handle errors
-  if (!is.data.frame(taxdf)) {
-    stop("`taxdf` should be a dataframe")
-  }
+  check_data_frame(taxdf)
 
-  if (!all(c(min_ma, max_ma) %in% colnames(taxdf))) {
-    stop("Either `min_ma` or `max_ma` is not a named column in `taxdf`")
-  }
+  check_column_presence(taxdf, max_ma)
+  check_column_presence(taxdf, min_ma)
 
-  if (!is.numeric(taxdf[, max_ma, drop = TRUE])) {
-    stop("The class of the max_ma column must be numeric.")
-  }
-  if (!is.numeric(taxdf[, min_ma, drop = TRUE])) {
-    stop("The class of the min_ma column must be numeric.")
-  }
+  check_range(taxdf, min_ma, 0, Inf)
+  check_range(taxdf, max_ma, 0, Inf)
 
-  if (any(taxdf[, c(min_ma, max_ma)] < 0)) {
-    stop("Maximum and minimum ages must be positive.")
-  }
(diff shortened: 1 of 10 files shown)


R/utils-rlang.R
```diff
@@ -66,70 +48,23 @@ check_range <- function(data, column, min, max) {
     to_report <- unique(rng)
     truncated <- if (length(to_report) > 5) " (first 5)" else ""
     to_report <- cli::cli_vec(head(to_report, n = 5), list(`vec-last` = ", "))
+
+    msg_range <- if (min == 0 && is.infinite(max)) {
+      "positive"
+    } else {
+      "between {min} and {max}"
+    }
```
etiennebacher: This feels a bit hacky but without this, the message would be `` All values of column "min_ma" in `taxdf` must be between 0 and Inf. ``, which looks worse to me.

R/tax_expand_time.R
```diff
@@ -63,67 +63,54 @@ tax_expand_time <- function(
...
+    check_column_presence(bins, "min_ma")
   }
 
-  if (is.null(bins) && (is.null(scale) || is.null(rank))) {
-    stop("Either `bin` or `scale` and `rank` must be specified.")
-  }
-  if (!is.null(bins)) {
-    if (!is.data.frame(bins)) {
-      stop("`bins` should be a dataframe.")
+  rows_with_max_ma_smaller_than_min_ma <- which(
```
LewisAJones: This specific error checking should probably be present in almost every function where we have the arguments `max_ma` and `min_ma`... perhaps worth considering whether we should have an internal function for this too (sorry @etiennebacher!)... I believe we have similar pre-existing code in `bin_time`.

LewisAJones (approved):
Just one comment... I'll leave this to you whether you want to implement now, discuss, or come back to it later (if later, please create an issue).

Otherwise, looks good!

R/tax_expand_time.R
```diff
@@ -63,67 +63,54 @@ tax_expand_time <- function(
...
-  }
-
-  if (
-    length(rank) != 1 ||
-      !(rank %in% c("stage", "epoch", "period", "era", "eon"))
-  ) {
-    stop("`rank` must be either: stage, epoch, period, era, or eon")
-  }
+  rlang::check_bool(ext_orig)
+  rlang::check_string(rank)
```
willgearty: This should probably go down below in the `is.null(bins)` check alongside the `scale` check

willgearty (approved):
Just one small comment; good to go once that's addressed

R/tax_expand_time.R
etiennebacher: I added a new `check_min_lower_than_max()` in this function as well as in `bin_time()` and `tax_range_time()` (+ some tests for those), let me know if you're happy with that.

willgearty (approved):

LewisAJones (approved):
Thanks for the changes! Looks good 😄

PR #326, opened 2026-09-03
Use `cli` and `rlang` in `phylo_check()`

etiennebacher:

### DIFF
--- a/R/phylo_check.R
+++ b/R/phylo_check.R
@@ -73,47 +73,42 @@ phylo_check <- function(
   out = "full_table",
   sort = "presence"
 ) {
-  #Errors for incorrect input
   if (!inherits(tree, "phylo")) {
-    stop("Phylogeny must be a phylo object")
-  }
-
-  if (!is.vector(list)) {
-    stop("List of taxa must be a vector")
-  }
-
-  if (any(grepl("[^[:alnum:][:space:]_]", list))) {
-    stop(
-      "Taxon names should not contain punctuation except spaces or
-         underscores"
+    cli::cli_abort(
+      "{.arg tree} must be of class {.cls phylo}, not {obj_type_friendly(tree)}."
     )
   }
 
-  if (length(out) == 0 || is.na(out)) {
-    stop("`out` must be of length 1.")
+  if (!is.vector(list)) {
+    cli::cli_abort(
+      "{.arg list} must be a vector, not {obj_type_friendly(list)}."
+    )
   }
 
-  if (length(sort) == 0 || is.na(sort)) {
-    stop("`sort` must be of length 1.")
+  bad_names <- unique(list[grepl("[^[:alnum:][:space:]_]", list)])
+  if (length(bad_names) > 0) {
+    truncated <- if (length(bad_names) > 5) " (first 5)" else ""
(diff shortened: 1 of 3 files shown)


willgearty (approved):

R/phylo_check.R
```diff
@@ -73,47 +73,42 @@ phylo_check <- function(
...
+    values = c("full_table", "diff_table", "counts", "tree")
+  )
 
-  if (sort != "az" && sort != "presence") {
-    stop("sort must either be 'az' or 'presence'")
-  }
+  rlang::check_string(sort)
+  sort <- rlang::arg_match(sort, values = c("presence", "az"))
 
   if (out != "full_table" && out != "diff_table" && sort != "presence") {
```
LewisAJones, line 109: I realise this isn't induced by you, but is the check for `sort` here not redundant?

R/phylo_check.R
```diff
@@ -73,47 +73,42 @@ phylo_check <- function(
...
     )
   }
 
-  if (length(out) == 0 || is.na(out)) {
-    stop("`out` must be of length 1.")
-  }
-
-  if (length(sort) == 0 || is.na(sort)) {
-    stop("`sort` must be of length 1.")
+  bad_names <- unique(list[grepl("[^[:alnum:][:space:]_]", list)])
```
LewisAJones, line 88: Should this also allow periods (`.`)? 

How common would it be to come across instances such as `Genus sp.` in trees @willgearty @bethany-j-allen?

R/phylo_check.R
```diff
@@ -73,47 +73,42 @@ phylo_check <- function(
...
   if (!is.vector(list)) {
-    stop("List of taxa must be a vector")
-  }
-
-  if (any(grepl("[^[:alnum:][:space:]_]", list))) {
-    stop(
-      "Taxon names should not contain punctuation except spaces or
-         underscores"
+    cli::cli_abort(
+      "{.arg list} must be a vector, not {obj_type_friendly(list)}."
```
LewisAJones, line 84: Not for this PR, but see #332

LewisAJones (approved):

R/phylo_check.R
etiennebacher: Do you mean `rlang::check_string(sort)`? If so, no it's not redundant, it's because `rlang` gives a bad error message if the user passes a character vector of length > 1:

``` r
foo <- function(x) {
  rlang::arg_match(x, c("a", "b", "c"))
}

foo(c("e", "f"))
#> Error in `foo()`:
#> ! `x` must be one of "a", "b", or "c", not "e".
```

Already tracked in https://github.com/r-lib/rlang/issues/1838

R/phylo_check.R
etiennebacher: I extracted this to https://github.com/palaeoverse/palaeoverse/issues/333 since it is about changing the behaviour and not about the `rlang` / `cli` usage.

R/phylo_check.R
LewisAJones: My comment was for L109, not sure why it has highlighted 106 and 107! Sorry!

etiennebacher: Ah yes I see now, indeed this ` && sort != "presence"` should probably be removed

PR #331, opened 2026-09-03
Use `cli` and `rlang` in `tax_range_strat()`

etiennebacher:

### DIFF
--- a/R/tax_range_strat.R
+++ b/R/tax_range_strat.R
@@ -129,55 +129,29 @@ tax_range_strat <- function(
   x_args = NULL,
   y_args = NULL
 ) {
-  if (!is.data.frame(occdf)) {
-    stop("`occdf` should be a data.frame")
-  }
+  check_data_frame(occdf)
+  check_column_presence(occdf, name)
+  check_column_presence(occdf, level)
 
-  if (!is.numeric(occdf[, level, drop = TRUE])) {
-    stop("`level` must be of class numeric")
+  if (!is.numeric(occdf[[level]])) {
+    cli::cli_abort(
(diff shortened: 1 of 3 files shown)


R/tax_range_strat.R
```diff
@@ -129,55 +129,29 @@ tax_range_strat <- function(
...
+  check_column_presence(occdf, name)
+  check_column_presence(occdf, level)
 
-  if (!is.null(group) && length(group) != 1) {
-    stop("`group` must be of length 1.")
-  }
-
-  if (!is.null(group) && !group %in% colnames(occdf)) {
-    stop("`group` is not a named column in `occdf`")
+  if (!is.numeric(occdf[[level]])) {
```
LewisAJones: Now we're implementing all of these checks throughout the package, it's probably good to pull such checks out into generalised functions. For example, this could be `check_class` or similar.

R/tax_range_strat.R
```diff
@@ -129,55 +129,29 @@ tax_range_strat <- function(
...
+    cli::cli_abort(
+      "Column {.val {level}} in {.arg occdf} must be of class {.cls character}, not {.cls {class(occdf[[level]])}}."
```
LewisAJones: Is this perhaps confusing? The name of the column needs to be specified as a character, but the values themselves should be numeric.

LewisAJones (changes requested):
Thanks @etiennebacher, a couple of comments!

R/tax_range_strat.R
etiennebacher: Yes this was a mistake

LewisAJones (approved):

willgearty (approved):

PR #335, opened 2026-09-07
More robust condition in `phylo_check()`

etiennebacher:
Follow-up of https://github.com/palaeoverse/palaeoverse/pull/326#discussion_r3948349924

### DIFF
--- a/R/phylo_check.R
+++ b/R/phylo_check.R
@@ -106,7 +106,7 @@ phylo_check <- function(
   rlang::check_string(sort)
   sort <- rlang::arg_match(sort, values = c("presence", "az"))
 
-  if (out != "full_table" && out != "diff_table" && sort != "presence") {
+  if (out != "full_table" && out != "diff_table") {
     cli::cli_warn(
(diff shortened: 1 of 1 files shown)


R/phylo_check.R
```diff
@@ -103,10 +103,13 @@ phylo_check <- function(
...
   )
 
+  # Define this before the check, otherwise `missing(sort)` always returns `FALSE`
+  missing_sort <- missing(sort)
+
   rlang::check_string(sort)
   sort <- rlang::arg_match(sort, values = c("presence", "az"))
 
-  if (out != "full_table" && out != "diff_table" && sort != "presence") {
+  if (out != "full_table" && out != "diff_table" && !missing_sort) {
```
etiennebacher, line 112: `sort != "presence"` was here so that we throw this warning only if `sort` is explicitly provided, the idea being that the user wouldn't set `sort = "az"` in the function since this is the default. Checking directly whether `sort` is explicitly passed by the user is more robust.

willgearty (approved):

LewisAJones (approved):

PR #337, opened 2026-09-08
Various fixes and improvements for error messages

etiennebacher:
Collection of small improvements now that several functions use `rlang` and `cli`:

- [x] extract class check from `check_range()`
- [x] add formatting rules
- [x] determine whether to use `obj_type_friendly()` 
- [x] use `rlang::check_number_*()` where it is better than `check_numeric()` (e.g. `bin_space()` and `bin_time()`)
- [x] use `{.code NA}` instead of `NA` everywhere
- [x]  `` list(`vec-last` = ", ") `` -> `` list(`vec-last` = ", ", `vec-sep2` = ", ") `` 
- [x] wrap in backticks instead of quotation marks in messages `All arguments must be named`
- [x] in `bin_time()`, `reps` should be >= 1 (+ add some tests for that)
- [x] dedicated test file for `utils-rlang`


R/phylo_check.R
```diff
@@ -90,7 +90,10 @@ phylo_check <- function(
   bad_names <- unique(list[grepl("[^[:alnum:][:space:]_]", list)])
   if (length(bad_names) > 0) {
     truncated <- if (length(bad_names) > 5) " (first 5)" else ""
-    bad_names <- cli::cli_vec(head(bad_names, n = 5), list(`vec-last` = ", "))
+    bad_names <- cli::cli_vec(
+      head(bad_names, n = 5),
+      list(`vec-last` = ", ", `vec-sep2` = ", ")
```
etiennebacher, line 95: `vec-sep2` specifies the separator when there are only 2 elements, `vec-last` does the same for > 2 elements.

tests/testthat/test-utils-rlang.R
```diff
@@ -0,0 +1,14 @@
...
+  expect_snapshot(check_class(dat, "x", "numeric"))
+
+  x <- "a"
+  class(x) <- c("foo", "bar", class(x))
+  dat <- data.frame(x = x)
+  expect_snapshot(
+    check_class(dat, "x", "numeric"),
+    error = TRUE
+  )
+})
```
etiennebacher, line 14: New test file because:

1. I didn't want to add a test for this scenario in all functions that use `check_class()`
2. I didn't want to edit just one one of the existing test files because this new test could be removed by mistake in the future without being detected by code coverage

R/utils-rlang.R
willgearty: I think this needs to be removed?

willgearty (approved):
One comment, otherwise looks good.

R/utils-rlang.R
etiennebacher: Yes, good catch

R/bin_space.R
```diff
@@ -122,11 +122,14 @@ bin_space <- function(
...
   check_column_presence(occdf, lng)
+
+  check_class(occdf, lat, "numeric")
+  check_class(occdf, lng, "numeric")
   check_range(occdf, lat, -90, 90)
   check_range(occdf, lng, -180, 180)
 
-  check_numeric(spacing, required_length = 1)
-  check_numeric(sub_grid, allow_null = TRUE, required_length = 1)
+  rlang::check_number_decimal(spacing, min = 0)
```
LewisAJones: `spacing = 0` is not actually possible and results in an unhelpful error message?

I guess this should be up to the user, but I'll note that when spacing is a low number (e.g. 10), it's prone to crash R even on a decent machine.

R/bin_space.R
```diff
@@ -122,11 +122,14 @@ bin_space <- function(
...
+  rlang::check_number_decimal(sub_grid, min = 0, allow_null = TRUE)
```
LewisAJones: Same again here.

R/phylo_check.R
LewisAJones: In general, if it warrants a comment here, it's probably worth adding as a line comment too?

R/utils-rlang.R
```diff
@@ -1,3 +1,15 @@
+# Error message formatting rules:
+#
+# - argument names are wrapped in backticks, e.g. `dis` should be greater than 1
```
LewisAJones, line 3: 🎉  Appreciate this! Thanks!

LewisAJones (changes requested):
Looks good, thanks @etiennebacher! Very satisfying to see this harmonisation. A couple of small suggested changes.

R/bin_space.R
etiennebacher: Yeah I thought about that but technically we allow small positive values so I wasn't confident in setting a threshold. Setting the minimum to 10 also seems arbitrary and doesn't really improve the error message IMO. On the other hand, the current error message for `spacing = 0` isn't great:

```r
> bin_space(occdf = tetrapods, spacing = 0)
Error:
! Error: Array length out of bounds (code: 1001, value: 4747561509943)
```

So I think I'll just add a custom message in this case.

etiennebacher: [`f80c7eb`](https://github.com/palaeoverse/palaeoverse/pull/337/changes/f80c7eb51307964b3d5c4a0f1d548bf870908bfd)

R/phylo_check.R
etiennebacher: It's used in multiple places in the code and I don't think we want a comment everywhere. It was just to simplify the review by avoiding a trip to the `cli` docs.

LewisAJones (approved):

PR #339, opened 2026-09-08
Check that args are named in `tax_unique()`

etiennebacher:
Part of #216 

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -15,7 +15,14 @@ ensure_args_are_named <- function(exceptions = NULL) {
   )]
 
   args_in_function_def <- names(formals(rlang::caller_fn()))
-  partially_matched_names <- setdiff(named_args, args_in_function_def)
+
+  # Dots could be passed to other functions and we can't check whether they are partially
+  # named.
+  if ("..." %in% args_in_function_def) {
+    partially_matched_names <- NULL
+  } else {
+    partially_matched_names <- setdiff(named_args, args_in_function_def)
+  }
   if (length(partially_matched_names) > 0) {
     cli::cli_abort(
       c(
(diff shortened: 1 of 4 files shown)


willgearty (approved):

R/utils.R
```diff
@@ -8,14 +8,35 @@
 #' @noRd
 ensure_args_are_named <- function(exceptions = NULL) {
   args_in_call_from_user <- rlang::call_args_names(rlang::caller_call())
+
+  # When the function that is being checked is used internally, the call may contain
```
LewisAJones, line 12: Does this imply that someone could do the following?

```
group_apply(occdf = occdf, group = "collection_no", fun = tax_unique, "genus", "family", "order", "class", "genus")
```

Do we want that?

etiennebacher: No it doesn't imply that. This addition is for the case where the call explicitly contains `...`, such as this block in `group_apply()`:

```r
  output_lst <- by(
    data = occdf,
    INDICES = occdf[, group, drop = FALSE],
    FUN = fun,
    ...
  )
```
In this case, `rlang::call_args(rlang::caller_call())` would report that the name of `...` is `""` and `...` would be counted as an unnamed argument.


`...` is not explicitly passed to `group_apply(occdf = occdf, group = "collection_no", fun = tax_unique, "genus", "family", "order", "class", "genus")` so this would report all unnamed args (see the PR for `group_apply()`, which depends on this one: https://github.com/palaeoverse/palaeoverse/pull/340/changes#diff-1db8efcc25ed078e93a87443c21c3e40f3d7a2e3a9aa3d0a279b9af010273b0a)

LewisAJones (approved):

Issue #162, opened 2026-05-11
Should we put `h3jsr` in `Suggests`?


etiennebacher:
This pulls a lot of dependencies but is used in a couple of functions only, should we put it in `Suggests` and add a `check_installed()`? Or maybe even replace it entirely with an alternative with less deps, if there are any (I'm not familiar with this type of packages yet).

```r
deepdep::plot_dependencies("palaeoverse", depth = 3)
```

<img width="826" height="512" alt="Image" src="https://github.com/user-attachments/assets/51ceb226-fe22-44b8-a397-4b4c82b7ea3c" />

willgearty:
I think moving to `Suggests` is a good idea. It's the only dependency that currently requires compilation, which has caused installation issues in the past (https://github.com/palaeoverse/palaeoverse/issues/154).

I think we've even discussed how necessary `palaeorotate()` is now that there is the rgplates package (although rgplates doesn't have the hexagon functionality, to my knowledge). That's probably too far, but given the package is only needed for a single branch of the `palaeorotate()` function, it seems like a lot to enforce as an `Imports`.

LewisAJones:
Good question, and great way of visualising this @etiennebacher. 

Installation issues have generally come up when individuals do not have V8 installed (https://github.com/palaeoverse/palaeoverse/issues/154). Installing V8 separately has largely addressed these issues in the past (but this is not ideal). I'm not sure if there is an alternative to h3jsr which doesn't require V8 but it would also be good to find an alternative which is not dependent on tidyr.

While extensive functionality does exist in rgplates for reconstructing palaeogeographic coordinates, I think `palaeorotate()` still offers many advantages and I'd be against removing it as a function:

- In my opinion, it is much simpler and intuitive to use, capable of directly taking occurrence dataframes.
- Multiple models can be called at the same time, and allows the calculation of differences between model outputs.
- It offers access to the pre-rotated grids, which are well documented, static (improving reproducibility), and offer a fast option for larger datasets.  

Note, in addition to `palaeorotate()`, h3jsr is also used in `bin_space()` so I think at this point in time, it probably needs to remain as an `Imports` but we should look for an alternative?

PR #344, opened 2026-09-09
Implement `space_bins()` and refactor `bin_space()`

### LINKED ISSUE #354: Refactor `bin_space()` and add `space_bins()`
`bin_space()` has a few issues:

- it creates bins and assign occurrences to them, which doesn't match the other binning functions (e.g. `time_bins()` to create the bins and `bin_time()` to assign occurrences to bins)
- its output type depends on the `return` parameter (data.frame if `return = FALSE`, a list with three elements if `return = TRUE`)

We decided to implement a new function  `space_bins()`:

- this has an argument `spacing`  that can take one numeric value (i.e. users who want to bin several times create several bins with `space_bins()` and then chain several `bin_space()`)
-  this returns the current  `grid_base` returned by `bin_space()` (i.e. we don't have access to `grid_info` anymore)
- its output has the class `"palaeo_space_bins"` and keeps the `spacing` info in an attribute

In `bin_space()`:

- remove `return`, `spacing`, and `sub_grid` arguments
- new argument `bins` that can take the output of `space_bins()`
- `bin_space()` checks the class, extracts the attribute, and uses this info as suffix of column names, e.g. `cell_lat_500` and `cell_lon_500`

etiennebacher:
Fixes #354

This adds `space_bins()` and removes arguments `return`, `spacing`, and `sub_grid` from `bin_space()`. I'm mostly unsure about whether `bins` should only accept the output of `space_bins()` or if we should allow custom inputs. The challenge is that currently this PR uses `bins` attributes in `bin_space()` to get info such as the resolution or spacing, so if we allow custom inputs then we need to find a way for users to pass this information as well.


tests/testthat/test-bin_space.R
```diff
@@ -67,36 +115,53 @@ test_that("bin_space error handling", {
...
   occdf <- tetrapods
   occdf$lng[1] <- 184
-  expect_snapshot(bin_space(occdf = occdf), error = TRUE)
+  expect_snapshot(bin_space(occdf, space_bins(1000)), error = TRUE)
   occdf$lng[1] <- "184"
-  expect_snapshot(bin_space(occdf = occdf), error = TRUE)
+  expect_snapshot(bin_space(occdf, space_bins(1000)), error = TRUE)
+})
+
+test_that("plot argument works", {
```
etiennebacher, line 146: This was untested before but it will be helpful to have this when extracting the `plot` argument in a future PR.

R/bin_space.R
```diff
@@ -68,20 +62,27 @@
...
+#'   bin_space(bins = space_bins(1000)) |>
+#'   bin_space(bins = space_bins(250))
+#'
+#' head(ex2)
 #'
 #' # EXAMPLE: rarefy
 #' # Load data
 #' occdf <- tetrapods[1:250, ]
 #'
 #' # Assign to spatial bin
```
willgearty:
```suggestion
#' # Assign to spatial bins and sub-bins
```

R/bin_space.R
willgearty: here we call them "cells", but elsewhere we call them "bins". Should we be consistent?

willgearty: (I suppose this even applies to how we name columns, arguments, and even the function itself)

R/bin_space.R
```diff
@@ -90,36 +91,38 @@
...
 #' })
+#' df
 #' @export
-bin_space <- function(
-  occdf,
-  lng = "lng",
-  lat = "lat",
-  spacing = 100,
-  sub_grid = NULL,
-  return = FALSE,
```
willgearty: Should these be deprecated for the time being?

R/bin_space.R
```diff
@@ -148,83 +142,36 @@ bin_space <- function(
...
-  # Which resolution should be used based on input distance/spacing?
-  # Use the h3jsr::h3_info_table to calculate resolution
-  grid <- h3jsr::h3_info_table[
-    which.min(abs(h3jsr::h3_info_table$avg_cendist_km - spacing)),
-  ]
-  # Add column grid specification
-  grid$grid <- c("primary")
+  spacing <- attr(bins, "spacing", exact = TRUE)
+  h3_resolution <- attr(bins, "h3_resolution", exact = TRUE)
+  avg_cendist_km <- attr(bins, "avg_cendist_km", exact = TRUE)
```
willgearty: I feel like we should accept any arbitrary grid, not just those from `h3jsr` (and the new docstring and error message both seem to imply this). Users should be able to make their own grid cells (as a `sfc_POLYGON` object) and still apply this function. That probably means this function would just become a wrapper for some `sf` functions instead of a wrapper for `h3jsr` functions? (also see https://github.com/palaeoverse/palaeoverse/issues/162 and https://github.com/palaeoverse/palaeoverse/pull/345)

R/bin_space.R
```diff
@@ -233,33 +180,19 @@ bin_space <- function(
...
       paste0(
         "Average spacing between adjacent cells in the primary grid was set to ",
-        round(grid$avg_cendist_km[1], digits = 2),
+        round(avg_cendist_km[1], digits = 2),
         " km. "
       ),
-      "i" = paste0("\nH3 resolution: ", grid$h3_resolution[1])
+      "i" = paste0("\nH3 resolution: ", h3_resolution[1])
     )
   )
```
willgearty: Not sure it makes sense to show this for this function anymore. Should be moved to `space_bins`, I think?

R/bin_space.R
```diff
@@ -148,83 +142,36 @@ bin_space <- function(
...
-  # Add column grid specification
-  grid$grid <- c("primary")
+  spacing <- attr(bins, "spacing", exact = TRUE)
+  h3_resolution <- attr(bins, "h3_resolution", exact = TRUE)
+  avg_cendist_km <- attr(bins, "avg_cendist_km", exact = TRUE)
 
+  #=== Grid binning  ===
   # Extract cell ID
-  occdf$cell_ID <- h3jsr::point_to_cell(occdf, res = grid$h3_resolution)
+  cell_name <- paste0("cell_ID_", spacing)
```
willgearty: I worry that this could be confusing to users/readers because it is the value they supplied to `space_bins`, not the actual spacing in km (which could be very different). Same goes for the other column names.

willgearty: if we do allow any bin input, how would we craft these column names...

NEWS.md
```diff
@@ -12,6 +12,16 @@ guidance to update existing code.
...
 * In `lat_bins_area()`, the argument `n` is deprecated and replaced by `n_bins` (#349).
 
+* In `bin_space()`, the arguments `spacing`, `sub_grid`, and `return` have been
+  removed. There is a new `bins` argument. The same functionality is made possible with
+  the new `space_bins()` (see "New features") (#344).
+
+## New features
+
+* New `space_bins()` to generate spatial bins, as a complement to `time_bins()` and 
+  `lat_bins_*()` functions. The output of this function can be passed to the `bins`
```
willgearty: Not sure this is necessary: "as a complement to `time_bins()` and `lat_bins_*()` functions"

R/bin_space.R
etiennebacher: I have now deprecated these args and updated the code so that `spacing` and `sub_grid` still work, but we cannot replace `return` since we don't have a way to return `grid_info` anymore. I added a custom error message instead of removing it from the function definition so that people see this message instead of 

```r
Error in `bin_space()`:
! unused argument (return = TRUE)
```

R/bin_space.R
etiennebacher: I suppose `bins` makes sense given the other `lat_bins_*()`, `time_bins()`, etc.

R/bin_space.R
etiennebacher: > I worry that this could be confusing to users/readers because it is the value they supplied to space_bins, not the actual spacing in km (which could be very different)

I'm not sure I follow, would you rather use the average spacing in km as reported in the message, e.g. "Average spacing between adjacent cells in the primary grid was set to 725.17 km."?

R/bin_space.R
etiennebacher: This makes sense to me. Going this way would also allow people to use other tools to generate the polygons they want, e.g. `a5R`: https://belian-earth.github.io/a5R/

If we go this way, we should also reconsider `space_bins()`:

- keep it with H3 only, but we should explain why we only allow generating H3 polygons
- keep it and allow it to be extended later, e.g. with A5 so that one could do `space_bins(tool = "a5")`
- remove it and show people how they can use `h3o`, `a5R`, and other tools to generate their polygons

PR #346, opened 2026-09-09
Add `NEWS` item for mandatory named arguments

etiennebacher:
Part of #216 

On top of the NEWS item, I don't really see where to document this in a general way. One way to do it would be to have an empty reference page that would just explain this behaviour and provide a list of which arguments don't have to be named in each function, but I'm not convinced this page would be easy to find. We could also say that error messages are clear enough.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,7 +14,7 @@
   # Partial argument names are not allowed
   tax_certainty(taxdf, nam = "identified_name")
   ```
-  This should be used instead:o
+  This should be used instead:
 
   ```r
   tax_certainty(taxdf, name = "identified_name")
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # palaeoverse (development version)
 
-This version contains many breaking changes. See the [2.0 migration guide]() for
+This version contains many breaking changes. See the 2.0 migration guide for
 guidance to update existing code.
 
 ## Breaking changes
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,16 +8,16 @@
   argument is a `data.frame` (#216). For example, the following cases would fail:
 
   ```r
-  # `occdf` can be unnamed, but `nrow` must be named
-  group_apply(occdf, group = "cc", nrow)
+  # `taxdf` can be unnamed, but `identified_name` must be named
+  tax_certainty(taxdf, "identified_name")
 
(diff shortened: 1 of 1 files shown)


willgearty:
I worry that NEWS.md is going to get bloated if we give examples for every breaking change in the release. Would it be better to just stick with the note about the change in NEWS.md and then have a separate article about migrating from palaeoverse 1.x to palaeoverse 2.x?

In terms of documenting this change, I think the error messages and the NEWS/article mentions are sufficient.

NEWS.md
```diff
@@ -2,6 +2,24 @@
 
 * `palaeoverse` requires R >= 4.1.0 (#181).
 
+## Breaking changes
+
+* All functions now require arguments to be fully named, except when the first 
+  argument is a `data.frame` (#216). For example, the following cases would fail:
```
LewisAJones: This is not strictly true as we also allow a tree in `phylo_check`, so it's just the first argument?

LewisAJones:
> I worry that NEWS.md is going to get bloated if we give examples for every breaking change in the release. Would it be better to just stick with the note about the change in NEWS.md and then have a separate article about migrating from palaeoverse 1.x to palaeoverse 2.x?
> 
> In terms of documenting this change, I think the error messages and the NEWS/article mentions are sufficient.

This is a fair point, worth just keeping it in the `NEWs.md` for now, and pulling it out later depending on how much "bloat" we have? I think there is value in the examples but agree if it becomes too long, nobody will read it anyway.

etiennebacher:
In the end I just removed the example, I agree that having a general migration guide would be useful.

LewisAJones (approved):

willgearty:
Should the migration guide be started here to appease the CI gods?

NEWS.md
```diff
@@ -1,7 +1,15 @@
 # palaeoverse (development version)
 
+This version contains many breaking changes. See the [2.0 migration guide]() for
```
etiennebacher:
```suggestion
This version contains many breaking changes. See the 2.0 migration guide for
```

willgearty (approved):
Yeah, that works for me. Just add an issue to keep a record that we need to make it (soonish).

etiennebacher:
I had planned to commit this directly but lost the connection on my phone 😄

Issue #167, opened 2026-05-14
Set up a parameter dictionary


etiennebacher:
Notes from our meetings in August:

- [ ] get rid of `occdf` and `taxdf` in all functions and instead have a single argument name. We haven’t decided yet the name of this new argument (candidates are `data`, `df`, `d`). Previous arguments will be deprecated. (#348)

- [ ] rename `n` to `n_bins` in `lat_bins_area()` (#349)

- [ ] change the default value of `append` from `FALSE` to `TRUE` in `tax_unique()`. We will clarify that `tax_unique()` also returns a dataframe when `append = FALSE`. (#350)

- [ ] harmonize names to `timescale`

- [ ] centralize this dictionary of parameters and their definitions internally.

- Some functions sometimes require specific column names and sometimes allow the user to specify the column names. For instance, bin_lat() gives the choice for the “lat” column in occdf but forces `bins` to have columns named “max”, “min”, and “bin”.
  - task: explore how often this happens and whether we should add more arguments to those functions. This may lead to functions that are hard to read because they have a lot of arguments.

- [ ] remove `group` from the three `tax_*()` functions have a `group` argument. The user should use `group_apply()` instead (we will add examples for that).

PR #349, opened 2026-09-10
Replace `n` by `n_bins` in `lat_bins_area()`

etiennebacher:
Part of #167 

This follows https://lifecycle.r-lib.org/articles/communicate.html#renaming-an-argument.

### DIFF
--- a/R/lat_bins_area.R
+++ b/R/lat_bins_area.R
@@ -47,11 +47,11 @@ lat_bins_area <- function(
   max = 90,
   r = 6371,
   plot = FALSE,
-  n
+  n = deprecated()
 ) {
   ensure_args_are_named()
 
-  if (!missing(n)) {
+  if (lifecycle::is_present(n)) {
     lifecycle::deprecate_warn(
       "2.0.0",
       "lat_bins_area(n)",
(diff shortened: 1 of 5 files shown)


willgearty (approved):
Looks good, thanks!

R/lat_bins_area.R
```diff
@@ -5,7 +5,7 @@
 #' based on calculating the curved surface area of spherical segments bounded
 #' by two parallel discs.
 #'
-#' @param n \code{numeric}. A single numeric value defining the number of
+#' @param n_bins \code{numeric}. A single numeric value defining the number of
```
LewisAJones, line 8: Whenever we are changing argument names, we should update the examples accordingly to prevent confusion for new users. We currently still use `n`.

LewisAJones (changes requested):
Thanks @etiennebacher! Just one comment.

R/lat_bins_area.R
etiennebacher: Right, I thought R CMD check would error if there's a warning in docs. Thanks!

LewisAJones: I guess technically it passes? 🥲

LewisAJones (approved):

PR #350, opened 2026-09-10
Change default value of `append` in `tax_unique()`

etiennebacher:
Part of #167 

### DIFF
--- a/R/tax_unique.R
+++ b/R/tax_unique.R
@@ -32,15 +32,19 @@
 #' @param append \code{logical}. Should the original dataframe be returned with
 #' the unique names appended as a new column?
 #'
-#' @return A \code{dataframe} of taxa, with each row corresponding to a unique
-#' "species" or "genus" in the dataset (depending on the chosen resolution).
-#' The dataframe will include the taxonomic information provided into the
-#' function, as well as a column providing the 'unique' names of each taxon. If
-#' \code{append} is \code{TRUE}, the original dataframe (\code{occdf}) will be
-#' returned with these 'unique' names appended as a new column. Occurrences that
-#' are identified to a coarse taxonomic resolution and belong to a clade which
+#' @return
+#' This always returns a dataframe.
+#'
+#' If \code{append} is \code{TRUE}, the original dataframe (\code{occdf})
+#' is returned with a new column providing the 'unique' names of each taxon
+#' ("species" or "genus" depending on the chosen resolution). Occurrences
+#' that are identified to a coarse taxonomic resolution and belong to a clade which
 #' is already represented within the dataset will have their 'unique' names
 #' listed as \code{NA}.
+#'
+#' If \code{append} is \code{FALSE}, this returns a new dataframe of taxa. The
+#' dataframe will include the taxonomic information provided into the function,
+#' as well as a column providing the 'unique' names of each taxon.
 #'
 #' @details Palaeobiologists usually count unique taxa by retaining only
 #' unique occurrences identified to a given taxonomic resolution, however
@@ -139,7 +143,7 @@ tax_unique <- function(
   ...,
   name = NULL,
   resolution = "species",
-  append = FALSE
+  append = TRUE
 ) {
   ensure_args_are_named(exceptions = "occdf")
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,8 +7,11 @@ guidance to update existing code.
 
 * `palaeoverse` requires R >= 4.1.0 (#181).
 
-* All functions now require arguments to be fully named, except for the first 
+* All functions now require arguments to be fully named, except for the first
   argument in some cases (#216).
+
+* In `tax_unique()`, the default value of `append` is changed from `FALSE` to
+  `TRUE` (#350).
 
 ## Bug fixes
 
--- a/man/tax_unique.Rd
+++ b/man/tax_unique.Rd
@@ -12,7 +12,7 @@ tax_unique(
   ...,
   name = NULL,
   resolution = "species",
-  append = FALSE
+  append = TRUE
 )
 }
 \arguments{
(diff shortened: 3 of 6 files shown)


R/tax_unique.R
```diff
@@ -32,16 +32,20 @@
...
-#' \code{append} is \code{TRUE}, the original dataframe (\code{occdf}) will be
-#' returned with these 'unique' names appended as a new column. Occurrences that
-#' are identified to a coarse taxonomic resolution and belong to a clade which
+#' @return
+#' This always returns a dataframe.
+#'
+#' If \code{append} is \code{TRUE}, the original dataframe (\code{occdf})
+#' is returned with a new column providing the 'unique' names of each taxon
+#' ("species" or "genus" depending on the chosen resolution). Occurrences
+#' that are identified to a coarse taxonomic resolution and belong to a clade which
```
willgearty, line 41: Are these rows also included when append is False?

willgearty (approved):

R/tax_unique.R
etiennebacher: No and I think this wasn't tested. I've added [`28820d5`](https://github.com/palaeoverse/palaeoverse/pull/350/changes/28820d58827d2b0ad66495c2299bc995ee2f0673), can you double check it is correct?

willgearty: Yep, that looks correct, thanks!

R/tax_unique.R
```diff
@@ -32,16 +32,22 @@
...
 #'
-#' @return A \code{dataframe} of taxa, with each row corresponding to a unique
-#' "species" or "genus" in the dataset (depending on the chosen resolution).
-#' The dataframe will include the taxonomic information provided into the
-#' function, as well as a column providing the 'unique' names of each taxon. If
-#' \code{append} is \code{TRUE}, the original dataframe (\code{occdf}) will be
-#' returned with these 'unique' names appended as a new column. Occurrences that
-#' are identified to a coarse taxonomic resolution and belong to a clade which
+#' @return
+#' This always returns a dataframe.
```
LewisAJones, line 36:
```suggestion
#' A \code{data.frame}.
```
This could be shortened? Should we be explicit about the class as we are with inputs (just here, everywhere, nowhere)?

R/tax_unique.R
LewisAJones, line 10:
```suggestion
#' @param occdf \code{data.frame}. A dataframe containing information on the
```
Should we be explicit about expected class here? Though I guess we also support tibbles?

LewisAJones, line 117: The examples should probably be updated to reflect this change in default behaviour?

R/tax_unique.R
```diff
@@ -32,16 +32,22 @@
...
+#'
+#' If \code{append} is \code{TRUE}, the original dataframe (\code{occdf})
+#' is returned with a new column providing the 'unique' names of each taxon
```
LewisAJones, line 39: I think we need to make a decision about our behaviour with "new" columns here (and throughout the package). Currently, if we had an existing column named `unique_name`, it would be overwritten by this function. So we probably either want to prevent that, or document (which is why I am raising it here) that that happens and specify the column names?

R/tax_unique.R
```diff
@@ -139,7 +145,7 @@ tax_unique <- function(
   ...,
   name = NULL,
   resolution = "species",
-  append = FALSE
+  append = TRUE
```
LewisAJones, line 148: While I stand by the decision to make a default value for append arguments throughout the package, I actually don't think this should be `TRUE` here. In fact, I am starting to question whether we should even have this argument in the first place.

For similar functions which interpret the data in some way (e.g. `tax_range_time`), we don't actually have such an argument and I believe most users would just want to get a table of the unique taxa. This is why we had `append = FALSE` originally.

LewisAJones (changes requested):
Thanks for this @etiennebacher, a few comments for discussion! Worth looking through them all before implementing any changes.

R/tax_unique.R
willgearty: But append=FALSE doesn't return rows that don't have a unique name. While that's probably preferred for most cases, some users likely want to know which rows aren't unique?

LewisAJones: Do you have an example of when somebody would use that information? I don't immediately see the use case. 

Here, unique names are interpretations based on the available data. Which occurrences are identified as representing unique names is therefore quite variable? In the case of an occurrence dataframe, all occurrences/rows should be unique in some regard (e.g. in geography/time)? I see this function as providing a summary more than anything. 

In general, we probably want to encourage users to use this function alongside `group_apply`... getting this summary by collection, time bin, spatial bin, etc (we do already have examples on this).

Regardless, whether we keep this argument or not, it looks like you agree @willgearty that `append=FALSE` would be the preferred usage of this.

As this is now a wider discussion rather than a small implementation change, do you have any thoughts on all this @bethany-j-allen as the original developer?

willgearty: I agree `append=FALSE` is a good default for this particular function. I'd have to look at all of the other functions that have `append` to see if I think the same default makes sense across the board.

I appreciate your point about summary vs. data manipulation/enhancement (e.g., `palaeorotate`), but I'm not sure I'm entirely convinced the `append` argument should be removed from this function.

LewisAJones: Side note, if we do decide to keep `append` here, we should also consider whether we should have `append` in other summary functions (e.g. `tax_range_time` and `tax_range_space`).

bethany-j-allen: I originally added the `append = TRUE` option for two reasons: (1) it is the smoothest route to using the data in downstream analyses, as otherwise you would need to join the output to your original dataframe anyway (and this could be messy, especially with higher taxa which are unique), and (2) `append = FALSE` gives you all of the unique names but `append = TRUE` includes all of the rows in the original dataframe so can be used to quickly calculate frequencies (I think?). As you say, the table of unique names given by `append = FALSE` is closer in spirit to the exploratory tools we wanted to develop, which I think is why I preferred this as the default option. I wrote the `append` option in because I thought it would be useful from my personal perspective, but if you think it is not in keeping with the rest of the package then it could be removed.

Issue #352, opened 2026-09-14
Move all `plot` arguments to S3 methods of `plot()`


etiennebacher:
Having a `plot` argument can be convenient for quick plots because we don’t need to create a new object, but at the same time the user could just write `plot(f(...))` instead of `f(..., plot = TRUE)`. Removing this argument cleans up the function signature and ensures that the function always returns the same output type.

Moreover, this will allow us to move dedicated plotting arguments (e.g. `plot_args`) to these custom methods.

The affected functions will always return a dataframe with a custom class and they will have custom S3 plot methods for all those functions.

PR #353, opened 2026-09-15
Remove `plot` argument and add S3 `plot()` method in `lat_bins_area()`

etiennebacher:
Part of #352.

I'll do the same for other concerned functions once this is merged.


R/palaeoverse-package.R
```diff
@@ -8,3 +8,8 @@
 #' @importFrom utils head
 ## usethis namespace: end
 NULL
+
+#' S3 methods for `plot()` for various `palaeoverse` objects
+#'
+#' @name plot_palaeo
+NULL
```
etiennebacher, line 15: We could store all `plot()` methods in a separate reference page, or keep each one in the same page as its related function, I don't have a strong preference on this.

R/lat_bins_area.R
```diff
@@ -34,19 +32,20 @@
...
+#'
 #' # Generate latitudinal bins for just the (sub-)tropics
-#' bins <- lat_bins_area(n = 6, min = -30, max = 30)
+#' lat_bins_area(n = 6, min = -30, max = 30)
+#'
 #' # Generate latitudinal bins and a plot
-#' bins <- lat_bins_area(n = 24, plot = TRUE)
-lat_bins_area <- function(n = 12, min = -90, max = 90, r = 6371, plot = FALSE) {
+#' plot(lat_bins_area(n = 24))
+lat_bins_area <- function(n = 12, min = -90, max = 90, r = 6371) {
```
etiennebacher: Do we want to deprecate this argument in all functions that have it, or should we just remove it?

R/lat_bins_area.R
```diff
@@ -90,26 +89,36 @@ lat_bins_area <- function(n = 12, min = -90, max = 90, r = 6371, plot = FALSE) {
...
+#' @param y Ignored
+#' @param ... Ignored
+#'
+#' @name plot_palaeo
+#' @export
+plot.palaeo_lat_bins_area <- function(x, y, ...) {
+  # We want to pass `plot(<something>)`
+  if (missing(y)) {
+    invisible()
+  }
```
etiennebacher, line 130: We need this `x, y, ...` signature even if we don't use `y` because it's what the `plot()` generic requires. Having `x, ...` only would raise an error in R CMD check, cf https://cran.r-project.org/doc/manuals/R-exts.html#Checking-packages-1:

> The R code is checked for possible problems using [codetools](https://cran.r-project.org/package=codetools). In addition, it is checked whether S3 methods have all the arguments of the corresponding generic

R/lat_bins_area.R
etiennebacher: I chose to deprecate it for now, in line with other args e.g. https://github.com/palaeoverse/palaeoverse/pull/349

R/palaeoverse-package.R
willgearty: I would put each one alongside the function that sets the method on the object. And within the same file as that function as well.

R/lat_bins_area.R
```diff
@@ -112,26 +124,41 @@ lat_bins_area <- function(
...
+  plot(
+    1,
+    type = "n",
+    xlim = c(-180, 180),
+    ylim = c(min(x$min), max(x$max)),
+    xlab = "Longitude (\u00B0)",
+    ylab = "Latitude (\u00B0)"
```
willgearty: Should we pass along `...`?

willgearty (approved):
Looks good, thanks!

R/lat_bins_area.R
LewisAJones: While I don't love it, it is probably better to deprecate and will help people update their code.

R/lat_bins_area.R
```diff
@@ -112,26 +124,41 @@ lat_bins_area <- function(
...
-  # plot latitudinal bins
-  if (plot) {
-    plot(
-      1,
-      type = "n",
-      xlim = c(-180, 180),
-      ylim = c(min(bins$min), max(bins$max)),
-      xlab = "Longitude (\u00B0)",
-      ylab = "Latitude (\u00B0)"
+  class(bins) <- c("palaeo_lat_bins_area", class(bins))
```
LewisAJones: Is there a strong reason to use "palaeo_lat_bins_area" over just "lat_bins_area"?

R/lat_bins_area.R
```diff
@@ -112,26 +124,41 @@ lat_bins_area <- function(
...
+  bins
+}
+
+#' @param x The object to plot
+#' @param y Ignored
+#' @param ... Ignored
+#'
+#' @name lat_bins_area
+#' @export
+plot.palaeo_lat_bins_area <- function(x, y, ...) {
```
LewisAJones: As above, is there a strong reason to use "palaeo_lat_bins_area" over just "lat_bins_area"?

R/lat_bins_area.R
```diff
@@ -112,26 +124,41 @@ lat_bins_area <- function(
...
+  # We want to pass `plot(<something>)`
+  if (missing(y)) {
+    invisible()
+  }
+  plot(
```
LewisAJones, line 180: Along similar lines of Will's comment, I envisioned some of the value of pulling `plot` out from the main function was to support more customisability of plots (e.g. colours, titles, etc) but with helpful defaults.

Thoughts?

R/lat_bins_area.R
```diff
@@ -112,26 +124,41 @@ lat_bins_area <- function(
...
+  class(bins) <- c("palaeo_lat_bins_area", class(bins))
+
+  if (isTRUE(plot)) {
+    plot(bins)
+  }
+
+  bins
+}
+
+#' @param x The object to plot
```
LewisAJones: Is there value in being more explicit here? e.g. "The output from `lat_bins_area`" or "A `lat_bins_area` object"?

R/lat_bins_area.R
```diff
@@ -112,26 +124,41 @@ lat_bins_area <- function(
...
+#' @param y Ignored
+#' @param ... Ignored
+#'
+#' @name lat_bins_area
```
LewisAJones, line 144: Currently, we're grouping the documentation for plotting with the documentation for `lat_bins_area` itself. Is this potentially confusing for users (I'm not convinced it is, but just flagging)? If we're not going to allow greater customisability of plots (see comment below), do we even need to expose these arguments in the documentation?

Image

R/lat_bins_area.R
etiennebacher: I added this and an extra check because we already set some args internally.

R/lat_bins_area.R
LewisAJones: Having `y ignored` throughout the documentation isn't ideal, but there also doesn't seem to be a clean way around this

LewisAJones (changes requested):

R/lat_bins_area.R
etiennebacher: I passed `...` to our `plot` method in https://github.com/palaeoverse/palaeoverse/pull/353/changes/3b53caadc691ec28285a3c697920ca6b9513de4b (probably while you were writing this review), what do you think? (I also added a couple of small commits after this one)

R/lat_bins_area.R
etiennebacher: I just thought it would be helpful to have the same "palaeo" prefix for all classes created by `palaeoverse`, including the one from `as_palaeo()`. This also reduces the risk of conflicts with other packages. In this case, `lat_bins_area` should be unique enough to not be used by other packages, but e.g. a class `time_bins` could exist elsewhere.

LewisAJones: I don't think those conflicts would happen frequently, but reducing the chance of this happening certainly isn't a bad idea. 

I was initially thinking that it would be nice to mirror the function name directly and avoid any potential confusion about what the object represents. For example, we often use "palaeo" as a prefix to imply the nature of something such as "palaeolatitude" for reconstructed latitudes. Perhaps an alternative standard here is to use "palaeoverse" instead of "palaeo" (I'm also mindful that there are many palaeo-flavoured packages), and use it as a suffix instead of a prefix. Using it as suffix has the added benefit of immediately reading the function name too?

Is there any value in having multiple classes? tibble does this:

```
library(tibble)
x <- tibble(a = 1:3)
> class(x)
[1] "tbl_df"     "tbl"        "data.frame"
```

etiennebacher: > Perhaps an alternative standard here is to use "palaeoverse" instead of "palaeo" (I'm also mindful that there are many palaeo-flavoured packages)

I don't mind using "palaeoverse" instead.

> use it as a suffix instead of a prefix. Using it as suffix has the added benefit of immediately reading the function name too?

I don't think this is a big deal because we don't expect users to use these class names directly anyway, this is just something for us to print in messages or for other package developers to support in their packages.

I like having it as prefix because it's immediately clear this is one of our classes. We do this in `polars` for instance:

image

> Is there any value in having multiple classes? tibble does this:

This is already the case here:

```r
> lat_bins_area(n_bins = 12) |> class()
[1] "palaeo_lat_bins_area" "data.frame"   
```

R/lat_bins_area.R
LewisAJones: It works, but I don't think it captures everything. For example, `col` is still fixed, and this also isn't listed as forbidden? 

I also wonder whether we actually want to make anything forbidden or just handle it elegantly? For example, if somebody provides a custom `xlab`, should we just overwrite the current one? Though, it makes sense with `type` in this case. 

Sorry for being picky and starting discussion on all this, but better to get this correct at this stage before full implementation.

R/lat_bins_area.R
etiennebacher: >  do we even need to expose these arguments in the documentation?

We have to expose them somewhere, so either each plot method is on the same page as the function producing the class passed to the plot method (the benefit being that it is immediately clear that a function has a custom plot method), or we gather all plot methods in a single reference page (but then we end up with generic `x` descriptions for instance).

R/lat_bins_area.R
LewisAJones: I think I'd prefer palaeoverse then... this makes it super clear that it is from our package, and with your justification of why you prefer it as a prefix, this makes a good combo.

Sorry, I should have been clearer, I meant having three classes i.e. "palaeoverse", "lat_bins_area", and "data.frame". However, if we're going with "palaeoverse_*", I think I'm happy!

R/lat_bins_area.R
LewisAJones: Let's stick with the former then! I don't think the current infrastructure supports it, but it would have been nice to have the `lat_bins_area` function listed and all arguments, and then immediately underneath, the `plot` function and relevant arguments.

etiennebacher: Yeah I agree, I hate it when there are multiple functions with many arguments on one page and all args are listed in one section. Just a limitation of R `man` pages though.

R/lat_bins_area.R
etiennebacher: The user can now customize `col`, `xlab`, and `ylab`: f07bc7756c8b59616fe4f68b4ededccaf07eba51

`type`, `xlim` and `ylim` should probably stay hardcoded.

R/lat_bins_area.R
```diff
@@ -112,26 +124,75 @@ lat_bins_area <- function(
...
+  }
+
+  bins
+}
+
+#' @param x An object of class `"palaeoverse_lat_bins_area"` (created by `lat_bins_area()`).
+#' @param y Ignored
+#' @param ... Extra arguments passed to [`plot()`][base::plot]. The following arguments are
+#' already set internally and must not be specified here: `type`, `xlim`, `ylim`.
+#' @param col Character vector of length 2 indicating the colours to use for the stripes.
```
LewisAJones, line 140: Currently, plot argument documentation do not conform with our other documentation standards (e.g. specifying the class).

R/lat_bins_area.R
```diff
@@ -112,26 +124,75 @@ lat_bins_area <- function(
...
+#' @param xlab Title of the x-axis.
```
LewisAJones, line 141: So our documentation standard for plot.* will be to only document those that have helpful default values but can be customised?

R/lat_bins_area.R
LewisAJones, line 140:
```suggestion
#' @param col Character vector of length 2 indicating the colours to use for the bins.
```

etiennebacher: I added it for the three args after `...`, but for `x` the following looks a bit weird to me (so I haven't included it yet):

```
#' @param x `palaeoverse_lat_bins_area`. An object created by `lat_bins_area()`.
```

Are you fine with this or do you prefer keeping the current description for `x`?

R/lat_bins_area.R
etiennebacher: I'm not sure I understand, do you mean that these options could also be passed in `...` and we could use the default values internally if they are absent? We can also do that to reduce the number of "special" arguments, but at the same time we would still need input validation e.g. for `col`, so it might be confusing to have input checks for some elements of `...` but not others.

LewisAJones: With this comment, I was just trying to establish our documentation rule surrounding plotting functions.

In this function, we currently specify default values for the arguments you've documented. However, we don't document the entire functionality available right? e.g. `main` or `asp`, etc (these are available via the `base::plot` documentation).

R/lat_bins_area.R
LewisAJones: I actually kind of like this... it makes it very clear what the expected input is?

It's longer than what we would probably like, but it's the class name we went with to avoid conflicts. We could also use `data.frame` I guess, but this is clearer.

As this is a decision that will get propagated out to all plotting, let's also get @willgearty's thoughts.

R/lat_bins_area.R
```diff
@@ -134,12 +134,12 @@ lat_bins_area <- function(
 }
 
 #' @param x An object of class `"palaeoverse_lat_bins_area"` (created by `lat_bins_area()`).
-#' @param y Ignored
+#' @param y Ignored.
```
LewisAJones: Worth adding a few words on why this is listed here but ignored?... Thinking from a user perspective and a first-time contributor.

R/lat_bins_area.R
willgearty: We couldn't possibly document the entire functionality (not even `plot.default()` does that). That's already documented by `par()` (which we could link to). Could also do `inheritDotParams` from `plot.default()`, I think?

R/lat_bins_area.R
willgearty: My gut reaction was to suggest using `data.frame` and then just keep the clarifying sentence afterwards.

R/lat_bins_area.R
willgearty: I just checked and `plot.histogram()` and `plot.lm()` don't have `y` arguments. I'm confused why it works there but not here?

etiennebacher: Actually `plot()` is a special case in `tools::checkS3methods()` and its second argument (`y`) is skipped:

```r
> tools::checkS3methods

[...]

check_args <- function(gName, gArgs, mName, mArgs) {
        [...]
        if (gName == "plot") 
            gArgs <- gArgs[-2L]

[...]
```

so we could remove it from this definition.

R/lat_bins_area.R
etiennebacher: Changed to:

```
#' @param x `data.frame`. An object of class `"palaeoverse_lat_bins_area"` created by `lat_bins_area()`.
```

R/lat_bins_area.R
etiennebacher: Just to recap a discussion with Lewis that can serve as reference for other `plot`-related PRs:

The current implementation is probably the best, i.e. all `plot()` arguments that we set internally and that are not essential to the plot output are added to the function definition, with their default value being what was set internally until now. Other args can be passed to `...` and don't need to be documented (and we couldn't document all of them anyway as explained by @willgearty).

In this case, `col`, `xlab`, and `ylab` are not essential so they are put in the function definition with a default value matching the previous implementation so that users don't have to explicitly pass these args. `type`, `xlim`, and `ylim` are key to the visualisation so they cannot be modified. 

> Could also do inheritDotParams from plot.default(), I think?

I don't think so because we want to specify which arguments cannot be modified by users (`type`, `xlim`, and `ylim` here).

R/lat_bins_area.R
etiennebacher: Removed in 4e01a0be33e4a80b66baf4121bbf87313662976c

R/lat_bins_area.R
willgearty: Sounds good to me 👍

R/lat_bins_area.R
```diff
@@ -112,26 +124,67 @@ lat_bins_area <- function(
...
+    plot(bins)
+  }
+
+  bins
+}
+
+#' @param x `data.frame`. An object of class `"palaeoverse_lat_bins_area"` created by `lat_bins_area()`.
+#' @param ... Extra arguments passed to [`plot()`][base::plot]. The following arguments are
+#' already set internally and must not be specified here: `type`, `xlim`, `ylim`.
+#' @param col `character`. Character vector of length 2 indicating the colours to use for the bins.
```
LewisAJones:
```suggestion
#' @param col `character`. A character vector of length 2 indicating the colours to use for the bins.
```

R/lat_bins_area.R
```diff
@@ -112,26 +124,67 @@ lat_bins_area <- function(
...
+#' @param xlab `character`. Title of the x-axis.
```
LewisAJones:
```suggestion
#' @param xlab `character`. The x-axis title.
```

R/lat_bins_area.R
```diff
@@ -112,26 +124,67 @@ lat_bins_area <- function(
...
+#' @param ylab `character`. Title of the y-axis.
```
LewisAJones:
```suggestion
#' @param ylab `character`. The y-axis title.
```

R/lat_bins_area.R
```diff
@@ -112,26 +124,67 @@ lat_bins_area <- function(
...
+
+  dots <- list(...)
+  if (length(dots) > 0) {
+    nms <- names(dots)
+    forbidden <- nms[nms %in% c("type", "xlim", "ylim")]
+    if (length(forbidden) > 0) {
+      cli::cli_abort(
+        c(
+          "{cli::qty(forbidden)} Cannot pass argument{?s} {.arg {forbidden}} when calling {.fn plot} on an object of class {.cls palaeoverse_lat_bins_area}.",
+          "i" = "{cli::qty(forbidden)}{?This/These} argument{?s} {?is/are} already set by `plot()` internally."
```
LewisAJones, line 166:
```suggestion
          "i" = "{cli::qty(forbidden)}{?This/These} argument{?s} {?is/are} already set by `plot.palaeoverse_lat_bins_area()` internally."
```

Or is this too specific and unhelpful?

R/lat_bins_area.R
```diff
@@ -112,26 +124,67 @@ lat_bins_area <- function(
...
+#' @param col `character`. Character vector of length 2 indicating the colours to use for the bins.
+#' @param xlab `character`. Title of the x-axis.
+#' @param ylab `character`. Title of the y-axis.
+#'
+#' @name lat_bins_area
+#' @export
+plot.palaeoverse_lat_bins_area <- function(
+  x,
+  ...,
+  col = c("#01665e", "#80cdc1"),
```
LewisAJones, line 148: Unnecessary at this point, but we could consider using the Palaeoverse colour palette / brand in future.

R/lat_bins_area.R
LewisAJones: Just trying to harmonise phrasing of documentation...

R/lat_bins_area.R
LewisAJones: Just trying to harmonise phrasing of documentation...

LewisAJones (approved):
Final points for consideration, but otherwise looks great! It should be easier for future plot-related PRs now we've got some key decisions out the way.

Thanks @etiennebacher!

R/lat_bins_area.R
etiennebacher: ~I thought it could help when debugging. Since `plot()` is generic, explicitly showing the method it was dispatched to helps ensuring that the user passed an object of the right class.~ Forget that, I misread your comment.

LewisAJones: I guess I was thinking that quoting `plot.palaeoverse_lat_bins_area()` was more helpful for debugging as these arguments are not forbidden in `plot()`, but our usage? However, I can see it from the other perspective too, so I'm happy with it either way.

etiennebacher: The class is already specified in the first line of the message so I don't think this is necessary.

willgearty (approved):
All looks good to me!

PR #358, opened 2026-09-21
Update `r-lib/actions` to 2.14.0

etiennebacher:
macOS workflows started failing because of https://github.com/r-lib/pak/issues/915. `r-lib/actions` v2.14.0 fixes it and in this case I don't think we should wait for the two-week cooldown period set in dependabot since this prevents us for having R CMD check results on macOS.


etiennebacher:
Blocked by https://github.com/r-lib/actions/issues/1113

There's a workaround there that seems to be working but let's see if an official fix is provided soon.

willgearty (approved):

LewisAJones (approved):

